PR #8, opened 2021-08-27
Support querying by file attributes

SeanLeRoy:
**Context**
Not all metadata displayed in the FMS File Explorer is considered an Annotation, some are what we call "file attributes". The distinction is more technical than anything, but essentially treating file attributes as a special case allows us to make them sortable among other things. However, this also separates them from the typical filtering means which they wouldn't necessarily fit the best anyway. Instead, we can do a sort of "fuzzy" search by doing a string search most of the time and a date range search in the case of "Uploaded".

**Description of Changes**
This adds a search bar atop the file list that users can use to fuzzy search in combination with their other filters, groupings, and sort. 
**Example**
fuzzy search

### DIFF
--- a/packages/core/components/DirectoryTree/DirectoryTree.module.css
+++ b/packages/core/components/DirectoryTree/DirectoryTree.module.css
@@ -5,6 +5,7 @@
 }
 
 .filter-display-bar {
+    flex-shrink: 0;
     width: 100%;
     margin-bottom: calc(var(--margin) / 2); /* --margin defined in App.module.css */
     padding-bottom: 3px;
@@ -37,7 +38,7 @@
     transform: translateY(-50%);
     width: 100%;
 
-    color: var(--brand-error); /* defined in App.module.css */;
+    color: var(--brand-error); /* defined in App.module.css */
     font-size: 2rem;
     text-align: center;
 }
--- a/packages/core/components/DirectoryTree/DirectoryTree.module.css
+++ b/packages/core/components/DirectoryTree/DirectoryTree.module.css
@@ -1,12 +1,13 @@
 .container {
-    --filter-display-height: 50px;
+    display: flex;
+    flex-direction: column;
     position: relative;
 }
 
 .filter-display-bar {
     width: 100%;
-    height: var(--filter-display-height);
     margin-bottom: calc(var(--margin) / 2); /* --margin defined in App.module.css */
+    padding-bottom: 3px;
 }
 
 .filter-display-bar.hidden {
@@ -19,9 +20,9 @@
 }
 
 .scroll-container {
+    flex: auto;
     overflow-y: auto;
     width: 100%;
-    height: calc(100% - var(--filter-display-height) - calc(var(--margin) / 2)); /* --margin defined in App.module.css */
     margin: 0;
     padding: 0;
 }
--- a/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
+++ b/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
@@ -1,0 +1,22 @@
+.search-bar-container {
+    display: flex;
+    margin-bottom: 5px;
+}
+
+.search-box, .search-box-selector {
+    flex: auto;
+}
+
+.search-box-selector {
+    display: flex;
+}
+
+.search-box-button {
+    color: black;
+    border: 1px solid black;
+}
+
+.file-attribute-selector {
+    width: 100%;
+    max-width: 185px;
+}
--- a/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
+++ b/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
@@ -3,16 +3,16 @@
     margin-bottom: 5px;
 }
 
-.search-box, .search-box-selector, .date-range-box {
+.filter-input div i, .cancel-button, .cancel-button:hover {
+    color: var(--primary-brand-purple); /* defined in App.module.css */
+}
+
+.filter-input, .search-box-selector, .date-range-box {
     flex: auto;
 }
 
 .search-box-selector, .date-range-box {
     display: flex;
-}
-
-.search-box-button, .search-box-button:hover {
-    color: black;
 }
 
 .file-attribute-selector {
--- a/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
+++ b/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
@@ -3,20 +3,35 @@
     margin-bottom: 5px;
 }
 
-.search-box, .search-box-selector {
+.search-box, .search-box-selector, .date-range-box {
     flex: auto;
 }
 
-.search-box-selector {
+.search-box-selector, .date-range-box {
     display: flex;
 }
 
-.search-box-button {
+.search-box-button, .search-box-button:hover {
     color: black;
-    border: 1px solid black;
 }
 
 .file-attribute-selector {
     width: 100%;
     max-width: 185px;
+}
+
+.date-range-box {
+    border: solid black 1px;
+    max-height: 32px;
+    overflow: hidden;
+}
+
+.date-range-separator {
+    display: flex;
+    font-size: small;
+    padding: 0 2px;
+}
+
+.date-range-separator i {
+    margin: auto;
 }
--- a/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
+++ b/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
@@ -1,4 +1,6 @@
 .search-bar-container {
+    border: solid black 1px;
+    border-radius: 3px;
     display: flex;
     margin-bottom: 5px;
 }
@@ -8,6 +10,7 @@
 }
 
 .filter-input, .search-box-selector, .date-range-box {
+    border: none;
     flex: auto;
 }
 
@@ -20,8 +23,12 @@
     max-width: 185px;
 }
 
+.file-attribute-selector div span:first-child {
+    border: none;
+    border-right: 1px dashed black;
+}
+
 .date-range-box {
-    border: solid black 1px;
     max-height: 32px;
     overflow: hidden;
 }
--- a/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
+++ b/packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
@@ -5,10 +5,6 @@
     margin-bottom: 5px;
 }
 
-.filter-input div i, .cancel-button, .cancel-button:hover {
-    color: var(--primary-brand-purple); /* defined in App.module.css */
-}
-
 .filter-input, .search-box-selector, .date-range-box {
     border: none;
     flex: auto;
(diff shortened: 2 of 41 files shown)


packages/core/components/FileMetadataSearchBar/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+            : oldStartDate;
+        const endDate = newDateRange.endDate
+            ? newDateRange.endDate.toISOString().split("T")[0]
+            : oldEndDate;
+        onSearch(`${startDate || endDate}${DATE_RANGE_SEPARATOR}${endDate || startDate}`);
+    }
+
+    function onAttributeSelection(_: React.FormEvent, option?: IDropdownOption) {
+        onResetSearch();
+        option && setSelectedAttribute(option);
```
MatteBailey, line 86: To be consistent with how you make these checks in the other functions, I would just use a normal `if` statement to check for `option`.

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+import FileMetadataSearchBar, { DATE_RANGE_SEPARATOR } from "../";
+import FileFilter from "../../../entity/FileFilter";
+import { AnnotationName } from "../../../constants";
+import { initialState, selection } from "../../../state";
+
+describe("<FileMetadataSearchBar />", () => {
+    // TODO: Figure out a way to test this. it doesn't seem like the <Searchbox />
+    // component recognizes the react-testing-library "onSubmit" event signals nor
+    // does it recognize an "Enter" keydown or keypress event.
```
MatteBailey, line 15: Do any of the suggestions here help?

https://stackoverflow.com/questions/62542988/how-to-test-fluent-ui-dropdown-with-react-testing-library

SeanLeRoy: I'll try this out!

MatteBailey (approved):
Looks great!

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
SeanLeRoy: It isn't a "click" per se that this is failing to register. The imported component "Searchbox" appears to listen to "onSubmit" events which ideally for us are triggered via "Enter" keypresses. However, I can't get any of the solutions I've seen thus far to successfully trigger the callback (perhaps there is another issue I'm missing though!).

GabeMedrash:
Minor though important top level "code reviewer experience" note for the future: can you please name your branches with a short description (and an accurate JIRA id?). E.g., `feature/FMS-1512-search-box`. When checking this out locally, it's helpful to be able to see which branch to checkout when I run `git branch -a` and not have to do too much investigation.

GabeMedrash:
Screen Shot 2021-09-01 at 10 52 43 AM

Screen Shot 2021-09-01 at 10 52 53 AM

GabeMedrash:
The filters that are created via the searchbox are reflected in the FMS File Explorer URL, but aren't parsing correctly. For example: `fms-file-explorer://{"groupBy":[],"filters":[{"name":"uploaded","value":"2021-08-01-to-2021-08-01"}],"openFolders":[],"sort":{"annotationName":"uploaded","order":"DESC"}}` isn't working.

GabeMedrash:
Screen Shot 2021-09-01 at 11 00 29 AM

IMO, it would look better if the filter medallions are place above the search bar.

packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
```diff
@@ -0,0 +1,44 @@
+.search-bar-container {
+    border: solid black 1px;
+    border-radius: 3px;
+    display: flex;
+    margin-bottom: 5px;
+}
+
+.filter-input div i, .cancel-button, .cancel-button:hover {
```
GabeMedrash, line 8: Here and below: fluentui provides explicit mechanisms for applying custom classnames to very specific subcomponents. Is there a reason why you're not using that mechanism? This is extremely fragile.

packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
```diff
@@ -0,0 +1,44 @@
...
+}
+
+.file-attribute-selector {
+    width: 100%;
+    max-width: 185px;
+}
+
+.file-attribute-selector div span:first-child {
+    border: none;
+    border-right: 1px dashed black;
```
GabeMedrash, line 28: IMO, solid would look a little better

packages/core/components/FileMetadataSearchBar/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+    function onResetSearch() {
+        if (fileAttributeFilter) {
+            dispatch(selection.actions.removeFileFilter(fileAttributeFilter));
+        }
+    }
+
+    function onSearch(filterValue: string) {
+        if (filterValue && filterValue.trim()) {
+            const fileFilter = new FileFilter(selectedAttribute.key as string, filterValue);
+            if (fileAttributeFilter) {
```
GabeMedrash, line 65: Modify this logic a bit and wrap the actions to be dispatched in a `batchActions`?

GabeMedrash (changes requested):
A few errors that I hit (detailed in other comments) when exercising the feature

packages/core/components/FileMetadataSearchBar/index.tsx
SeanLeRoy: They have middleware that would otherwise get skipped in batchActions (unless that doesn't happen in this repo?)

packages/core/components/FileMetadataSearchBar/FileMetadataSearchBar.module.css
SeanLeRoy: Works for me :+1:

packages/core/components/FileMetadataSearchBar/index.tsx
GabeMedrash: Ah, that sounds correct

packages/core/components/DirectoryTree/DirectoryTree.module.css
```diff
@@ -1,12 +1,13 @@
 .container {
-    --filter-display-height: 50px;
+    display: flex;
+    flex-direction: column;
     position: relative;
 }
 
 .filter-display-bar {
```
GabeMedrash, line 7: Looks like there's something moderately funky happening: the bottom of the filter display bar is disappearing behind the search box UI:
Screen Shot 2021-09-07 at 9 59 46 AM

packages/core/components/FileMetadataSearchBar/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+) as IDropdownOption;
+// Color chosen from App.module.css
+const PURPLE_ICON_STYLE = { icon: { color: "#827aa3" } };
+export const DATE_RANGE_SEPARATOR = "-to-"; // Not arbitrary, defined per contract with FES
+
+// Because the datestring comes in as an ISO formatted date like 01-02-21
+// creating a new date from that would result in a date displayed as the
+// day before due to the UTC offset, to account for this we can add in the offset
+// ahead of time.
+function extractDateFromDateString(dateString?: string): Date | undefined {
```
GabeMedrash, line 37: Might as well write [a few](https://mochajs.org/#dynamically-generating-tests) test cases for this

packages/core/components/FileMetadataSearchBar/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+    text: a.displayName,
+}));
+const FILE_NAME_OPTION = FILE_ATTRIBUTE_OPTIONS.find(
+    (o) => o.key === AnnotationName.FILE_NAME
+) as IDropdownOption;
+// Color chosen from App.module.css
+const PURPLE_ICON_STYLE = { icon: { color: "#827aa3" } };
+export const DATE_RANGE_SEPARATOR = "-to-"; // Not arbitrary, defined per contract with FES
+
+// Because the datestring comes in as an ISO formatted date like 01-02-21
```
GabeMedrash, line 33: To me, ISO formatted means something like `YYYY-MM-DDTHH:mm:ss.sssZ`. Is that what the expected inputs to this are? Or `MM-DD-YY`?

packages/core/components/FileMetadataSearchBar/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+    date.setMinutes(date.getTimezoneOffset());
+    return date;
+}
+
+/**
+ * This component renders a dynamic search bar for querying file records by
+ * basic file attributes that are otherwise not queryable through usual means like
+ * in the <AnnotationSidebar /> for example.
+ */
+export default function FileMetadataSearchBar() {
```
GabeMedrash, line 51: When I use the search bar, copy the FMS Explorer URL, refresh the app and paste the URL back into the app, there isn't much visual indication of what filter is applied. And equally important, unless I know how the URL string works, I don't have a way to clear the filter.  IMO, this is an important piece to put into place before the feature is ready. I see two options:
1. Populate the search bar with the attribute filter applied, if any.
2. Figure out how to shoehorn attribute filters into the "filter display bar" 

Option 1 seems more straightforward (both as a developer and user of the application) to me?

packages/core/components/FileMetadataSearchBar/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+                    styles={PURPLE_ICON_STYLE}
+                    value={extractDateFromDateString(startDate)}
+                />
+                <div className={styles.dateRangeSeparator}>
+                    <Icon iconName="Forward" />
+                </div>
+                <DatePicker
+                    borderless
+                    className={styles.filterInput}
+                    ariaLabel="Select a end date"
```
GabeMedrash, line 116: Super nit: "a" -> "an"

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
GabeMedrash: Are you planning to resolve this TODO before merge?

GabeMedrash (changes requested):

packages/core/components/DirectoryTree/DirectoryTree.module.css
SeanLeRoy: Should be solved

packages/core/components/FileMetadataSearchBar/index.tsx
SeanLeRoy: It expects YYYY-MM-DD I'll update the comment to reflect that

kmitcham (approved):

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { expect } from "chai";
+import * as React from "react";
+import { Provider } from "react-redux";
+
+import FileMetadataSearchBar, { DATE_RANGE_SEPARATOR, extractDateFromDateString } from "../";
+import FileFilter from "../../../entity/FileFilter";
+import { AnnotationName, TOP_LEVEL_FILE_ANNOTATIONS } from "../../../constants";
+import { initialState, selection } from "../../../state";
+
+describe("<FileMetadataSearchBar />", () => {
```
GabeMedrash, line 12: Great looking tests 👍

GabeMedrash (approved):
LGTM!

PR #10, opened 2021-10-14
File Selection Sharing / Dataset Feature Expansion

SeanLeRoy:
**Context**
The new file sharing system in the FMS File Explorer will allow for sharing what we have been calling "Datasets" (fixed, immutable sets of file metadata frozen at a point-in-time) and also a new, similar, idea more broadly described as a "Collection". These "Collections" will be a set of File Ids that will allow the File Explorer Service to narrow down the possible files queries can be ran against effectively preserving a file selection. The hope with this is that we will be able to hit at least three major use cases:
* Quick file selection sharing (select files, right click, select "Share", send FMS File Explorer URL to whoever)
* Long-term file selection sharing. Whether to be used as a tiny-url of sorts in a paper or by other services like camera alignment to direct people to specific sets of files.
* & as we do currently, programmatic users who would like their machine learning algorithms, for example, to be ran with the same set of file metadata each time

**Description of Changes**
This feature expands on the existing concept of "Datasets" in the app by giving the user the ability to actually view said Dataset within the app in a new component, a header ribbon. Datasets are now described in the app as Fixed Collections where a Non-Fixed Collection is actually a set of real-time files the user can use to narrow down the possible files to query against. Collections, as we are describing them, are also shareable via the FMS File Explorer URL. 

A collection can be marked as "Private" where, if true, results in *other* users not having said collection available to them in their available collection dropdown. However, a private collection may shared with other users by having the collection selected and passing along the FMS File Explorer URL where, when received, will be automatically detected and loaded in as if it were always available. 

The existing feature of creating a Python snippet was not removed, but rather relocated to within the header ribbon as an export option. The hope behind this is that the context menu created when files are selected can be simplified to the concept of sharing. Users can "share" file selections by creating collections and then, if desired, exporting a Python snippet to pass along.

**Open Questions**
Should the hierarchy/filters be removed when the collection is changed?
Should the collection be auto-selected when created? Perhaps not ideal for the quick share method (the default one)?

**Example**
https://user-images.githubusercontent.com/41307451/137370102-621a8989-63a3-4289-9999-daf1aa19d789.mov


packages/core/state/selection/logics.ts
```diff
@@ -418,6 +439,18 @@ const selectNearbyFile = createLogic({
     type: [SELECT_NEARBY_FILE],
 });
 
+/**
+ * Interceptor responsible for processing the changed collection into
+ * a refresh action so that the resources pertain to the current collection
+ */
+const changeCollectionLogic = createLogic({
+    async process(_: ReduxLogicDeps, dispatch, done) {
+        dispatch(interaction.actions.refresh() as AnyAction);
```
MatteBailey, line 448: Having a logic that only dispatches another action seems like a smell to me. Could this be handled instead by adding some logic to the reducer, or by having a different logic handle both the refresh action and this one?

MatteBailey (approved):
This looks great! A ton of work here, good job.

packages/core/components/SearchableDropdown/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+export default function SearchableDropdown(props: Props) {
+    const searchBoxReference = React.useRef(null);
+    const [showDropdown, setShowDropdown] = React.useState(false);
+
+    function onMenuDismiss() {
+        props.onSearch("");
+        setShowDropdown(false);
+    }
+
+    return (
```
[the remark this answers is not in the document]
GabeMedrash: Meh, in that case, I agree with you

packages/core/entity/FileExplorerURL/test/fileexplorerurl.test.ts
```diff
@@ -350,22 +385,53 @@ describe("FileExplorerURL", () => {
...
+                datasetService
+            );
+
+            // Assert
+            expect(result).to.not.be.empty;
+        });
+
+        it("Returns error message when not in expected JSON format", async () => {
+            // Arrange
+            sinon.stub(datasetService, "getDataset").rejects(new Error("No Dataset found"));
```
GabeMedrash, line 434: From the test above, I expect this test to fail _because of this_ as well as whatever is happening with the `encodedUrl` below. This test would be stronger if there was just one condition that is under test/could cause a failure, and/or if the assertion was more specific.

packages/core/state/interaction/logics.ts
```diff
@@ -111,29 +115,17 @@ const downloadManifest = createLogic({
...
-                              annotationName: sortColumn.annotationName,
-                              ascending: sortColumn.order === SortOrder.ASC,
-                          }
-                        : undefined,
-                };
-                selections = [selection];
-            } else {
-                const fileSelection = selection.selectors.getFileSelection(state);
-                selections = fileSelection.toCompactSelectionList();
+                fileSelection = new FileSelection([
```
GabeMedrash, line 118: I'm not entirely certain how to interpret these changes, but, I'm not getting expected behavior from this feature when I have a collection selected. How did you test this change?

packages/core/state/interaction/logics.ts
```diff
@@ -505,69 +497,150 @@ const showContextMenu = createLogic({
...
+                hierarchy: selection.selectors.getAnnotationHierarchy(state),
+            });
+            navigator.clipboard.writeText(url);
+
+            batch(() => {
+                dispatch(succeedShareableFileSelectionLinkGeneration(collection));
+                dispatch(
+                    processSuccess(
+                        generateShareableFileSelectionLinkId,
+                        `Successfully created collection ${collection.name} & copied shareable link to clipboard.`
```
GabeMedrash, line 613: Extraordinarily minor, but, I think the ampersand is a little too causal.

GabeMedrash:
Bug report:
1. Create fixed collection with annotations A and B.
2. Group files by annotation C.
3. Switch to newly created collection.

Screen Shot 2021-11-01 at 1 27 37 PM

packages/core/entity/FileExplorerURL/test/fileexplorerurl.test.ts
SeanLeRoy: Ah hmm true! I'll split this up.

packages/core/state/interaction/logics.ts
SeanLeRoy: Ah this change isn't why this is failing, the CsvService wasn't pointing at the correct dataset API resource

GabeMedrash (approved):

PR #11, opened 2021-12-02
Fixed -> Frozen

SeanLeRoy:
This was a common request (to change the word "Fixed" to "Frozen") that I did before the demo today. Seems like it should carry over to the release for today as well.


GabeMedrash (approved):

MatteBailey (approved):

kmitcham (commented):
Too bad we also have Frozen cell lines.  Hard to find a good word that isn't already used in other contexts.  But once you teach the word, that becomes the best word to use.

PR #14, opened 2021-12-16
Update for 2022

GabeMedrash:
**Context:**
This project relies heavily on many OSS packages, each of which has its own large dependency tree. Changing one often has implications for another, perhaps because they both rely on the same `peerDependency`, perhaps because they both rely on the same runtime dependency. On top of it all, the NPM OSS community is quite dynamic--waiting a while to update dependencies often means there are many intermediate version updates you need to skip over. Sometimes this means actual breaking changes, sometimes not.

This is a large-scale change (LSC), intended to remove all reported vulnerabilities from our dependency tree and generally bring them up to date. Because we didn't do a great job doing this incrementally throughout the last year, we end up with LSC like this. Perhaps we can do better: maybe, as a policy, check on major dependency updates every time we do a feature.

**This changeset:**
- Update major runtime dependencies
- Update major build tooling
- Remove `lerna` build tool and use "native" npm workspaces (closes [FMS-1469](https://aicsjira.corp.alleninstitute.org/browse/FMS-1469))
- Transition remaining `enzyme` component tests to `@testing-library/react` (closes [FMS-1480](https://aicsjira.corp.alleninstitute.org/browse/FMS-1480))


packages/web/webpack/constants.js
```diff
@@ -6,7 +6,6 @@ const devServer = {
 const Env = {
     DEVELOPMENT: "dev",
     PRODUCTION: "production",
-    STAGE: "staging",
```
kmitcham, line 9: This is intentional?  We never want staging?

kmitcham (approved):

packages/core/components/DirectoryTree/DirectoryTreeNodeHeader.tsx
```diff
@@ -2,7 +2,7 @@ import classNames from "classnames";
 import { Spinner, SpinnerSize } from "@fluentui/react";
 import * as React from "react";
 import { useSelector } from "react-redux";
-import Tippy from "@tippy.js/react";
+import Tippy from "@tippyjs/react";
```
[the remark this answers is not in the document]
GabeMedrash: Yep, they removed the "." from the name in v4: https://github.com/atomiks/tippyjs-react/releases/tag/v4.0.0

packages/web/webpack/constants.js
GabeMedrash: Yep--there's no staging deployment for this tool, because it's not really "deployed" as such. It's "distributed." The language here could maybe be updated if it's confusing, but there are just two types of builds, "dev" builds and "non-dev" builds. The "non-dev" builds are what are distributed, so, I figured it's fair to continue to call them "production" builds. Let me know if you think otherwise.

GabeMedrash: This change I could bring back--I was trying to keep `web` and `desktop` config more or less the same, and while it doesn't make sense for `desktop` to have a "staging" (because it isn't deployed, it's distributed, so there's only "dev" and "not dev"), it would make sense for the `web` version of this to have a "staging." All said, the `web` version of this tool has never been deployed anywhere.

package.json
```diff
@@ -2,73 +2,74 @@
...
-    "@babel/cli": "7.13.x",
-    "@babel/core": "7.13.x",
-    "@babel/plugin-proposal-class-properties": "7.13.x",
-    "@babel/plugin-proposal-optional-chaining": "7.13.x",
-    "@babel/preset-env": "7.13.x",
-    "@babel/preset-react": "7.12.x",
-    "@babel/preset-typescript": "7.13.x",
-    "@babel/register": "7.13.x",
-    "@testing-library/react": "11.1.x",
+    "@babel/cli": "7.x",
```
MatteBailey, line 22: Is babel still needed for TypeScript in 2022? Can't you just the TypeScript compiler?

MatteBailey (approved):

package.json
GabeMedrash: Arguably `ts-loader` could have been used instead of `babel-loader` for years now. If I'm remembering correctly, however, I've been using `babel` instead because other parts of the development toolchain use it (like testing), so I figured it was better to only use one tool where possible, and I believe this follows the lead of `create-react-app`.

PR #15, opened 2022-01-06
Allen institute/fix/fms 1519

yrkim98:
Added a grey "no image" icon to the thumbnail container if no thumbnail is available.


packages/core/components/FileDetails/index.tsx
```diff
@@ -157,7 +165,26 @@ export default function FileDetails(props: FileDetails) {
                                         uri={`http://aics.corp.alleninstitute.org/labkey/fmsfiles/image${thumbnailUriPath}`}
                                     />
                                 </div>
-                            )}
+                            ) : fileDetails ? (
```
GabeMedrash, line 168: Can you find a way to do this without having a nested ternary?

GabeMedrash (changes requested):
Nice work, Brian! Looks great. Refactor out that nested ternary and this is ready to ship!

SeanLeRoy (approved):

packages/core/components/FileDetails/index.tsx
yrkim98: its pretty similar- but would it be better to use && and || like this? 
`
{ displayThumbnail && (
   //display thumbnail
) || (
   fileDetails && (
      //show icon
   )
)}
`

GabeMedrash: Yeah, that feels a little too similar to me (😄)--for me, the readability concerns of a nested ternary come down to over-terness. Here's how I might approach this: above the return statement, use more verbose logic to assign a React element to a variable, then in the JSX expression returned from this function, make use of that variable. From 10,000 ft, that looks like:
```
function MyReactComponent() {

    let thumbnail = null;
    if (something) {
        thumbnail = (
            <div>...</div>
        );
    } else if (somethingElse) {
        thumbnail = (
             <Foo />
        );
    }

     return (
        <>
            ...
            {thumbnail}
        </>
     );
}
```

packages/core/components/FileDetails/index.tsx
```diff
@@ -86,15 +88,59 @@ export default function FileDetails(props: FileDetails) {
...
+                        className={classNames(styles.fileThumbnailContainer, styles.noThumbnail, {
+                            [styles.thumbnailDefault]: windowState.state === WindowState.DEFAULT,
+                            [styles.thumbnailMaximized]:
+                                windowState.state === WindowState.MAXIMIZED,
+                        })}
+                    />
+                </div>
+            );
+        }
     }
```
yrkim98, line 144: @GabeMedrash Is this too verbose? I initially created another if else statement after the getting the thumbnailURI to create the thumbnail, but I thought this was cleaner since we don't have to save the thumbnailURI.

kmitcham: I dunno about Gabe, but I find this much more readable.  Remember, the person (maybe future you) who comes back to work on this latter may be stressed, up too late and under time pressure-clarity and ease of understanding are powerful.

GabeMedrash: Brian--not at all, I think the code communicates well. Here's one suggestion for a minor improvement (cuts out just a tiny bit of cyclomatic complexity): instead of initializing `thumbnail` to `null`, you could initialize it to the "no image" placeholder. 

Also, one more question for you: is the `div` that wraps the `svg` necessary? What's it doing?

yrkim98:
Should be all good now- I took out the div I was not using and I initialize the thumbnail as the default placeholder. Also added a check to see if a file is selected so that the placeholder doesnt show on startup.

GabeMedrash (approved):
Well done!

PR #21, opened 2022-08-29
[SWE-230] Use new "Uploaded" annotation querying contract in FEA

tyler-foster:
Small change to make it so that we can use the new FES API for querying against date/datetime annotations in the Explorer App, which will allow us to remove some code in FES. I was originally going to try and figure out a way to make this a smoother process on the FES side, but this ended up being the least messy solution I could come up with

Basically just replaces the old `YYYY-MM-DD-to-YYYY-MM-DD` filter with `RANGE(YYYY-MM-DD,YYYY-MM-DD)`

### DIFF
--- a/packages/core/components/FileMetadataSearchBar/index.tsx
+++ b/packages/core/components/FileMetadataSearchBar/index.tsx
@@ -28,7 +28,6 @@ const FILE_NAME_OPTION = FILE_ATTRIBUTE_OPTIONS.find(
 ) as IDropdownOption;
 // Color chosen from App.module.css
 const PURPLE_ICON_STYLE = { icon: { color: "#827aa3" } };
-export const DATE_RANGE_SEPARATOR = "-to-"; // Not arbitrary, defined per contract with FES
 
 // Because the datestring comes in as an ISO formatted date like 2021-01-02
 // creating a new date from that would result in a date displayed as the
@@ -41,6 +40,14 @@ export function extractDateFromDateString(dateString?: string): Date | undefined
     const date = new Date(dateString);
     date.setMinutes(date.getTimezoneOffset());
     return date;
+}
+
+function extractDatesFromRangeOperatorFilterString(filterString: string) {
+    // Regex with capture groups for identifying strings using the RANGE() operator with full ISO strings
+    // e.g. RANGE(2022-01-01T00:00:00.000Z,2022-01-31T00:00:00.000Z)
+    // Captures "2022-01-01T00:00:00.000Z" and "2022-01-31T00:00:00.000Z"
+    const RANGE_OPERATOR_REGEX = /RANGE\(([\d\-:TZ.]+),([\d\-:TZ.]+)\)/g;
+    return RANGE_OPERATOR_REGEX.exec(filterString);
 }
 
 /**
@@ -75,15 +82,20 @@ export default function FileMetadataSearchBar() {
     }
 
     function onDateRangeSelection(newDateRange: { startDate?: Date; endDate?: Date }) {
-        const [oldStartDate = undefined, oldEndDate = undefined] =
-            fileAttributeFilter?.value.split(DATE_RANGE_SEPARATOR) || [];
+        let oldStartDate;
+        let oldEndDate;
+        const splitFileAttributeFilter = extractDatesFromRangeOperatorFilterString(
+            fileAttributeFilter?.value
+        );
+        if (splitFileAttributeFilter !== null) {
+            oldStartDate = splitFileAttributeFilter[1];
+            oldEndDate = splitFileAttributeFilter[2];
+        }
         const startDate = newDateRange.startDate
-            ? newDateRange.startDate.toISOString().split("T")[0]
+            ? newDateRange.startDate.toISOString()
             : oldStartDate;
-        const endDate = newDateRange.endDate
-            ? newDateRange.endDate.toISOString().split("T")[0]
-            : oldEndDate;
-        onSearch(`${startDate || endDate}${DATE_RANGE_SEPARATOR}${endDate || startDate}`);
+        const endDate = newDateRange.endDate ? newDateRange.endDate.toISOString() : oldEndDate;
+        onSearch(`RANGE(${startDate || endDate},${endDate || startDate})`);
     }
 
     function onAttributeSelection(_: React.FormEvent, option?: IDropdownOption) {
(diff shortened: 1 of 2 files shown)


packages/core/components/FileMetadataSearchBar/index.tsx
```diff
@@ -43,6 +42,28 @@ export function extractDateFromDateString(dateString?: string): Date | undefined
...
+function extractDatesFromRangeOperatorFilterString(
+    filterString: string
+): { startDate: Date; endDate: Date } | null {
+    // Regex with capture groups for identifying ISO datestrings in the RANGE() filter operator
+    // e.g. RANGE(2022-01-01T00:00:00.000Z,2022-01-31T00:00:00.000Z)
+    // Captures "2022-01-01T00:00:00.000Z" and "2022-01-31T00:00:00.000Z"
+    const RANGE_OPERATOR_REGEX = /RANGE\(([\d\-:TZ.]+),([\d\-:TZ.]+)\)/g;
+    const exec = RANGE_OPERATOR_REGEX.exec(filterString);
+    if (exec && exec.length === 3) {
+        // Length of 3 because we use two capture groups
```
kmitcham, line 54: good comment.

packages/core/components/FileMetadataSearchBar/index.tsx
```diff
@@ -43,6 +42,28 @@ export function extractDateFromDateString(dateString?: string): Date | undefined
...
+        const startDate = new Date(exec[1]);
+        // The RANGE() filter uses an exclusive upper bound.
+        // However, we want to present dates in the UI as if the upper bound was inclusive.
+        // To handle that, we'll subtract a day from the upper bound used by the filter, then present the result
```
kmitcham, line 58: wait, is upper bound the most recent?  in dates, upper usually means more milliseconds.  Subtracting a day would be the wrong thing.

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
```diff
@@ -116,26 +113,30 @@ describe("<FileMetadataSearchBar />", () => {
         expect(getByText("Sun Aug 30 2020")).to.not.be.empty;
```
kmitcham, line 113: could we add checks of 2020-08-30, 2020-08-31, 2020-09-01, 2021-03-03,2021-03-04, 2021-03-05?  Just to pin down the edges being what we want.

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
```diff
@@ -116,26 +113,30 @@ describe("<FileMetadataSearchBar />", () => {
...
             </Provider>
         );
-        const day = 17;
-        const date = new Date();
-        const expectedDate = `${date.getFullYear()}-${("0" + (date.getMonth() + 1)).slice(
-            -2
-        )}-${day}`;
-        const expectedRange = `${expectedDate}${DATE_RANGE_SEPARATOR}${expectedDate}`;
+        const startDate = new Date();
+        const day = 15; // Arbitrary day of the month that isn't likely to render twice
```
kmitcham, line 125: I don't understand what is meant by 'render twice' here.

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
```diff
@@ -116,26 +113,30 @@ describe("<FileMetadataSearchBar />", () => {
...
+        startDate.setDate(day);
+        startDate.setHours(0);
+        startDate.setMinutes(0);
+        startDate.setSeconds(0);
+        startDate.setMilliseconds(0);
+        const endDate = new Date(startDate);
+        endDate.setDate(endDate.getDate() + 1);
```
kmitcham, line 132: so endDate is one day after start day, but in the old test, it was one MONTH later?  Just making sure I understand.

kmitcham (approved):

packages/core/components/FileMetadataSearchBar/index.tsx
```diff
@@ -74,16 +95,26 @@ export default function FileMetadataSearchBar() {
...
+            fileAttributeFilter?.value
+        );
+        if (splitFileAttributeFilter !== null) {
+            oldStartDate = splitFileAttributeFilter.startDate;
+            oldEndDate = splitFileAttributeFilter.endDate;
+        }
+
+        const newStartDate = startDate || oldStartDate || endDate;
+        const newEndDate = endDate || oldEndDate || startDate;
+        if (newStartDate && newEndDate) {
```
GabeMedrash, line 112: When would this `if` statement not evaluate to `true`?

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
GabeMedrash: I agree that it would be nice to have some additional test coverage around this logic. There are some existing patterns in the codebase related to test parameterization to avoid copy/pasting large tests just to change some inputs/expectations, let me know if you want a pointer to one of those.

packages/core/components/FileMetadataSearchBar/index.tsx
tyler-foster: So if I have a filter like `RANGE(day1, day2)`, "day1" is the lower bound, and "day2" is the upper bound. The upper bound in this query is exclusive, following the same logic as Python's built-in `range()` function (e.g. `list(range(0, 10))` generates a list of numbers starting with 0 and ending with 9).
However, in the File Explorer App, it would be unintuitive for users to select their exclusive upper bound via the UI we have. I wanted to allow users to pick an _inclusive_ upper bound in the UI, which means if we want to use this RANGE() filter in generating queries we have to do a little behind-the-scenes magic to make that work. So when they pick their upper bound, the code does a +1 to make the filter work, and when rendering the UI based off of an existing filter the code does a -1 to make the UI display what the user would expect

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
tyler-foster: Notice how in this picture of the calendar component, some dates are listed twice, e.g. "30" and "1". This can screw with the testing-library function
image

packages/core/components/FileMetadataSearchBar/index.tsx
tyler-foster: It shouldn't ever in practice... but the linter kept yelling at me since they could technically both be `null` if I called the function with both parameters being `null`...

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
tyler-foster: Parameterized this test and added a few cases to it

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
tyler-foster: I don't honestly know why it did that. The expectation in this test is (and was) for the start date and the end date to be the same in the UI if you only have a start date selected

packages/core/components/FileMetadataSearchBar/test/FileMetadataSearchBar.test.tsx
kmitcham: maybe "Show up twice in the calender popup"?

GabeMedrash (approved):

PR #22, opened 2022-09-02
Remove restriction on sorting on annotations

SeanLeRoy:
Sorting on annotations (non file-property data, so anything that isn't filename, fileid, filesize, etc.) has been disabled because performance when doing so was so slow it didn't seem worth it to allow it. However, with the recent performance upgrades the hope is that we can enable this feature. Testing against a local FES that was pointed at staging made it seem possible to sort on annotations without performance becoming unbearable.

### DIFF
--- a/packages/core/components/FileList/Header.module.css
+++ b/packages/core/components/FileList/Header.module.css
@@ -31,14 +31,6 @@
 }
 
 .header-cell {
-  cursor: not-allowed;
-}
-
-.header-cell-tooltip {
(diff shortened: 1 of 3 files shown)


kmitcham (approved):
Might want to comment some rough numbers for what 'unbearable' means as a breadcrumb for future work; if someone comes into this code and touches something, what performance should be expected (to an order of magnitude).

GabeMedrash (approved):

PR #23, opened 2022-09-07
Feature/swe 224 prevent unnecessary calls to files resource

SeanLeRoy:
The `<InfiniteLoader />` component has been requesting for a small batch of files upon the initial render of the `<FileList />` component (which represents the list of files seen in the UI). Nearly immediately after this request the height of the component is measured and given to the `<InfiniteLoader />` causing it to make another request for another batch of files that contains the original batch as a subset. This is unnecessary for two reasons:
1) Until the files request has completed, the `<InfiniteLoader />` is told that the original subset of files requested has not loaded (which is true), but what it is really asking is "should I, the component, ask for these files?" which is not necessarily the same. For example, say the batch of files initially fetched is from index 0 - 13 and the second batch is 0 - 53. The indexes 0 - 13 is entirely contained within the latter and do not need re-fetched, but we were essentially saying "even though those files will shortly be available, re-request them.". Instead, the component will now be told to ask for files that are both not being fetched and not fetched already. Note: reason/solution 2 solves this initial duplicate, but does not solve future ones caused by scrolling which is why both combined are helpful.
2) The component does not need to do that initial request at all, it can just wait until the height is measured or calculated. The measured height is used in the root `<FileList />` case which is the first view you see when you open the app as opposed to a `<FileList />` that is within a folder otherwise it uses a calculated height based on the total count of files represented by the list. The measured height calculation and re-render is very quick so just waiting that out doesn't seem to cause any delays or problems visually. The calculated height would likely delay the render if we waited for it before rendering the list so I gave it a default.

**Testing**
Checking out the network calls locally made to FES and comparing the startIndex, endIndex, limit, & offset in comparison to each other to see how much overlap and duplication is happening. A few (< 10) files may get re-fetched due to dynamic pagination and over-fetching, but for the most part I can confirm files are less likely to get fetched > 1 time. The initial fetch is also not present & it doesn't seem to have affected the visuals of a file list being rendered (i.e. I don't notice any additional delay).

**Note**
I did look into using a fixed page size when paging to try to further reduce our requests to the File Explorer Service (FES), but it started to look a lot like what we are already doing so likely not worth doing.

### DIFF
--- a/packages/core/components/FileList/index.tsx
+++ b/packages/core/components/FileList/index.tsx
@@ -16,7 +16,7 @@ import useFileAccessContextMenu from "./useFileAccessContextMenu";
 
 import styles from "./FileList.module.css";
 
-const DEBOUNCE_WAIT_FOR_DATA_FETCHING = 50; // ms
+const DEBOUNCE_WAIT_FOR_DATA_FETCHING = 100; // ms
 
 // This is an arbitrary value that needs to be set to some number > 0 in order to prompt react-window-infinite-loader
 // to start calling `loadMoreItems`.
@@ -64,7 +64,7 @@ export default function FileList(props: FileListProps) {
     >();
     const dataDrivenHeight = rowHeight * totalCount + 3 * rowHeight; // adding three additional rowHeights leaves room for the header + horz. scroll bar
     const calculatedHeight = Math.min(MAX_NON_ROOT_HEIGHT, dataDrivenHeight);
-    const height = isRoot ? measuredHeight : calculatedHeight;
+    const height = isRoot ? measuredHeight : calculatedHeight || MAX_NON_ROOT_HEIGHT;
 
     const listRef = React.useRef<FixedSizeList | null>(null);
     const outerRef = React.useRef<HTMLDivElement | null>(null);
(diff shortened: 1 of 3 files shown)


kmitcham (approved):

packages/core/components/FileList/index.tsx
```diff
@@ -104,47 +104,56 @@ export default function FileList(props: FileListProps) {
...
-                    }}
-                </InfiniteLoader>
+                {/* 
+                    When this component isRoot the height is measured. It takes
+                    a few milliseconds for that to happen along with a re-render, but
+                    by then the <InfiniteLoader /> used here will already have made a 
+                    request for files based on that height that it didn't need to
+                    if it had just waited until the measured height was present.
+                 */}
+                {height !== 0 && (
```
GabeMedrash, line 114: Doesn't the change on line 67 (`|| MAX_NON_ROOT_HEIGHT`) prevent this condition from ever being `false`?

GabeMedrash (approved):

packages/core/components/FileList/index.tsx
SeanLeRoy: No, at least that is not my intention. The default `|| MAX_NON_ROOT_HEIGHT` is used to give a default to the calculated height when it is not present, which at the moment is unlikely to happen, but with future work (like SWE-223) it is possible it would. Meanwhile the `height !== 0` case it to prevent rendering the `<InfiniteLoader />` before the measured height (which is different than the calculated height) is available; no default is needed for the measured height since the measurement & re-render happens within a few milliseconds..

hughes036 (approved):

PR #24, opened 2022-09-08
[SWE-223] Remove blocking totalCount() call when loading new <FileList>'s

tyler-foster:
[Ticket here](https://aicsjira.corp.alleninstitute.org/browse/SWE-223).
#### Context
See these folders in the File Explorer, e.g. AICS-0?
image
Right now, opening them up and seeing their file contents takes an extra long time. This isn't just because some of our Mongo queries are slow, but rather that we're waiting on queries that don't need to be waited on. The culprit is that little count of files in the bottom right of the file list. We first wait to figure out what that number should be, and _then_ we begin asking for files. Problem is, counting those files can take a long time on its own (think 10+ seconds), and we want people to be able to look at files instead of staring at loading icons.

Turns out we can shoot off both the getfiles() and the count() queries at (roughly) the same time, and that results in having a (roughly) twice as fast display of the `<FileList>` contents, which is what this ticket aims to accomplish. In combination with other ongoing work to improve the speed of the queries themselves, this should make the File Explorer App feel more responsive.

#### Changes
* Moved the query for determining `<FileList>`'s file count into an asynchronous function call that is triggered when the component mounts

#### Testing
I added one new test in [packages/core/components/FileList/test/FileList.test.tsx](https://github.com/AllenInstitute/aics-fms-file-explorer-app/compare/feature/SWE-223_make_count_calls_parallel?expand=1#diff-a9b35d6758685a6a6ba1654a62c0239e03629c07404b8209126e7ba23fff7fca) for ensuring the count() query is triggered when `<FileList>` mounts and properly updates state once it returns a value. Peeped at the "Network" tab in the app to get an idea of how much faster the file list renders are now

### DIFF
--- a/packages/core/components/DirectoryTree/useDirectoryHierarchy.tsx
+++ b/packages/core/components/DirectoryTree/useDirectoryHierarchy.tsx
@@ -129,28 +129,12 @@ const useDirectoryHierarchy = (
         async function getContent() {
             if (isLeaf || hierarchy.length === 0) {
                 // if we're at the top or bottom of the hierarchy, render a FileList
-                try {
-                    const totalCount = await fileSet.fetchTotalCount();
-                    if (!cancel) {
-                        dispatch(
-                            receiveContent(
-                                <FileList
-                                    fileSet={fileSet}
-                                    isRoot={isRoot}
-                                    sortOrder={sortOrder}
-                                    totalCount={totalCount}
-                                />
-                            )
-                        );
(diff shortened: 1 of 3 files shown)


GabeMedrash (approved):
Nice!

kmitcham (approved):

hughes036 (approved):
I like your 'storytelling' approach to this pr.

PR #25, opened 2022-09-13
Feature/swe 278 remove total count

SeanLeRoy:
[Per this work in the File Explorer Service](https://aicsbitbucket.corp.alleninstitute.org/projects/SW/repos/file-explorer-service/pull-requests/118/overview) the `totalCount` property and `hasMore` property should be avoided since calculating them requires extra work on the server that need not happen.

**Testing**
Tested by opening the application and running through the usual features of the application to see if they behave as expected as well as double checking the responses look correct.

### DIFF
--- a/packages/core/entity/RestServiceResponse/index.ts
+++ b/packages/core/entity/RestServiceResponse/index.ts
@@ -2,7 +2,6 @@ export interface Response<T> {
     data: T[];
     offset: number;
     responseType: string;
-    totalCount: number;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/core/entity/RestServiceResponse/index.ts
```diff
@@ -1,6 +1,5 @@
 export interface Response<T> {
     data: T[];
-    hasMore: boolean;
     offset: number;
     responseType: string;
     totalCount: number;
```
kmitcham, line 5: should this be deleted?

kmitcham (approved):

packages/core/entity/RestServiceResponse/index.ts
SeanLeRoy: Yes, thank you!

GabeMedrash (approved):
You might try re-triggering the Github Action for this build. Curious if the failure was a temporary thing or some issue with the build/certs that we need to look at.

PR #28, opened 2023-02-01
[SWE-706] Add payload as properties

SeanLeRoy:
This will cause the redux action payload to be spread across the event metrics as properties, hopefully informing us of specific annotation usage


SeanLeRoy:
Build is failing due to cert, looking into fix now.

tyler-foster (approved):

kmitcham (approved):

PR #31, opened 2023-02-10
Feature/swe 718 add uploaded views

SeanLeRoy:
This feature is meant to help provide A) an introduction to a feature we intended to implement since the origin of this app to test the waters & B) provide a simple way to narrow down the possible set of files to query against to improve performance. Ideally this feature would be moved forward to have ways to prompt the user to do this or provide suggestions.

I also added an empty search results page & some info tooltips for collections/views.

**Testing**
Tested locally and added some unit tests

https://user-images.githubusercontent.com/41307451/218176243-30326899-ace5-439e-8d13-739de8cf2225.mov

### DIFF
--- a/packages/core/App.tsx
+++ b/packages/core/App.tsx
@@ -1,5 +1,4 @@
 import "normalize.css";
-import classNames from "classnames";
 import { initializeIcons, loadTheme } from "@fluentui/react";
 import * as React from "react";
 import { batch, useDispatch, useSelector } from "react-redux";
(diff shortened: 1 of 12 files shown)


kmitcham (approved):

hughes036 (approved):
Looks like a great feature!  I feel like it may be too hidden though.  What if that tool bar was shown by default, but the user could opt to hide it once they acknowledge that they have seen it.  Also, what if the filters were buttons rather than drop down options.  

My suggestions come from realizing that users seem unaware of features that are considerably more visible than this (the url bar).

PR #42, opened 2023-03-03
Feature/swe 776 download feature upgrade

SeanLeRoy:
This adds a new "Download" option to the context menu present when you right-click a selection of files. In addition to this the download button in the file details is a bit more obvious and paired with an "Open" button (to hopefully encourage opening files through the app like this rather than downloading them). This was a UX request submitted as a user request, a pretty common bit of feedback we get that people can't find our download feature.

**Testing**
Tested manually and with some unit tests

### DIFF
--- a/packages/core/components/ContextMenu/items.ts
+++ b/packages/core/components/ContextMenu/items.ts
@@ -98,7 +98,7 @@ export default function getContextMenuItems(dispatch: Dispatch) {
                 text: "Download",
                 title: "Download selected files",
                 onClick() {
-                    dispatch(interaction.actions.downloadSelectedFiles());
+                    dispatch(interaction.actions.downloadFiles());
                 },
             },
         ],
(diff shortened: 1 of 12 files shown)


kmitcham (approved):

packages/core/components/ContextMenu/items.ts
```diff
@@ -92,6 +93,14 @@ export default function getContextMenuItems(dispatch: Dispatch) {
                     dispatch(interaction.actions.showManifestDownloadDialog());
                 },
             },
+            {
+                key: ContextMenuActions.DOWNLOAD,
+                text: "Download",
```
pgarrison, line 98: Tiny quality of life improvement if it's easy: we could do something like this here: `fileSelection.length > 1 ? "Download all" : "Download"`

pgarrison (approved):
Love it

packages/core/components/ContextMenu/items.ts
SeanLeRoy: For the context menu stuff we have been just following what file browsers (like MacOS Finder or even Google Drive does) and I think they don't make a distinction for downloading one or multiple items. If we did update it I feel like not also appending "all" to "Open" & "Open with" might be misleading potentially? Let me know your thoughts!

Here is Google Drive for example:
Screen Shot 2023-03-07 at 11 33 19 AM

pgarrison: Borrowing from the Apple/Google UX teams seems like a great strategy! Let's keep it

PR #47, opened 2023-04-27
Feature/swe 775 plate UI integration

yrkim98:
Adding Open With > PlateUI option to quickly see plate information
https://aicsjira.corp.alleninstitute.org/browse/UR-92

### DIFF
--- a/packages/core/components/FileList/useFileAccessContextMenu.ts
+++ b/packages/core/components/FileList/useFileAccessContextMenu.ts
@@ -5,7 +5,8 @@ import FileFilter from "../../entity/FileFilter";
 import { interaction, selection } from "../../state";
 import { ContextMenuItem } from "../ContextMenu";
 import getContextMenuItems, { ContextMenuActions } from "../ContextMenu/items";
-import { FileExplorerServiceBaseUrl } from "../../constants";
+import { FmsFile } from "../../services/FileService";
+import { AnnotationName, FileExplorerServiceBaseUrl } from "../../constants";
 
 /**
  * Custom React hook for creating the file access context menu.
@@ -21,15 +22,17 @@ export default function useFileAccessContextMenu(filters?: FileFilter[], onDismi
     const { executionEnvService } = useSelector(interaction.selectors.getPlatformDependentServices);
     const [plateLink, setPlateLink] = React.useState("");
 
-    fileSelection.fetchAllDetails().then((fileDetails) => {
+    fileSelection.fetchFocusedItemDetails().then((fileDetails: FmsFile | undefined) => {
+        if (!fileDetails) return;
         // Grabbing plate barcode
-        const platebarcode = fileDetails[0].annotations.find((x) => x.name === "Plate Barcode");
-        //If theres a barcode make plateUI option available
+        const platebarcode = fileDetails.annotations.find(
+            (x) => x.name === AnnotationName.PLATE_BARCODE
+        );
+        // If there's a barcode, make plateUI option available
         if (platebarcode?.values) {
             const barcode: string = platebarcode.values[0].toString();
-            setPlateLink(createPlateLink(global.fileExplorerServiceBaseUrl, barcode));
+            setPlateLink(createPlateLink(barcode));
         }
-        return;
     });
 
     return React.useCallback(
@@ -134,10 +137,9 @@ export default function useFileAccessContextMenu(filters?: FileFilter[], onDismi
     );
 }
 
-function createPlateLink(env: string, barcode: string): string {
-    if (env === FileExplorerServiceBaseUrl.STAGING) {
-        return `http://stg-aics.corp.alleninstitute.org/labkey/aics_microscopy/AICS/editPlate.view?Barcode=${barcode}`;
-    } else {
-        return `http://aics.corp.alleninstitute.org/labkey/aics_microscopy/AICS/editPlate.view?Barcode=${barcode}`;
-    }
+function createPlateLink(barcode: string): string {
+    // fileExplorerServiceBaseUrl may not be defined: default is Production LabKey
+    const baseURL = global.fileExplorerServiceBaseUrl || FileExplorerServiceBaseUrl.PRODUCTION;
+    const baseURLHttp = baseURL.replace("https", "http"); // LabKey does not support HTTPS yet
+    return `${baseURLHttp}/labkey/aics_microscopy/AICS/editPlate.view?Barcode=${barcode}}`;
 }
--- a/packages/core/components/FileList/useFileAccessContextMenu.ts
+++ b/packages/core/components/FileList/useFileAccessContextMenu.ts
@@ -27,7 +27,7 @@ export default function useFileAccessContextMenu(filters?: FileFilter[], onDismi
         const platebarcode = fileDetails[0].annotations.find((x) => x.name === "Plate Barcode");
 
         //If theres a barcode make plateUI option available
-        if (platebarcode != undefined && platebarcode.values != undefined) {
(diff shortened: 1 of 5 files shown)


packages/core/components/FileList/useFileAccessContextMenu.ts
```diff
@@ -19,6 +21,46 @@ export default function useFileAccessContextMenu(filters?: FileFilter[], onDismi
     const userSelectedApplications = useSelector(interaction.selectors.getUserSelectedApplications);
     const { executionEnvService } = useSelector(interaction.selectors.getPlatformDependentServices);
 
+    const [hasPlateBarCode, setHasPlateBarCode] = React.useState(false);
+    let plateUILink: string | undefined = undefined;
+
+    fileSelection.fetchAllDetails().then((fileDetails) => {
+        console.log(fileDetails);
```
SeanLeRoy: Generally we remove logs like this

packages/core/components/FileList/useFileAccessContextMenu.ts
```diff
@@ -19,6 +21,46 @@ export default function useFileAccessContextMenu(filters?: FileFilter[], onDismi
...
+        const annotations = fileDetails[0].annotations;
+        const platebarcode = annotations.find((x) => x.name === "Plate Barcode");
```
SeanLeRoy: This should be a constant defined [here (you'd need to add it to it first)](https://github.com/AllenInstitute/aics-fms-file-explorer-app/blob/master/packages/core/constants/index.ts#L16)

packages/core/components/FileList/useFileAccessContextMenu.ts
```diff
@@ -19,6 +21,46 @@ export default function useFileAccessContextMenu(filters?: FileFilter[], onDismi
...
+        if (platebarcode != undefined && platebarcode.values != undefined) {
```
SeanLeRoy, line 30: Simpler using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)
```suggestion
        if (!platebarcode?.values) {
```

packages/core/components/FileList/useFileAccessContextMenu.ts
```diff
@@ -19,6 +21,46 @@ export default function useFileAccessContextMenu(filters?: FileFilter[], onDismi
...
+            console.log("has barcode");
```
SeanLeRoy: These can/should be removed too (also line 36)

packages/core/components/FileList/useFileAccessContextMenu.ts
```diff
@@ -19,6 +21,46 @@ export default function useFileAccessContextMenu(filters?: FileFilter[], onDismi
     const userSelectedApplications = useSelector(interaction.selectors.getUserSelectedApplications);
     const { executionEnvService } = useSelector(interaction.selectors.getPlatformDependentServices);
 
+    const [hasPlateBarCode, setHasPlateBarCode] = React.useState(false);
```
SeanLeRoy, line 22: Rather than have two variables for similar pieces of information consider converting this to `plateUILink` and consider an `undefined` `plateUILink` to be a sentinel value representing not having a plate barcode for disabling the menu item on line 103

SeanLeRoy (changes requested):

kmitcham (approved):

packages/core/components/FileList/useFileAccessContextMenu.ts
yrkim98: was fixed in [my prev. commit](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/47/commits/0b411a367e1e160e56602ee1f1e5466ef5ceb085)

packages/core/components/FileList/useFileAccessContextMenu.ts
yrkim98: not understanding, Is this refering to the plateUI links?? or something else

packages/core/components/FileList/useFileAccessContextMenu.ts
yrkim98: [This is the way i initially tried, but couldnt get working](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/47/commits/141ffe9d781553348a903a66a2fb08a6cee09822).

I have to use `useState` so that the view re-renders- but`setPlateLink` never sets `plateLink` when called- am I missing something?

packages/core/components/FileList/useFileAccessContextMenu.ts
```diff
@@ -18,6 +19,28 @@ export default function useFileAccessContextMenu(filters?: FileFilter[], onDismi
...
+            const barcode = platebarcode?.values[0];
+            if (global.fileExplorerServiceBaseUrl === FileExplorerServiceBaseUrl.STAGING) {
+                setPlateLink(
+                    `http://stg-aics.corp.alleninstitute.org/labkey/aics_microscopy/AICS/editPlate.view?Barcode=${barcode}`
+                );
+            } else if (
+                global.fileExplorerServiceBaseUrl === FileExplorerServiceBaseUrl.PRODUCTION
+            ) {
+                setPlateLink(
+                    `https://aics.corp.alleninstitute.org/labkey/aics_assays/AICS/Assays/plateDetail.view?Barcode=${barcode}`
```
tyler-foster, line 38: It looks like if the user is on staging, they get linked to the `aics_microscopy` version of the plate UI, but if they're on prod, they get linked to the `aics_assays` one. Maybe the plate link should be generated by its own function, something like:

 ```
function createBarcodeLink(env, barcode) {
 // and the middle part of this could even be declared as a constant somewhere
 return `http://{env}.corp.alleninstitute.org/labkey/aics_microscopy/AICS/editPlate.view?Barcode=${barcode}`
}
```
Oh, and maybe double-check that the `http` vs `https` part is consistent in both

File: packages/core/components/FileList/useFileAccessContextMenu.ts (hunk 2 above)
Comment (tyler-foster) [reply packages/core/components/FileList/useFileAccessContextMenu.ts]: I believe SEan means the "Plate Barcode" part should be a constant. So the change would look like `x.name === AnnotationName.PLATE_BARCODE` after you add `PLATE_BARCODE` to the constants file

**Comment by pgarrison** [comment]
LabKey module for viewing plates needs to be updated to accommodate selecting a plate by barcode instead of Plate ID. Relevant code [here](https://aicsbitbucket.corp.alleninstitute.org/projects/SW/repos/labkey-aics-microscopy/browse/resources/web/aics_microscopy/react_apps/plates/plate/state/form/logics.js#718-720)

File: packages/core/components/FileList/useFileAccessContextMenu.ts (hunk 6 above)
Comment (SeanLeRoy) [reply packages/core/components/FileList/useFileAccessContextMenu.ts]: Rather than know about the hosts at all you can pull this directly from the state [here](https://github.com/AllenInstitute/aics-fms-file-explorer-app/blob/master/packages/core/state/interaction/selectors.ts#L18).
```
const fileExplorerBaseUrl = useSelector(interactionSelectors.getFileExplorerBaseUrl);
const plateUIRef = `${fileExplorerBaseUrl}/labkey/aics_microscopy/AICS/editPlate.view?Barcode=${barcode}`
```

packages/core/components/FileList/useFileAccessContextMenu.ts
SeanLeRoy: I'm not seeing an issue in the commit you linked, that is what I was suggesting and should work

SeanLeRoy (approved):

PR #71, opened 2024-03-20
Feature/standardize property filters

### LINKED ISSUE #54: Make file property filters more normal
The file properties like "Uploaded" are treated different... should it be??

Instead, we could migrate the file properties into the main annotations list and have their filter modules slot into the regular annotations filter popover.

aswallace:
## Description of Changes

This moves the file properties that were previously in the FileMetadataSearchBar into the filter side bar to make their behavior/appearance more consistent with the other annotation filters. 3 new filter form types are also added: date range, regular numerical range, and a search box. The search box is primarily intended for text-type annotations that can't provide selectable values from the file explorer API (e.g., file name, file path). I also left the ListPicker component in place for numerical/date annotations where there are few enough options that it's not significantly easier to use a range.

Also removed the FileMetadataSearchBar from the UI and deleted the component + its tests

## How to Review

Aside from the new form components, the changes are mostly to annotation and filter logic -- mainly, when file properties are included as part of the annotation lists and when they aren't. My main concern is how extensible this solution will be (especially for Sean's csv work), since it's still pretty specific and hardcoded to handling file properties. 

## Testing

Tested manually and with unit tests. 

## Screenshots of new UI components
Search box filter
Search box filter

Numerical range filter
Numerical range filter

Date range filter
Date range filter

closes #54


packages/core/components/DateRangePicker/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+ * This component renders a simple form for selecting a minimum and maximum date range
+ */
+export default function DateRangePicker(props: DateRangePickerProps) {
+    const { onSearch, onReset, currentRange } = props;
+
+    function onDateRangeSelection(startDate: Date | null, endDate: Date | null) {
+        // Derive previous startDate/endDate from current filter state, if possible
+        let oldStartDate;
+        let oldEndDate;
+        const splitFileAttributeFilter = extractDatesFromRangeOperatorFilterString(
```
SeanLeRoy, line 41: Looks like each time this function is used it then splits the dates into a start and end. Seems like a simpler pattern would be to have the function return undefined values for the start and end date keys so that you can always use object destructuring to yank the start and end date values our and then do any processing you need to on those dates.

packages/core/components/FileMetadataSearchBar/index.tsx
SeanLeRoy, line 1: Is this file unused now?

packages/core/components/RangePicker/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+ * It is best suited for selecting items that are numbers.
+ */
+export default function RangePicker(props: RangePickerProps) {
+    const { errorMessage, items, loading, onSearch, onReset, currentRange, units } = props;
+
+    const overallMin = React.useMemo(() => {
+        return items[0]?.displayValue.toString() ?? "";
+    }, [items]);
+    const overallMax = React.useMemo(() => {
+        return items.at(-1)?.displayValue.toString() ?? "";
```
SeanLeRoy, line 40: Should this be +1 due to max being exclusive?

packages/core/components/RangePicker/index.tsx
SeanLeRoy, line 1: This forms may be best put into the `AnnotationFilterForm` folder

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -5,8 +5,12 @@ import { useDispatch, useSelector } from "react-redux";
 import { AnnotationType } from "../../entity/AnnotationFormatter";
 import FileFilter from "../../entity/FileFilter";
 import ListPicker, { ListItem } from "../../components/ListPicker";
+import RangePicker from "../../components/RangePicker";
```
SeanLeRoy, line 8: strange import

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -5,8 +5,12 @@ import { useDispatch, useSelector } from "react-redux";
...
+import SearchBoxForm from "../SearchBoxForm";
+import DateRangePicker from "../DateRangePicker";
 import { interaction, metadata, selection } from "../../state";
 import useAnnotationValues from "./useAnnotationValues";
+import { Spinner, SpinnerSize } from "@fluentui/react";
```
SeanLeRoy, line 13: This should be at the top with other third party imports (alphabetical)

packages/core/components/SearchBoxForm/index.tsx
SeanLeRoy, line 1: Should this one have tests similar to the RangePicker and DateRangePicker?

packages/core/components/RangePicker/index.tsx
SeanLeRoy, line 1: For consistency it seems this should be named `NumberRangePicker` IMO

SeanLeRoy:
So excited for this!

>My main concern is how extensible this solution will be (especially for Sean's csv work), since it's still pretty specific and hardcoded to handling file properties.

This seems fine! I think what would be ideal here is to have this get approved and merged into `main` then I'll build a branch off this work that wraps up the conversion of file properties into generic annotations. Generally we should be okay special casing the display of top level file properties and their filtering options, but should opt towards making their filtering and overall inclusion to the `metadata` blocks of state as blended as possible.

packages/core/components/FileMetadataSearchBar/index.tsx
aswallace: Yep! Was still there since I wasn't sure if we were ready to completely get rid of this component yet

packages/core/components/RangePicker/index.tsx
aswallace: re meeting yesterday, leaving this exclusive and making it clearer in UI that it is

packages/core/components/RangePicker/index.tsx
aswallace: I agree. Only thing is that the ListPicker gets used in other components, so may be trading some consistency

packages/core/components/SearchBoxForm/index.tsx
aswallace: added in [2d7c92f](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/71/commits/2d7c92f6bddf5f1aa2520aa6b3ffdda4ae015352)

packages/core/components/FileMetadataSearchBar/index.tsx
aswallace: Removed in [fa23f0f](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/71/commits/fa23f0f6c1956f2e54ee7eacb7e2bdc59fee28be)

packages/core/components/DateRangePicker/index.tsx
aswallace: updated in [e18657f](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/71/commits/e18657f1d8fc7e7a7a585dfbb8e3110f9dbe5eb6)

SeanLeRoy (approved):

packages/core/entity/AnnotationFormatter/test/formatters.test.ts
```diff
@@ -79,8 +85,14 @@ describe("Annotation formatters", () => {
 
             // no colon in UTC offset
             { input: "2018-05-24T00:00:00+0000", expected: "2018-05-24" },
-        ];
 
+            // range format
+            {
+                input: "RANGE(2018-05-24T00:00:00+0000,2019-05-26T00:00:00+0000)",
+                expected: "2018-05-24, 2019-05-26",
```
pgarrison, line 92: Love this addition to the tests

packages/core/state/metadata/selectors.ts
```diff
@@ -15,6 +15,12 @@ export const getSortedAnnotations = createSelector(getAnnotations, (annotations:
     Annotation.sort(annotations)
 );
 
+export const getSupportedAnnotations = createSelector(
+    getSortedAnnotations,
+    // TODO: Modify selector to be less case by case
```
pgarrison, line 20: I'm afraid I don't understand this TODO

packages/core/components/NumberRangePicker/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    }, [items]);
+    const overallMax = React.useMemo(() => {
+        return items.at(-1)?.displayValue.toString() ?? "";
+    }, [items]);
+
+    const [searchMinValue, setSearchMinValue] = React.useState(
+        extractValuesFromRangeOperatorFilterString(currentRange?.value).minValue ?? overallMin
+    );
+    const [searchMaxValue, setSearchMaxValue] = React.useState(
+        extractValuesFromRangeOperatorFilterString(currentRange?.value).maxValue ?? overallMax
```
pgarrison, line 47: Seems like it would be nice if FileFilter could store a `Range` object with `min` and `max` fields instead of just a string, but that's probably beyond the scope of this PR

packages/core/components/NumberRangePicker/NumberRangePicker.module.css
```diff
@@ -0,0 +1,66 @@
...
+label,input{
+    display: block;
+}
+
+input::-webkit-outer-spin-button,
+input::-webkit-inner-spin-button {
+    /* display: none; <- Crashes Chrome on hover */
+    -webkit-appearance: none;
+    margin: 0; /* <-- Apparently some margin are still there even though it's hidden */
+}
```
pgarrison, line 66: Would be great to see screenshots to review changes with substantial CSS work

packages/core/components/FilterDisplayBar/FilterMedallion.tsx
```diff
@@ -63,7 +63,9 @@ export default function FilterMedallion(props: Props) {
         }
     }, [textRef, filters]);
 
-    const operator = filters.length > 1 ? "ONE OF" : "EQUALS";
+    const firstFilterValue = filters[0].value.toString();
+    const operator =
+        filters.length > 1 ? "ONE OF" : firstFilterValue.includes("RANGE") ? "BETWEEN" : "EQUALS";
```
pgarrison, line 68: I love the ternary operator but nesting it feels like it's pushing the limits of readability. (FilterList too)

packages/core/components/DateRangePicker/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+// Because the datestring comes in as an ISO formatted date like 2021-01-02
+// creating a new date from that would result in a date displayed as the
+// day before due to the UTC offset, to account for this we can add in the offset
+// ahead of time.
+export function extractDateFromDateString(dateString?: string): Date | undefined {
+    if (!dateString) {
+        return undefined;
+    }
+    const date = new Date(dateString);
+    date.setMinutes(date.getTimezoneOffset());
```
pgarrison, line 27: 👍

pgarrison (approved):
Looks good!

packages/core/components/NumberRangePicker/NumberRangePicker.module.css
aswallace: Will add! In this case though this is just re-use of other css modules elsewhere in the codebase

aswallace: Added screenshots to PR description!

PR #75, opened 2024-04-04
Feature/persistant columns

### LINKED ISSUE #57: Re-use columns selected last time

BrianWhitneyAI:
The purpose of this PR is to resolve #57 and introduce persistent columns (Display Annotations) for users after closing the app.

### DIFF
--- a/packages/core/state/index.ts
+++ b/packages/core/state/index.ts
@@ -8,6 +8,7 @@ import { PersistedConfig, PersistedConfigKeys } from "../services/PersistentConf
 import interaction, { InteractionStateBranch } from "./interaction";
 import metadata, { MetadataStateBranch } from "./metadata";
 import selection, { SelectionStateBranch } from "./selection";
+import Annotation from "../entity/Annotation";
 
 export { interaction, metadata, selection };
 
@@ -62,6 +63,13 @@ export function createReduxStore(options: CreateStoreOptions = {}) {
             csvColumns: persistedConfig && persistedConfig[PersistedConfigKeys.CsvColumns],
             userSelectedApplications:
                 persistedConfig && persistedConfig[PersistedConfigKeys.UserSelectedApplications],
+        },
+        selection: {
+            displayAnnotations:
+                persistedConfig &&
+                persistedConfig[PersistedConfigKeys.DisplayAnnotations]?.map(
+                    (annotation) => new Annotation(annotation)
+                ),
         },
     });
     return configureStore<State>({
--- a/packages/core/state/metadata/logics.ts
+++ b/packages/core/state/metadata/logics.ts
@@ -20,6 +20,7 @@ const requestAnnotations = createLogic({
         const { getState, httpClient } = deps;
         const applicationVersion = interaction.selectors.getApplicationVersion(getState());
         const baseUrl = interaction.selectors.getFileExplorerServiceBaseUrl(getState());
+        const displayAnnotations = selection.selectors.getAnnotationsToDisplay(getState());
         const annotationService = new AnnotationService({
             applicationVersion,
             baseUrl,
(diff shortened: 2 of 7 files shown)


packages/core/state/index.ts
```diff
@@ -8,6 +8,7 @@ import { PersistedConfig, PersistedConfigKeys } from "../services/PersistentConf
 import interaction, { InteractionStateBranch } from "./interaction";
 import metadata, { MetadataStateBranch } from "./metadata";
 import selection, { SelectionStateBranch } from "./selection";
+import Annotation from "../entity/Annotation";
```
SeanLeRoy, line 11: This import belongs with the above grouping (so as a new line above line 6)

packages/core/state/metadata/logics.ts
```diff
@@ -20,14 +21,24 @@ const requestAnnotations = createLogic({
...
+        let annotations: Annotation[] = [];
+
         try {
-            const annotations = await annotationService.fetchAnnotations();
+            annotations = await annotationService.fetchAnnotations();
+        } catch (err) {
+            console.error("Failed to fetch annotations", err);
+        }
+
+        dispatch(receiveAnnotations(annotations));
```
SeanLeRoy, line 39: We don't want to dispatch this if we don't have `annotations` & the defaults wouldn't be able to render anyway without them. In that case it seems like it would be best to have the catch statement look like:
```
            console.error("Failed to fetch annotations", err);
            done();
            return;
```

and the `dispatch(receiveAnnotations(annotations));` should move into the try block as the last line in it.

packages/core/state/selection/actions.ts
```diff
@@ -157,15 +157,19 @@ export const DESELECT_DISPLAY_ANNOTATION = makeConstant(
 );
 
 export interface DeselectDisplayAnnotationAction {
-    payload: Annotation | Annotation[];
+    payload: {
```
SeanLeRoy, line 160: ? why

packages/desktop/src/renderer/index.tsx
```diff
@@ -73,16 +73,22 @@ const store = createReduxStore({
...
+            type: annotation.type,
+        })),
         [PersistedConfigKeys.UserSelectedApplications]: userSelectedApplications,
     };
     if (JSON.stringify(appState) !== JSON.stringify(persistentConfigService.getAll())) {
         persistentConfigService.persist({
-            [PersistedConfigKeys.CsvColumns]: csvColumns,
-            [PersistedConfigKeys.UserSelectedApplications]: userSelectedApplications,
+            ...appState,
             [PersistedConfigKeys.HasUsedApplicationBefore]: hasUsedApplicationBefore,
```
SeanLeRoy, line 92: Seems like this should also be in the `appState` object - this isn't because of your changeset just seems buggy because otherwise isn't this always true?
`JSON.stringify(appState) !== JSON.stringify(persistentConfigService.getAll())`
If you agree I'd put `[PersistedConfigKeys.HasUsedApplicationBefore]: true` into the `appState` object.

packages/core/state/selection/actions.ts
BrianWhitneyAI: I thought this made it a little easer to read mirroring the SelectDisplayAnnotationAction so that both have action.payload.annotation as the annotation rather than action.payload for one and action.payload.annotation for another

packages/core/state/selection/reducer.ts
```diff
@@ -123,9 +123,9 @@ export default makeReducer<SelectionStateBranch>(
             sortColumn: action.payload,
         }),
         [DESELECT_DISPLAY_ANNOTATION]: (state, action) => {
-            const displayAnnotations = without(
```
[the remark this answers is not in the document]
BrianWhitneyAI: this was the source of the reselect bug I was talking about, for some reason the annotations returned by the persist are slightly different and this dont get removed with the `without` function. The same is true for comparing the objects directly. This is what i was saying about not liking my solution as much.

packages/desktop/src/renderer/index.tsx
BrianWhitneyAI: resolved in [f3b21ab](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/75/commits/f3b21abbf790b9287163be67ba25e771407927e0)

packages/core/state/metadata/logics.ts
BrianWhitneyAI: resolved in [f3b21ab](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/75/commits/f3b21abbf790b9287163be67ba25e771407927e0)

SeanLeRoy (approved):

packages/core/state/index.ts
BrianWhitneyAI: Moving this import causes the app to go blank. I am completely lost on why.

aswallace: Pulled locally and tried this as well, also caused a load error for me when put anywhere from line 8 up:
`Uncaught ReferenceError: Cannot access 'Annotation' before initialization`

packages/core/state/selection/reducer.ts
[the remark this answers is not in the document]
aswallace: Looks like lodash's `without` function doesn't support this kind of object comparison -- unless the objects share a reference, they aren't considered equal. I'm curious if this use of `without` worked when the payload to `deselectDisplayAnnotation` was previously just `annotation` instead of `{annotation}`?
But also am fine with the use of `filter` here

packages/core/state/metadata/logics.ts
```diff
@@ -20,14 +21,25 @@ const requestAnnotations = createLogic({
...
-            const annotations = await annotationService.fetchAnnotations();
+            annotations = await annotationService.fetchAnnotations();
+            dispatch(receiveAnnotations(annotations));
+        } catch (err) {
+            console.error("Failed to fetch annotations", err);
+            done();
+            return;
+        }
+
+        if (!displayAnnotations.length) {
```
aswallace, line 42: When I initially loaded this branch locally, got an error on load that caused the app to fail because `displayAnnotations` was set to `[undefined]` which technically has length 1. Brian and I were able to special case around it temporarily, but since then haven't been able to replicate the error. Wanted to check if anyone else (@SeanLeRoy ?) has seen this bug locally?

PR #76, opened 2024-04-12
feature/recent-hierarchy-annotations

### LINKED ISSUE #57: Re-use columns selected last time

BrianWhitneyAI:
The purpose of this PR is to resolve #57 and reuse hierarchy annotations from the previous session. I went about this by creating a persistent store of recently selected annotations. I set the limits of this store to be ordered from the most recent at the top, descending and limited to 7 stored unique elements.


BrianWhitneyAI:
I have noticed with the last two persistent state tickets that I have been working on that the comparison of annotation objects is difficult and inconsistent. In the future, we may want to add a method to `Annotation[]` that makes this easier.

BrianWhitneyAI:
Additionally, I think the `Annotation.sort` method is a little funky and could be reworked.

pgarrison:
How did you test this change?

packages/core/state/selection/reducer.ts
```diff
@@ -169,14 +171,33 @@ export default makeReducer<SelectionStateBranch>(
...
-        }),
+            // remove duplicates and limit to 7 most recent
+            const recentHierarchyAnnotations = [
+                ...new Map(
+                    recentHierarchyAnnotationsUnfiltered.map((annotation) => [
+                        annotation.name,
+                        annotation,
+                    ])
+                ).values(),
+            ].slice(0, 7);
```
pgarrison, line 189:
```suggestion
            const recentHierarchyAnnotations = lodash.uniqBy(recentHierarchyAnnotationsUnfiltered, 'name').slice(0, 7);
```
I think this can be simplified with a change like this.

pgarrison, line 189: Am I understanding that this will only preserve 7 columns? So if the user picks 8 columns the first one they picked will not be included when they re-open the explorer? Why not keep the entire last set of columns?

packages/desktop/src/services/PersistentConfigServiceElectron.ts
```diff
@@ -42,12 +42,31 @@ const OPTIONS: Options<Record<string, unknown>> = {
...
+        [PersistedConfigKeys.HasUsedApplicationBefore]: {
+            type: "boolean",
+        },
         // ImageJExecutable is Deprecated
         [PersistedConfigKeys.ImageJExecutable]: {
             type: "string",
         },
-        [PersistedConfigKeys.HasUsedApplicationBefore]: {
-            type: "boolean",
+        // ImageJExecutable is Deprecated
```
pgarrison, line 53: Stray comment copied

packages/core/state/metadata/selectors.ts
```diff
@@ -9,14 +9,29 @@ import { Dataset } from "../../services/DatasetService";
...
+    (recentAnnotations, annotations) => {
+        const unfilteredSupportedAnnotations = [
+            ...SEARCHABLE_TOP_LEVEL_FILE_ANNOTATIONS,
+            ...recentAnnotations,
+            ...annotations,
+        ];
+
+        return [
+            ...new Map(
+                unfilteredSupportedAnnotations.map((annotation) => [annotation.name, annotation])
```
aswallace, line 31: Having the same issue I had on the other persisted config branch where `recentHierarchyAnnotations` gets initialized to `[undefined]`. In this case, this line is where the error manifests & causes the app to error out on render

packages/core/state/metadata/selectors.ts
```diff
@@ -9,14 +9,29 @@ import { Dataset } from "../../services/DatasetService";
...
+export const getSupportedAnnotations = createSelector(
+    [getRecentHierarcyAnnotations, getSortedAnnotations],
+    (recentAnnotations, annotations) => {
+        const unfilteredSupportedAnnotations = [
+            ...SEARCHABLE_TOP_LEVEL_FILE_ANNOTATIONS,
+            ...recentAnnotations,
+            ...annotations,
+        ];
+
+        return [
```
SeanLeRoy, line 29: This looks like it removes the `Annotation.sort`

SeanLeRoy (changes requested):
I think this could be simplified by not creating a new field `recentHierarchyAnnotations` and instead using the existing field `annotationHierarchy`. The purpose of this feature is to preserve the hierarchy between application uses so simply persisting the `annotationHierarchy` list and then rehydrating it to the state upon application start would accomplish that without needing to create and track the hierarchy in a separate field.

packages/core/state/metadata/selectors.ts
BrianWhitneyAI: Yes. If I am interpreting this correctly, Annotation.sort would alphabetize the list which would defeat the purpose of having annotations close to the top.

BrianWhitneyAI:
Closing in favor of #81.

Issue #69, opened 2024-03-18
Create thumbnail view


SeanLeRoy:
Allow users to toggle between file row and thumbnail view in the explorer. Also, start finding thumbnails automatically for OME ZARRs using their top level config file to find the lowest resolution image

PR #77, opened 2024-04-15
Feature/create thumbnail view

aswallace:
### Description
This allows users to toggle between a list view and two different size thumbnail views in the explorer per #69.
The main new component is `LazilyRenderedThumbnail`, which follows the patterns set by `LazilyRenderedRow.`
The major logic changes are in `FileList` (the parent component) which contains the `InfiniteLoader` component. Note: `InfiniteLoader` is an external component that only supports lists by default, so needed to add in some customization to accommodate a grid.

### Testing
Tested manually and with new unit tests.
- Made sure the row/column to standard index conversion is consistent between view types and the right file stays selected between views
- Different views still work when grouping via the annotation hierarchy
- Using shift or ctrl to select still works as expected with all view types

### To do: 
There are a couple UX things that would be great to run by someone:
- What do we want the toggle & its icons to look like? Currently they're in the top right near the font size toggle
- Do we want custom icons for different file types (e.g., csv, tiff, czi ?) Right now they all use the same default "no image" icon

### Screenshots
Large size thumbnails
Large thumbnails

Small size thumbnails
Small thumbnails

Using annotation hierarchy to group
Screenshot 2024-04-15 at 11 43 19 AM

### DIFF
--- a/packages/core/components/FileDetails/index.tsx
+++ b/packages/core/components/FileDetails/index.tsx
@@ -252,7 +252,6 @@ export default function FileDetails(props: FileDetails) {
                     >
                         <Icon iconName="GridViewMedium"></Icon>
                     </ActionButton>
-
                     <ActionButton
                         className={classNames(styles.fontSizeButton, {
                             [styles.disabled]:
--- a/packages/core/components/FileDetails/index.tsx
+++ b/packages/core/components/FileDetails/index.tsx
@@ -1,4 +1,4 @@
-import { ActionButton, IButtonStyles } from "@fluentui/react";
+import { ActionButton, IButtonStyles, Icon } from "@fluentui/react";
 import classNames from "classnames";
 import * as React from "react";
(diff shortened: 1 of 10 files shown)


packages/core/components/FileDetails/index.tsx
```diff
@@ -226,6 +230,65 @@ export default function FileDetails(props: FileDetails) {
...
+                                selection.actions.setFileGridColumnCount(
+                                    THUMBNAIL_SIZE_TO_NUM_COLUMNS.LARGE
+                                )
+                            );
+                        }}
+                        title="Large thumbnail view"
+                    >
+                        <Icon iconName="GridViewMedium"></Icon>
+                    </ActionButton>
+
```
SeanLeRoy, line 255: (nit) empty line here

SeanLeRoy (approved):
This looks fantastic!! Great job 🎉

PR #79, opened 2024-04-24
Finish Integrating Multi-Data Source Feature + Revamp Style

SeanLeRoy:
**Description of Change**
This change is intended to be part II of enabling the explorer to read from a various data sources (CSVs, JSON, Parquet) - in this section I am largely just doing cosmetic changes or trying to improve/consolidate relevant code. Specifically what has happened is the "Annotation Sidebar/Hierarchy", "Filter bar", "Header", and "URL bar" have all been consolidated into the "QuerySidebar". This is beneficial by localizing interactions regarding an individiual query + removes app specific language like "Annotation Hierarchy" in favor "Group", "Filter", & "Sort". It enables us to allow users to have multiple queries that they can switch to.

**How to Review**
_I am happy to go through this changeset with anyone_, but for your own browsing:

Tons of this changeset is applying style changes / moving components around visually with the same functionality as it had before hence tons of deleting code and `.css` changes. I'd consider focusing on `/entity/` files then the `/components/<file_name>.tsx` files.

**Testing**
Tons of testing by manually going through each of the features in the desktop run version of the application + adding / modifying various unit tests. **Please let me know if you see tests you'd like or encounter a bad UX!**

**Before**
Screen Shot 2024-04-24 at 11 23 06 AM

**After**
new-gif

PR #81, opened 2024-05-02
feature/persistant-annotations

### LINKED ISSUE #57: Re-use columns selected last time

BrianWhitneyAI:
### Description of Changes 

This PR proposes prioritizing recently used Groups, Filters, and Sorts in the ordering of their respective selection dropdowns ( resolves #57 ). This PR is a continuation of #76  feature/recent-hierarchy-annotations , but is based on the new generalize-ui #79 branch.

### Testing 

Monitored state through various selections and deselections of Groups, Filters, and Sorts including those with repeated no-file-found options.

I did experience some issues with reloading the application where it would disconnect on the first reload but succeed on the second reload. I believe this is an inherent issue to the base branch or my development environment as is still present without these changes.

This PR also passes tests locally with the exception of 
```
  9) @renderer FileDownloadServiceElectron
       promptForDownloadDirectory
         complains about non-writeable directory when given:
     Error: Trying to stub property 'invoke' of undefined

```
I believe that this is unique to my local environment since I had to use WSL to run the full test suite. In reviewing this PR it would be super helpful to run tests locally with your dev env just to make sure everything is Ok.

### Possible Changes/Adjustments 

1) Limiting the length of recent annotations to a smaller number, I foresee a situation where a user has a lot of interaction and the recent list becomes very long (and unalphabetized) and this feature becomes a hindrance 
2) A adjustment to ‘On Start” state to not include undefined elements.

### DIFF
--- a/packages/core/state/selection/reducer.ts
+++ b/packages/core/state/selection/reducer.ts
@@ -99,8 +99,6 @@ export default makeReducer<SelectionStateBranch>(
             fileGridColumnCount: action.payload,
         }),
         [SET_FILE_FILTERS]: (state, action) => {
-            console.log("Filter: ", action.payload);
-
             const recentAnnotations = Array.from(
                 new Set([
                     // get annotaionName from each filter and add to recents
--- a/packages/core/state/selection/reducer.ts
+++ b/packages/core/state/selection/reducer.ts
@@ -1,5 +1,5 @@
 import { makeReducer } from "@aics/redux-utils";
-import { omit } from "lodash";
+import { castArray, omit } from "lodash";
 
 import interaction from "../interaction";
 import { THUMBNAIL_SIZE_TO_NUM_COLUMNS } from "../../constants";
@@ -49,6 +49,7 @@ export interface SelectionStateBranch {
     filters: FileFilter[];
     isDarkTheme: boolean;
     openFileFolders: FileFolder[];
+    recentAnnotations: string[];
     selectedQuery?: Query;
     shouldDisplaySmallFont: boolean;
     shouldDisplayThumbnailView: boolean;
@@ -73,6 +74,7 @@ export const initialState = {
     fileSelection: new FileSelection(),
     filters: [],
     openFileFolders: [],
+    recentAnnotations: [],
     shouldDisplaySmallFont: false,
     queries: [],
     shouldDisplayThumbnailView: false,
@@ -96,13 +98,26 @@ export default makeReducer<SelectionStateBranch>(
             ...state,
             fileGridColumnCount: action.payload,
         }),
-        [SET_FILE_FILTERS]: (state, action) => ({
-            ...state,
-            filters: action.payload,
+        [SET_FILE_FILTERS]: (state, action) => {
+            console.log("Filter: ", action.payload);
 
-            // Reset file selections when file filters change
-            fileSelection: new FileSelection(),
-        }),
+            const recentAnnotations = Array.from(
+                new Set([
+                    // get annotaionName from each filter and add to recents
+                    ...castArray(action.payload.map((filter: any) => filter.annotationName)),
+                    ...state.recentAnnotations,
+                ])
+            );
+            //let recentAnnotations: string[] = [];
+
+            return {
+                ...state,
+                filters: action.payload,
+                recentAnnotations,
+                // Reset file selections when file filters change
+                fileSelection: new FileSelection(),
+            };
+        },
         [SORT_COLUMN]: (state, action) => {
             if (state.sortColumn?.annotationName === action.payload) {
                 // If already sorting by this column descending
@@ -147,10 +162,16 @@ export default makeReducer<SelectionStateBranch>(
             ...state,
             queries: action.payload,
         }),
-        [SET_SORT_COLUMN]: (state, action) => ({
-            ...state,
-            sortColumn: action.payload,
-        }),
+        [SET_SORT_COLUMN]: (state, action) => {
+            const recentAnnotations = Array.from(
+                new Set([...castArray(action.payload.annotationName), ...state.recentAnnotations])
+            );
+            return {
+                ...state,
+                recentAnnotations,
+                sortColumn: action.payload,
+            };
+        },
         [interaction.actions.REFRESH]: (state) => ({
             ...state,
             availableAnnotationsForHierarchyLoading: true,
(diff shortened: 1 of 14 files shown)


packages/desktop/src/renderer/index.tsx
```diff
@@ -102,6 +104,10 @@ store.subscribe(() => {
         {}
     );
     if (JSON.stringify(appState) !== JSON.stringify(oldAppState)) {
+        if (appState.RECENT_ANNOTATIONS[0] == undefined) {
```
BrianWhitneyAI, line 107: This is the part I do not like. For some reason when a new user opens the app appstate.RECENT_ANNOTATIONS defaults to a list with one defined element rather than an empty list. This breaks when trying to persist the state because the typing does not match. While resetting on the first persist does work its hacky and would be better to fix at the source.

packages/core/state/selection/selectors.ts
```diff
@@ -76,3 +77,37 @@ export const getUnavailableAnnotationsForHierarchy = createSelector(
...
+        );
+
+        // get the File name annotaion in a list (if Present)
+        const fileNameAnnotation = annotations.filter(
+            (annotation) =>
+                annotation.name === AnnotationName.FILE_NAME || annotation.name === "File Name"
+        );
+
+        // combine all annotation lists
+        const combinedAnnotations = fileNameAnnotation.concat(
```
kmitcham, line 97: Making sure I understand -this results in:  FileName, (recent annotations if any, in original order), (all other annotations, sorted)

packages/core/state/selection/reducer.ts
```diff
@@ -96,13 +98,24 @@ export default makeReducer<SelectionStateBranch>(
...
-            filters: action.payload,
+        [SET_FILE_FILTERS]: (state, action) => {
+            const recentAnnotations = Array.from(
+                new Set([
+                    // get annotaionName from each filter and add to recents
+                    ...castArray(action.payload.map((filter: any) => filter.annotationName)),
+                    ...state.recentAnnotations,
+                ])
+            );
+            //let recentAnnotations: string[] = [];
```
SeanLeRoy, line 109: ?

packages/core/state/selection/reducer.ts
```diff
@@ -96,13 +98,24 @@ export default makeReducer<SelectionStateBranch>(
             ...state,
             fileGridColumnCount: action.payload,
         }),
-        [SET_FILE_FILTERS]: (state, action) => ({
-            ...state,
-            filters: action.payload,
+        [SET_FILE_FILTERS]: (state, action) => {
+            const recentAnnotations = Array.from(
```
SeanLeRoy, line 102: This could be simpler, also removes unnecessary castArray:
```
        [SET_FILE_FILTERS]: (state, action) => ({
            ...state,
            filters: action.payload,
            recentAnnotations: uniq([
                    ...action.payload.map((filter: any) => filter.annotationName),
                    ...state.recentAnnotations,
             ]),
            // Reset file selections when file filters change
            fileSelection: new FileSelection(),
       }),
```

packages/core/state/selection/reducer.ts
```diff
@@ -147,10 +160,16 @@ export default makeReducer<SelectionStateBranch>(
             ...state,
             queries: action.payload,
         }),
-        [SET_SORT_COLUMN]: (state, action) => ({
-            ...state,
-            sortColumn: action.payload,
-        }),
+        [SET_SORT_COLUMN]: (state, action) => {
```
SeanLeRoy, line 163: A bit simpler
```
        [SET_SORT_COLUMN]: (state, action) => ({
            ...state,
            recentAnnotations: uniq([...castArray(action.payload?.annotationName), ...state.recentAnnotations]),
            sortColumn: action.payload,
        }),
```

Also I think you'll need the optional chaining check that this adds because `action.payload` could be undefined

packages/core/state/selection/reducer.ts
```diff
@@ -175,14 +194,22 @@ export default makeReducer<SelectionStateBranch>(
             ...state,
             fileSelection: action.payload,
         }),
-        [SET_ANNOTATION_HIERARCHY]: (state, action) => ({
```
SeanLeRoy, line 178: A bit simpler (also removes unnecessary castArray)
```
        [SET_ANNOTATION_HIERARCHY]: (state, action) => ({
            ...state,
            annotationHierarchy: action.payload,
            availableAnnotationsForHierarchyLoading: true,
            recentAnnotations: uniq([...action.payload, ...state.recentAnnotations]),
            // Reset file selections when annotation hierarchy changes
            fileSelection: new FileSelection(),
        }),
```

packages/core/state/selection/selectors.ts
```diff
@@ -76,3 +77,37 @@ export const getUnavailableAnnotationsForHierarchy = createSelector(
             allAnnotations.filter((annotation) => !availableAnnotations.includes(annotation.name))
         )
 );
+
+// COMPOSED SELECTORS
+export const getSortedAnnotations = createSelector(
+    [getAnnotations, getRecentAnnotations],
+    (annotations: Annotation[], recentAnnotationNames: string[]) => {
```
SeanLeRoy, line 84: FWIW These typings are implicit and can be removed

packages/core/state/selection/selectors.ts
```diff
@@ -76,3 +77,37 @@ export const getUnavailableAnnotationsForHierarchy = createSelector(
...
+        // Create Array of annotations from recentAnnotationNames
+        const recentAnnotations = recentAnnotationNames.flatMap((name) =>
```
SeanLeRoy, line 86: This could be simpler by having this inner callback return an individual element rather than an array like so:
```
        const recentAnnotations = recentAnnotationNames.flatMap((name) =>
                annotations.find(annotation => annotation.name === name)
        )
        .filter(annotation => !!annotation);
```

packages/core/state/selection/selectors.ts
```diff
@@ -76,3 +77,37 @@ export const getUnavailableAnnotationsForHierarchy = createSelector(
...
+            annotations.filter((annotation) => annotation.name === name)
+        );
+
+        // get the File name annotaion in a list (if Present)
+        const fileNameAnnotation = annotations.filter(
```
SeanLeRoy, line 91: I think you are looking for `find` here which returns the first element that matches the condition found or `undefined` if none are found

SeanLeRoy: This should make sure this isn't undefined

packages/core/state/selection/selectors.ts
```diff
@@ -76,3 +77,37 @@ export const getUnavailableAnnotationsForHierarchy = createSelector(
...
+        );
+
+        // combine all annotation lists
+        const combinedAnnotations = fileNameAnnotation.concat(
+            recentAnnotations,
+            ...Annotation.sort(annotations)
+        );
+
+        // create map for filtering duplicate annotations.
+        const combinedAnnotationsMap = new Map();
```
SeanLeRoy, line 103: Could use reduce or similar function to build this map up FWIW

packages/desktop/src/renderer/index.tsx
SeanLeRoy: I still think we should find a fix for this rather than special case it if at all possible. My immediate hunch is this is because of the way you are grabbing the values from the persisted storage on load based on your description.
I believe we discussed this, but did you check to see if this:
```
    const recentAnnotations =
        persistedConfig && persistedConfig[PersistedConfigKeys.RecentAnnotations];
```
Could be causing an array like so `[undefined]`? Probably worth trying to see if:
```
    const recentAnnotations = persistedConfig?.[PersistedConfigKeys.RecentAnnotations]?.length
           ? persistedConfig?.[PersistedConfigKeys.RecentAnnotations]
           : [];
```
fixes it?

packages/core/state/selection/selectors.ts
```diff
@@ -76,3 +77,37 @@ export const getUnavailableAnnotationsForHierarchy = createSelector(
             allAnnotations.filter((annotation) => !availableAnnotations.includes(annotation.name))
         )
 );
+
+// COMPOSED SELECTORS
+export const getSortedAnnotations = createSelector(
```
SeanLeRoy, line 82: Tests around this selector would be good

SeanLeRoy (commented):
I think something this is missing is a visual indicator to the user that describes why these annotations are at the top of the list.
I think making it so the `AnnotationSelector` component does the special sort for recent annotations rather than the `getSortedAnnotations` would make it so that `AnnotationSelector` could then add some icon or something that would inform the user. Something I've seen before is an icon that looks like a redo icon on the list item, that could work well here for the recent annotations.
Screen Shot 2024-05-07 at 12 37 15 PM

PR #98, opened 2024-05-13
Feature/translate query to python snippet

### LINKED ISSUE #80: Translate query into Python Snippet
In the "Code Snippet" modal we would like to provide options for exporting the current query into programmatic queries. As an initial shot at this lets start with adding Python using Pandas. To do this:
In the modal:
* Add a dropdown for selecting which language to export the query to with just "Python (pandas)" as the one option for now
* Instead of TODO render the code it would take to query a pandas dataframe and get the same result as what would render in the file list of the app
* Test with various versions of pandas to get a minimum version to set for the pandas install

aswallace:
### Context 
In the Code Snippet section, provide the code that would allow the File Explorer query to be converted to Python (pandas)
Should close #80 

### Testing
Tested manually in Python 3.8-3.10 using the Variance Paper dataset. 
Works as is for Pandas 1.5-2.2. I tested manually down to Python 3.7 and pandas 1.3, but the `groupby` method changed after 1.5 so will throw an error for older versions. It's not a huge deprecation so if we wanted we could make it more dynamic or add a dropdown option for older Python/pandas versions

### To Dos: 
See #99 
Currently only works for external data sources & string type fields, which means it doesn't consider range filters and the sorting may not work as expected for numerical and date fields. 

### Screenshot
image


aswallace:
> Lets add the version floor to pandas to make sure it works for people / is explicit. I'd also include a note at the top that says this only works for Python version 3.7+

Will do! Also since we have that dropdown and the difference isn't substantial, I could theoretically differentiate options for Python 3.8+ pandas >= 1.5 and separate dropdown option for pandas < 1.5

aswallace:
> Lets add the version floor to pandas

Added in [34b3d35](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/98/commits/34b3d35cdd913a3da204ad30c31b42acf215b7df)

packages/core/entity/FileExplorerURL/index.ts
```diff
@@ -157,4 +157,79 @@ export default class FileExplorerURL {
...
+            filterQueryString = filterQueryString.concat(`.query('${value}')`);
+        });
+
+        const sortQueryString = urlComponents.sortColumn
+            ? this.convertSortToPython(urlComponents.sortColumn)
+            : "";
+        // const fuzzy = [] // TO DO: support fuzzy filtering
+
+        const hasQueryElements = groupByQueryString || filterQueryString || sortQueryString;
+        const imports = "import pandas\n";
```
BrianWhitneyAI, line 195: FWIW I think most people using python might be more familiar with `import pandas as pd`

packages/core/entity/FileExplorerURL/index.ts
```diff
@@ -157,4 +157,79 @@ export default class FileExplorerURL {
...
+        const comment = hasQueryElements ? "#Query on dataframe df" : "#No options selected";
+        const fullQueryString = `${comment}${
```
BrianWhitneyAI, line 197: We may want to set the output of the query to a variable for users:
```
import pandas
#Convert current datasource file to a pandas dataframe
df = pandas.read_csv('C:\Users\Brian\Desktop\Variance Paper Dataset.csv').astype('str')

#Query on dataframe df
df.groupby('Plate Barcode', group_keys=True).apply(lambda x: x)
```
```
import pandas
#Convert current datasource file to a pandas dataframe
df = pandas.read_csv('C:\Users\Brian\Desktop\Variance Paper Dataset.csv').astype('str')

#Query on dataframe df
df_queried= df.groupby('Plate Barcode', group_keys=True).apply(lambda x: x)
```

BrianWhitneyAI (approved):
I think that the code snippet is confusing when not using your own datasource. Having the option seems like it is supported for AICS FMS. Maybe include some logic that disables the code snippet option if the source url is undefined/not readable?

packages/core/entity/FileExplorerURL/index.ts
```diff
@@ -157,4 +157,79 @@ export default class FileExplorerURL {
...
+        return `\`${filter.name}\`=="${filter.value}"`;
+    }
+
+    private static convertCollectionToPython(collection: Collection | undefined) {
+        if (collection?.uri) {
+            const comment = "#Convert current datasource file to a pandas dataframe";
+            const extension = collection.uri.substring(collection.uri.lastIndexOf(".") + 1);
+            // Currently suggest setting all fields to strings; otherwise pandas assumes type conversions
+            // TO DO: Address different non-string type conversions
+            const code = `df = pandas.read_${extension}('${collection.uri}').astype('str')`;
```
BrianWhitneyAI, line 229: windows will need to denote the url as raw 
`df = pandas.read_csv(r'C:\Users\Brian\Desktop\Variance Paper Dataset.csv').astype('str')
`

SeanLeRoy (approved):
Resolving brian's suggestions then merging sounds good to me 👍

packages/core/entity/FileExplorerURL/index.ts
aswallace: updated in [4c46839](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/98/commits/4c46839ed84944cd4653718cf96ba1c93cd5d293)

packages/core/entity/FileExplorerURL/index.ts
aswallace: Added in [4c46839](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/98/commits/4c46839ed84944cd4653718cf96ba1c93cd5d293)

aswallace:
The only comment I didn't resolve was the windows part, and it's partially because with the new data source system I don't think I have access to the file paths of the data sources anymore. Unsure how best to address this

packages/core/entity/FileExplorerURL/index.ts
```diff
@@ -181,13 +185,15 @@ export default class FileExplorerURL {
...
+    private static convertDataSourceToPython(source: Source | undefined) {
+        if (source) {
+            const comment =
+                "#Convert current datasource file to a pandas dataframe\n" +
+                "#You may need to manually update the path to the file";
+
             // Currently suggest setting all fields to strings; otherwise pandas assumes type conversions
             // TO DO: Address different non-string type conversions
-            const code = `df = pandas.read_${extension}('${collection.uri}').astype('str')`;
+            const code = `df = pd.read_${source.type}('${source.name}').astype('str')`;
```
SeanLeRoy, line 196: I think you want `source.uri` here

SeanLeRoy: There is only no uri when we query against AICS FMS

aswallace: Fixed in [9b14578](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/98/commits/9b14578d20d689b15227e9c03bd0adba2f3fbf5e)

packages/core/entity/FileExplorerURL/index.ts
aswallace: Added OS check in [9b14578](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/98/commits/9b14578d20d689b15227e9c03bd0adba2f3fbf5e) and [afb7f3c](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/98/commits/afb7f3c5be3ed9974ad5d539e4aac3e5de823ce2)

PR #106, opened 2024-05-17
Make the explorer (BioFile Finder) deployable to the web

SeanLeRoy:
**Description**
This changeset adds the necessary platform specific services to enable the web version of the explorer (biofile finder). A large part of this is just moving things that were specific to a desktop format or were incompatible with how we could do things in the web into a format or workflow that works with both platforms. 

Some things were added that could/should have been added before this PR in retrospec like having the selected query mirror the query args and a couple style changes like making it so you can select any metadata tag as a column rather than just those that are "available".

**How to Review**
I recommend spending time checking out the `web/src/services` and the various `logics.tsx`.

**Testing**
I tested this manually using AICS FMS on desktop and on the web (localhost) and added various unit tests


aswallace:
Currently having trouble running the desktop version of the app on this branch but it may just be my local set up, would be great if someone else could also verify.

Seeing this error on render: `Uncaught TypeError: Cannot read properties of undefined (reading 'sortColumn')`

packages/core/components/QuerySidebar/tutorials/ShareView.ts
```diff
@@ -9,5 +9,5 @@ export const SHARE_VIEW_TUTORIAL = new Tutorial("Sharing current view (URL)")
     .addStep({
         targetId: Tutorial.ADD_QUERY_BUTTON_ID,
         message:
-            'Start a new query from a URL by clicking the New Query + button then "Import from URL" then pasting or typing a URL into the textbox. Press Enter to load the query.',
+            'If in a web browser paste it into your URL bar in a new tab otherwise start a new query from a URL by clicking the New Query + button then "Import from URL" then pasting or typing a URL into the textbox. Press Enter to load the query.',
```
aswallace, line 12: nit: 'If in a web browser, paste the link into your URL bar in a new tab. Otherwise, start a...'
Also maybe instead of 'then "Import from URL" then pasting or typing a URL into the textbox' could be 'then paste or type a URL into the "Paste URL" textbox'

SeanLeRoy:
> Currently having trouble running the desktop version of the app on this branch but it may just be my local set up, would be great if someone else could also verify.

No such error for me ☹️

packages/core/components/QuerySidebar/tutorials/ShareView.ts
SeanLeRoy: Fixed in #b1bbfe1

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -197,129 +178,22 @@ export default class FileDownloadServiceElectron implements FileDownloadService
...
+            outFilePath,
             requestOptions,
             url,
             writeStreamOptions,
         } = options;
+        // let outFilePath = options.outFilePath;
+        // if (outFilePath === undefined) {
+        //     const defaultDirectory = await this.getDefaultDownloadDirectory();
+        //     outFilePath = path.join(defaultDirectory, path.basename(url));
+        // }
```
aswallace, line 195: still needed?

BrianWhitneyAI:
> > Currently having trouble running the desktop version of the app on this branch but it may just be my local set up, would be great if someone else could also verify.
> 
> No such error for me ☹️

Likewise did not see an error

aswallace (approved):
Tested locally
- [x] downloading to parquet, json, csv
- [x] downloading individual files
- [x] opening files in 3d web viewer and labkey
- [x] query arg persistence

aswallace:
> Currently having trouble running the desktop version of the app on this branch but it may just be my local set up, would be great if someone else could also verify.

Appears to be a persistence issue again for me -- On initial load, my persisted query does not have a `parts` attribute so it's reading as undefined

packages/core/components/FileList/useFileAccessContextMenu.ts
```diff
@@ -8,6 +8,138 @@ import FileFilter from "../../entity/FileFilter";
...
+            disabled: !filePath,
+            target: "_blank",
+        });
+        if (plateLink) {
+            savedApps.push({
+                key: ContextMenuActions.OPEN_PLATE_UI,
+                text: "LabKey Plate UI",
+                title: "Open this plate in the Plate UI",
+                href: plateLink,
+                target: "_blank",
```
aswallace, line 50: LabKey open to a page that says `Hello from production loadbalancer.` Is the intended page?

packages/core/components/FileList/useFileAccessContextMenu.ts
```diff
@@ -8,6 +8,138 @@ import FileFilter from "../../entity/FileFilter";
...
+                {
+                    key: "default-apps-border",
+                    itemType: ContextualMenuItemType.Divider,
+                },
+                {
+                    key: ContextMenuActions.OPEN_WITH_OTHER,
+                    text: "Other...",
+                    title: "Select an application to open the selection with",
+                    onClick() {
+                        dispatch(interaction.actions.promptForNewExecutable(filters));
```
aswallace, line 81: Testing this locally, does not seem to open the selected app for me

BrianWhitneyAI (approved):

packages/core/components/FileList/useFileAccessContextMenu.ts
SeanLeRoy: Huh yeah it seems to be loading the correct page, I didn't change the URL this directs to (intentionally anyway). I'll add an issue to follow this up with Infra to see if we need to use a different URL or if this loadbalancer is broken

packages/core/components/Modal/index.tsx
```diff
@@ -3,17 +3,17 @@ import { useDispatch, useSelector } from "react-redux";
...
 import DataSourcePrompt from "./DataSourcePrompt";
+import MetadataManifest from "./MetadataManifest";
 
 export interface ModalProps {
     onDismiss: () => void;
 }
 
 export enum ModalType {
-    CsvManifest = 1,
+    CodeSnippet = 1,
```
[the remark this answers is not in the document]
SeanLeRoy: Reordered to fit alphabetical order, arbitrary choice that shouldn't have any downstream affects

packages/desktop/src/services/FileDownloadServiceElectron.ts
SeanLeRoy: oof nope removed in https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/106/commits/31d33b29fcb89c5faa32e62aab5f0637bfac79cb

SeanLeRoy:
> > Currently having trouble running the desktop version of the app on this branch but it may just be my local set up, would be great if someone else could also verify.
> 
> Appears to be a PersistedConfig issue again for me -- On initial load, my persisted query did not have a `parts` attribute so it read as undefined. Was able to get around it in a hacky way but may not be a permanent solution

I wonder if this is because you have tried a previous iteration of this branch then. The schema for the persisted config changes for the "queries" key during the development of this branch. Normally when changing the schema I'd write a migration in, but since it was a branch in development it didn't seem it would matter alas

PR #117, opened 2024-06-04
Feature/create datasets page

### LINKED ISSUE #101: Create datasets page
The website will have a page with some internal and external datasets listed. Hopefully by the time you are reading this we will have an S3 bucket you can populate this page with otherwise lets discuss where those will be hosted. Besides that this page should largely just be reflected by what the style document Lyndsay created.

aswallace:
### Description
Creates a page (web only) to display internal & external publicly available datasets that users can load into the app.
Replicates the Dataset Page document in Figma. Points to dataset manifest in BioFile Finder S3 bucket (production)

### Changes
- Created page for dataset route, plus supporting components for the table itself
- Created dataset details panel that appears when a dataset is selected
- Able to load datasets directly from the dataset table or panel into the app
- Added logic to state to use the existing Database Service (to DuckDB). This part is definitely worth looking at again; I tried to reuse existing logic, but there are some things that I think make sense to keep separate (e.g., dataset manifest source should not be considered a data source in the app, so service should return null if there isn't a manifest found)

### Testing
Manually & unit tests

### Screenshots

Dataset page
Dataset panel

closes  #101


packages/core/components/Modal/DataSourcePrompt/index.tsx
```diff
@@ -68,21 +68,7 @@ export default function DataSourcePrompt({ onDismiss }: Props) {
     const onEnterURL = throttle(
         (evt: React.FormEvent) => {
             evt.preventDefault();
-            const uriResource = dataSourceURL
```
aswallace, line 71: Separated the logic in lines 71-85 into a separate method since gets reused in DatasetRow and DatasetDetails

packages/web/src/entity/PublicDataset/index.ts
```diff
@@ -0,0 +1,166 @@
...
+        displayLabel: "Related publication",
+        name: "related_publication",
+    },
+    VERSION: {
+        displayLabel: "Version",
+        name: "version",
+    },
+};
+
+export const DATASET_DISPLAY_FIELDS = [
```
aswallace, line 74: Limited set used for the details panel

packages/web/src/components/Header/Help.module.css
```diff
@@ -16,11 +16,6 @@
 
 .help-menu {
   padding: 4px 0;
-  filter: drop-shadow(black 10px 14px 18px);
```
aswallace, line 19: This styling was not actually being applied/having any effect

packages/core/state/interaction/selectors.ts
```diff
@@ -119,6 +122,24 @@ export const getFileService = createSelector(
...
+ * Selector specifically for open-source dataset manifest, re-using regular file service interface.
+ * Used in web only
+ */
+export const getPublicDatasetManifestService = createSelector(
+    [getDatasetManifestSource, getPlatformDependentServices],
+    (datasetManifestSource, platformDependentServices): FileService | undefined => {
+        if (!datasetManifestSource || !platformDependentServices) {
+            return undefined;
+        }
+        return new DatabaseFileService({
```
SeanLeRoy, line 135: Is this any different than the normal DatabaseFileService returned from the `getFileService` selector?

packages/web/src/components/OpenSourceDatasets/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Page for displaying public-facing datasets
+ * Currently using placeholder text and data
+ */
+export default function OpenSourceDatasets() {
+    const dispatch = useDispatch();
+    // Begin update action so dataset manifest is ready for table child component
+    dispatch(
+        metadata.actions.updateDatasetManifest("Dataset Manifest", DatasetManifestUrl.PRODUCTION)
```
SeanLeRoy, line 19: This action seems like a `getter` to me similar to how we do `getAnnotations` - also looks like this is relying on the component `OpenSourceDatasets` to only render once to prevent this action triggering multiple times, wrapping in a `useEffect` or similar might be a nice guard.

packages/web/src/components/OpenSourceDatasets/useDatasetDetails.ts
```diff
@@ -0,0 +1,52 @@
+import * as React from "react";
+import { useSelector } from "react-redux";
+
+import PublicDataset from "../../entity/PublicDataset";
+import FileSet from "../../../../core/entity/FileSet";
+import { interaction } from "../../../../core/state";
+
+// /**
```
SeanLeRoy, line 8: Double comment

packages/web/src/services/DatabaseServiceWeb.ts
```diff
@@ -72,6 +73,30 @@ export default class DatabaseServiceWeb implements DatabaseService {
         }
     }
 
+    public async addDatasetManifest(name: string, uri: string): Promise<void> {
```
SeanLeRoy, line 76: Looks like this is duplicating a lot of the addDataSource stuff

packages/core/state/interaction/selectors.ts
aswallace: The only difference is the dependence on `dataManifestSource`/`getDatasetManifestSource` instead of on `dataSource`/`getDataSource`. I considered reusing the same dataSource state, but was concerned about the side effects of having them both pointing to the same thing (e.g., would it cause things to re-render in app that shouldn't)

aswallace: Also, `getFileService` defaults to `httpFileService` if no data source is provided, which I don't think we want to do here. This one returns null if there's no dataset manifest

packages/web/src/components/OpenSourceDatasets/index.tsx
aswallace: Added useEffect guard but still haven't significantly changed logic for this one; it's trying to imitate the logic used for getDataSources in the way it interacts with the DatabaseService

packages/web/src/components/OpenSourceDatasets/DatasetTable.module.css
```diff
@@ -0,0 +1,57 @@
...
+  font-family: var(--font-family);
+  font-size: 14px;
+  font-weight: 400;
+}
+
+.table-content > div:last-of-type {
+  border-bottom: none;
+}
+
+/* .table-content div {
```
aswallace, line 38: Lines in table should wrap once before using overflow ellipses
TO DO:  Convert to ticket

packages/core/entity/FileExplorerURL/index.ts
```diff
@@ -48,6 +48,17 @@ export const DEFAULT_AICS_FMS_QUERY: FileExplorerURLComponents = {
     sortColumn: new FileSort(AnnotationName.UPLOADED, SortOrder.DESC),
 };
 
+export const getNameAndTypeFromSourceUrl = (dataSourceURL: string) => {
+    const uriResource = dataSourceURL.substring(dataSourceURL.lastIndexOf("/") + 1).split("?")[0];
+    const name = `${uriResource} (${new Date().toLocaleDateString()} ${new Date().toLocaleTimeString()})`;
+    let extensionGuess = uriResource.split(".").pop();
+    if (!(extensionGuess === "csv" || extensionGuess === "json" || extensionGuess === "parquet")) {
+        console.warn("Guess that the source is a CSV file since no extension easily found");
```
kmitcham, line 56: nit: Maybe "Assuming the source is csv since no extension was recognized"

kmitcham (approved):

SeanLeRoy (approved):

packages/core/entity/FileExplorerURL/index.ts
aswallace: Changed in [f5bcbcf](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/117/commits/f5bcbcf033ba64d00b3bbddf185c30e73f30ce1b)[f5bcbcf](https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/117/commits/f5bcbcf033ba64d00b3bbddf185c30e73f30ce1b)

PR #121, opened 2024-06-12
Add analytics to webapp

### LINKED ISSUE #116: Add analytics
I think thao has an example of how we do this

aswallace:
This adds Google Tag Manager to the webapp to collect basic data about user traffic/interaction.
closes #116 

### Testing
Checked that gtm is showing up with a 200 in the Network tab for all the current routes.
Once these changes are deployed, we'll need to reach out to Thao to verify its presence

### Notes
- Do we need to implement page/event tracking for React Router? 
  Edit: It sounds like most things will be configured through the Google Tag Manager interface itself, and then we can chat with the analytics team if we need something more advanced than what it currently will do
- As is, will get data from everywhere this is deployed, including dev/staging. May need to decide later if we want to keep it that way or narrow down where we collect

### DIFF
--- a/packages/web/webpack/index.html
+++ b/packages/web/webpack/index.html
@@ -4,9 +4,23 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="icon" type="image/png" href="logo.png" />
+    <!-- Google Tag Manager -->
+    <script>(function (w, d, s, l, i) {
+            w[l] = w[l] || []; w[l].push({
+                'gtm.start':
+                    new Date().getTime(), event: 'gtm.js'
+            }); var f = d.getElementsByTagName(s)[0],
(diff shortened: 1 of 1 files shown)


kmitcham (approved):

packages/web/webpack/index.html
```diff
@@ -4,9 +4,23 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="icon" type="image/png" href="logo.png" />
```
BrianWhitneyAI, line 6: Nit but I don't think you need the closing / for link

BrianWhitneyAI (approved):

packages/web/webpack/index.html
aswallace: I'm seeing some mixed info about whether that's still the case for html5 (vs html4)

PR #122, opened 2024-06-18
Feature/complete dataset page styling

### LINKED ISSUE #119: Add incomplete functionality/styling to datasets page
The following tasks were skipped in the MVP of the dataset page since were considered lower priority and/or needed to be looked into further: 
- Clicking on the header of the dataset manifests table should allow users to sort that column in desc/asc order. Header should have a hover state to indicate clickability
  - This requires diving into the FluentUI Shimmered DetailsList component
  - Was not prioritized since we don't have datasets yet
- Responsive rows:  Should have a faded overlay on the right on overflow, and hover buttons should stick to the right side of the table
- Currently, after a user loads a dataset into the app, it takes 3 clicks of the browser's back button to get back to the dataset page (i.e., there are multiple route changes instead of just one)

aswallace:
### Description
This is a follow-up to #101 , finishing some secondary UX tasks for the datasets page. Closes #119 
- Added sorting to dataset manifest table columns & styling to indicate clickable headers
- Added links + styling to reference publication and DOI
- Refactored dataset annotations into their own class, both to standardize them & also to get rid of DatasetColumns.ts (which had gross constants & duplicate code)
- Included a partial fix to the router issue where back button has to be clicked multiple times to go back from app to datasets page, but only works after the first time re-routing. This seems to be somewhat built in to the way react-router pushes history 

### Not addressed
I have not found a feasible way to make a responsive horizontal overlay on the rows where the hover-state buttons will stick to the right side of the panel when the screen is small (instead of having to scroll right to find them) because of the way the table sets up inheritance. However, this is less urgent, since the same functionality is available when users click on the row to open the side panel

### Testing
- Tested on a mock manifest with more datasets since we don't have enough datasets to sort on yet
- Added a couple unit tests

### Screenshots
The gray overlay on `CREATION DATE` is a hover state
image


packages/web/src/components/DatasetDetails/DatasetDetailsRow.module.css
```diff
@@ -31,3 +31,14 @@
   /* Allow whitespace to be rendered (including carriage returns & new lines) */
   white-space: pre-wrap; 
 }
+
+.link {
+  text-decoration: none;
```
SeanLeRoy, line 36: Might be unnecessary now (all `a` elements have this style now on `main`)

packages/web/src/components/DatasetDetails/test/DatasetDetails.test.tsx
```diff
@@ -72,6 +72,28 @@ describe("<DatasetDetails />", () => {
             expect(getByText(mockDataset.name)).to.exist;
             expect(getByText(mockDataset.description)).to.exist;
         });
+        it("renders links for ref publication and DOI if provided", () => {
```
SeanLeRoy, line 75: 👍 thank you for adding tests to new features 🚀

packages/web/src/entity/PublicDataset/index.ts
```diff
@@ -20,60 +20,40 @@ export interface PublicDatasetProps {
...
-    },
-    SOURCE: {
-        displayLabel: "Source",
-        name: "source",
-    },
-    VERSION: {
-        displayLabel: "Version",
-        name: "version",
-    },
+    CREATION_DATE: new DatasetAnnotation("Creation date", "created", 112),
```
SeanLeRoy, line 45: Good idea 👍

packages/web/src/components/OpenSourceDatasets/DatasetTable.tsx
```diff
@@ -54,16 +75,37 @@ export default function DatasetTable(props: DatasetTableProps) {
...
         const fieldContent = item[column?.fieldName as keyof PublicDatasetProps] as string;
         if (!fieldContent) return <>--</>;
+        if (
+            column?.fieldName === DatasetAnnotations.RELATED_PUBLICATON.name &&
+            item?.[DatasetAnnotations.DOI.name as keyof PublicDatasetProps]
+        ) {
+            return (
+                <a
+                    className={classNames(styles.link, styles.doubleLine)}
+                    href={item[DatasetAnnotations.DOI.name as keyof PublicDatasetProps]}
```
SeanLeRoy, line 85: Since `item` is typed as `PublicaDatasetProps` do we gain from having the name of `doi` be used as a constant here? It seems like it is making `Typescript` work against you.

SeanLeRoy (approved):

packages/web/src/components/OpenSourceDatasets/DatasetTable.tsx
aswallace: good point, it don't think the typescript pain is worth it here. Will change this

PR #142, opened 2024-06-25
Prevent duplicate query creation on desktop

### LINKED ISSUE #133: Desktop, on reload a new query is added each time 🐛

SeanLeRoy:
**Description**
On desktop the query args were being used to create a duplicate query on refresh. In general we don't really want to use the query args on the desktop so I just added a check to make sure we were on web being this conditional path was allowed.

**Testing**
Tested this on desktop and on web. Unsure of a way to mock the `window.location.search` in unit tests so for now I could not add a test for this 😢 if you have any ideas of how to do so lmk

**Related Issue**
Resolves #133

### DIFF
--- a/packages/core/state/interaction/logics.ts
+++ b/packages/core/state/interaction/logics.ts
@@ -50,6 +50,7 @@ const checkAicsEmployee = createLogic({
     type: INITIALIZE_APP,
     async process(deps: ReduxLogicDeps, dispatch, done) {
         const queries = selection.selectors.getQueries(deps.getState());
+        const isOnWeb = interactionSelectors.isOnWeb(deps.getState());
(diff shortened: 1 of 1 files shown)


aswallace:
> Unsure of a way to mock the window.location.search in unit tests so for now I could not add a test for this 😢 if you have any ideas of how to do so lmk

I'm pretty sure you can do 
```            
Object.defineProperty(window, "location", {
    value: {
        search: "?group=Cell+Line"
    },
}); 
```

in combination with 
```
state:  mergeState(initialState, {
    interaction: {
       isOnWeb: true,
    },
}),
```

aswallace (approved):

BrianWhitneyAI (approved):

kmitcham (approved):

PR #149, opened 2024-06-28
Enable list picking for string values even when over 100 values available

### LINKED ISSUE #86: Add toggle between list picker & search form
For annotations with > 100 values of certain data types we enforce a search form entry style. This may not be totally compatible with how all users interact with data, lets let them toggle between the two types.

SeanLeRoy:
**Description**
This changeset adds a list picker for values like "Clone" where we have over 100 String type values so we would normally just render a search box, but now we have a radio selection for switching between search box and list picking for this use case. This is essentially for users that may not know what value they are looking for just yet (or which are available).

**Testing**
Working on desktop and web.

**Related Issue**
Resolves #86


packages/core/components/AnnotationFilterForm/SearchBoxForm/index.tsx
```diff
@@ -1,32 +1,73 @@
...
+                    key: 'search-box',
+                    text: 'Search box',
+                },{
+                    key: 'list-picker',
+                    text: 'List picker',
+                    disabled: props.items.length === 0,
+                }]}
+                onChange={(_, selection) => {
+                    props.onDeselectAll();
+                    setIsListPicking(selection?.key === "list-picker");
```
aswallace, line 49: minor: This has small side effect that you can't edit the list. e.g., if you've selected via list picker and close the modal, and then re-open from edit, it opens back to search box, and then if you try to go back to list picker it clears

aswallace (approved):
Only minor issue I spotted was the non-editable thing

griffinfujioka (approved):

PR #151, opened 2024-06-28
Fix resize handle

### LINKED ISSUE #141: Fix right sidebar dragging
The right sidebar where the file details are displayed is intended to be draggable for expanding or minimizing the sidebar width, however at the moment it doesn't seem to actually move anything other than the contents.

SeanLeRoy:
The resize handle was broken because the width of the rest of the app was set to a fixed `80%` which meant that it wasn't considering the adjustment from the variable `--file-details-width` which is what determines the right sidebar pane's width. After changing the width calcuation this worked, added some styling changes to make it not have strange horizontal scroll behavior.

Resolves #141

### DIFF
--- a/packages/core/components/FileDetails/index.tsx
+++ b/packages/core/components/FileDetails/index.tsx
@@ -99,57 +99,55 @@ export default function FileDetails(props: Props) {
             className={classNames(styles.root, styles.expandableTransition, props.className)}
             id={FILE_DETAILS_PANE_ID}
         >
-            <div className={styles.fileDetailsContent}>
-                <div
-                    className={styles.resizeHandle}
-                    onMouseDown={(e) => resizeHandleOnMouseDown(e)}
-                    // TODO:???
-                    onDoubleClick={resizeHandleDoubleClick}
-                >
-                    <div />
-                </div>
-                <div className={styles.paginationAndContent}>
-                    <Pagination className={styles.pagination} />
-                    <div className={styles.overflowContainer}>
-                        {fileDetails && (
-                            <>
-                                <div className={styles.thumbnailContainer}>
-                                    <FileThumbnail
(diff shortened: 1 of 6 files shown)


packages/core/components/FileDetails/index.tsx
```diff
@@ -99,57 +99,55 @@ export default function FileDetails(props: Props) {
             className={classNames(styles.root, styles.expandableTransition, props.className)}
             id={FILE_DETAILS_PANE_ID}
         >
-            <div className={styles.fileDetailsContent}>
```
SeanLeRoy, line 102: Only change here is I deleted the `<div className={styles.fileDetailsContent}>` element which changed the indentation

aswallace (approved):

packages/core/components/FileDetails/index.tsx
```diff
@@ -99,57 +99,55 @@ export default function FileDetails(props: Props) {
...
-                                <h4>Information</h4>
-                                <FileAnnotationList
-                                    className={styles.annotationList}
-                                    fileDetails={fileDetails}
-                                    isLoading={isLoading}
+            <div
+                className={styles.resizeHandle}
+                onMouseDown={(e) => resizeHandleOnMouseDown(e)}
+                // TODO:???
+                onDoubleClick={resizeHandleDoubleClick}
```
aswallace, line 106: unrelated to PR, I'm curious what this to do is/what the behavior will be

BrianWhitneyAI (approved):

packages/core/components/FileDetails/index.tsx
SeanLeRoy: Ah I can remove that `TODO` this actually just resets the width of the pane back to `20%` from whatever it was

PR #186, opened 2024-07-30
Use regex to enable queries to search against comma separated values

### LINKED ISSUE #185: Unable to use comma separated values in data sources
As in columns with multiple values for a given row

SeanLeRoy:
# Description
Antoine pointed out that there isn't a way to include multiple values per column per row (see screenshot for example). This was intended to be included as a feature from the start using a comma (",") as a delimeter. However, it seems it was forgotten in a couple places which prevents it from actually working. This changeset uses regex to search using this in mind.

Screen Shot 2024-07-29 at 5 07 18 PM

Screen Shot 2024-07-29 at 5 08 13 PM

Additionally, I simplified the `fetchValues` method to re-use the query code found in fetching values under a hierarchy.

# Related Issue
Resolves #185

### DIFF
--- a/packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
+++ b/packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
@@ -94,7 +94,10 @@ export default class DatabaseAnnotationService implements AnnotationService {
             } else {
                 sqlBuilder.where(
                     annotationValues
-                        .map((value) => `"${annotationToFilter}" = '${value}'`)
+                        .map(
+                            (value) =>
+                                `REGEXP_MATCHES("${annotationToFilter}", '(\\b${value}\\b)') = true`
+                        )
                         .join(") OR (")
                 );
             }
--- a/packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
+++ b/packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
@@ -101,10 +101,10 @@ export default class DatabaseAnnotationService implements AnnotationService {
         });
 
         const rows = await this.databaseService.query(sqlBuilder.toSQL());
-        const rowsSplitByDelimeter = rows
-            .flatMap((row) => row[annotation].split(DatabaseService.LIST_DELIMETER))
+        const rowsSplitByDelimiter = rows
+            .flatMap((row) => row[annotation].split(DatabaseService.LIST_DELIMITER))
             .map((value) => value.trim());
-        return uniq(rowsSplitByDelimeter);
+        return uniq(rowsSplitByDelimiter);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -81,23 +76,35 @@ export default class DatabaseAnnotationService implements AnnotationService {
...
                 );
             }
         });
 
         const rows = await this.databaseService.query(sqlBuilder.toSQL());
-        return rows.map((row) => row[hierarchy[path.length]]);
+        const rowsSplitByDelimeter = rows
+            .flatMap((row) => row[annotation].split(DatabaseService.LIST_DELIMETER))
+            .map((value) => value.trim());
+        return uniq(rowsSplitByDelimeter);
```
aswallace, line 107: nit: delimiter. Delimeter is a unit of measurement for subway sandwiches

aswallace (approved):

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -81,23 +76,35 @@ export default class DatabaseAnnotationService implements AnnotationService {
...
+        Object.keys(filtersByAnnotation).forEach((annotationToFilter) => {
+            const annotationValues = filtersByAnnotation[annotationToFilter];
             if (annotationValues[0] === null) {
-                sqlBuilder.where(`"${annotation}" IS NOT NULL`);
+                sqlBuilder.where(`"${annotationToFilter}" IS NOT NULL`);
             } else {
                 sqlBuilder.where(
-                    annotationValues.map((value) => `"${annotation}" = '${value}'`).join(") OR (")
+                    annotationValues
+                        .map((value) => `"${annotationToFilter}" = '${value}'`)
```
aswallace, line 97: Not sure if I'm commenting in the right place, but noticed that when doing a grouped query (e.g., group by color), if a file has multiple values (red, green, blue) and you filter on just color=green it won't return that file. When no grouping is applied color=green returns the red,green,blue file

packages/core/services/DatabaseService/index.ts
```diff
@@ -10,6 +10,7 @@ import SQLBuilder from "../../entity/SQLBuilder";
  * Service reponsible for querying against a database
  */
 export default abstract class DatabaseService {
+    public static readonly LIST_DELIMETER = ",";
```
kmitcham, line 13: nit: anya caught this: LIST_DELIMITER

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
SeanLeRoy: Resolved in https://github.com/AllenInstitute/aics-fms-file-explorer-app/pull/186/commits/4df670694af1d9232c63892b2ea2d2dc0b8172a1
I wasn't using regex when grabbing annotation values to use as folders in the heirarchy

PR #188, opened 2024-07-30
Create footer with cookie settings button

### LINKED ISSUE #180: "Cookie Settings" anchor link: add this element to BioFile Finder homepage for user to revisit their consent choices
**UPDATED**
- [ ] Add the following HTML button element to the bottom of the homepage to allow access to the cookie consent form 
```
<!-- OneTrust Cookies Settings button start -->
<button id="ot-sdk-btn" class="ot-sdk-show-settings">Cookie Settings</button>
<!-- OneTrust Cookies Settings button end -->
```
- [ ] Styling: centering the anchor element - footer is not required, *note: font-family within the cookie banner and consent form is inherited from the main `<html>` - please make sure it is set so the banner and consent form look consistent with everything else.

aswallace:
## Context
Users need to be able to re-access their cookie settings after the first time they've dismissed the pop-up. 

## Changes
- Added a footer component to web only
- Added `OneTrust` button component with custom styling overrides
- Modified some `main` styling to account for footer in scroll boundary

## Screenshots
image

closes #180

### DIFF
--- a/packages/web/src/components/Footer/Footer.module.css
+++ b/packages/web/src/components/Footer/Footer.module.css
@@ -1,0 +1,19 @@
+.footer {
+  height: fit-content;
+  margin: 0 auto;
+  padding: var(--margin) 0;
+  border-top: 1px solid var(--border-color);
+  text-align: center;
+}
+
+.footer-link {
+  color: var(--primary-text-color) !important;
+  border: none !important;
+  font-size: var(--s-paragraph-size) !important;
+}
+
+.footer-link:hover {
+  color: var(--highlight-text-color) !important;
+  border: 0px solid var(--primary-text-color) !important;
+  background-color: transparent !important;
+}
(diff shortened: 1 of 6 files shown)


packages/web/src/components/Footer/Footer.module.css
```diff
@@ -0,0 +1,19 @@
...
+
+.footer-link {
+  color: var(--primary-text-color) !important;
+  border: none !important;
+  font-size: var(--s-paragraph-size) !important;
+}
+
+.footer-link:hover {
+  color: var(--highlight-text-color) !important;
+  border: 0px solid var(--primary-text-color) !important;
```
BrianWhitneyAI, line 17: Border is 0px? why does it need '0px solid var(--primary-text-color)'

BrianWhitneyAI (approved):
LGTM!

packages/web/src/components/Footer/Footer.module.css
aswallace: Good catch, was midway through changing this to `none` but didn't push

kmitcham (approved):

PR #191, opened 2024-08-01
Fix reg ex to account for spaces

SeanLeRoy:
Previously the regular expression that is used to look for a value among a comma separated list of values was not matching on values that had whitespace in varying amounts. I updated the regular expression to look explicitly for one of:
* A match at the beginning of a comma separated list
* A match in the middle of a comma separated list
* A match at the end of a comma separated list
* A match that is the entire value (AKA no list, just one value)

### DIFF
--- a/packages/core/entity/FileSet/index.ts
+++ b/packages/core/entity/FileSet/index.ts
@@ -197,7 +197,13 @@ export default class FileSet {
             } else {
                 sqlBuilder.where(
                     filterValues
-                        .map((fv) => `REGEXP_MATCHES("${annotation}", '(\\b${fv}\\b)') = true`)
+                        .map(
+                            (fv) =>
+                                // Ex. This regex will match on a value
+                                // that is at the start, middle, end, or only value in a comma separated list
+                                // of values (,\s*Position,)|(^\s*Position\s*,)|(,\s*Position\s*$)|(^\s*Position\s*$)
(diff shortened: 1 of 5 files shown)


BrianWhitneyAI (approved):

aswallace (approved):

packages/core/entity/FileSet/index.ts
```diff
@@ -197,7 +197,13 @@ export default class FileSet {
             } else {
                 sqlBuilder.where(
                     filterValues
-                        .map((fv) => `REGEXP_MATCHES("${annotation}", '(\\b${fv}\\b)') = true`)
+                        .map(
+                            (fv) =>
+                                // Ex. This regex will match on a value
+                                // that is at the start, middle, end, or only value in a comma separated list
+                                // of values (,\s*Position,)|(^\s*Position\s*,)|(,\s*Position\s*$)|(^\s*Position\s*$)
+                                `REGEXP_MATCHES("${annotation}", '(,\\s*${fv}\\s*,)|(^\\s*${fv}\\s*,)|(,\\s*${fv}\\s*$)|(^\\s*${fv}\\s*$)') = true`
```
kmitcham, line 205: could the regex be a constant, and tested somewhere against examples?

kmitcham (approved):

PR #204, opened 2024-08-08
feature/zarr-download

### LINKED ISSUE #192: Enable downloading OME ZARRs
Because OME ZARRs are a directory we can't download them like normal. We likely need to zip them and then try to download the zip.

BrianWhitneyAI:
## Description 
The purpose of this PR is to resolve #192 and add downloadability for zarr files both locally and from remote URLs.


BrianWhitneyAI:
At an initial pass here's what's going on:

This only works with S3 buckets right now. I tested it on the variance dataset. working on local...

###DESKTOP:
The download for desktop will give you the actual file but will propagate the file after creation. This does not currently work with our download notifications and shows 0/0 bytes completed. I have tested it with files up to 300 mbs but have no idea what this'll do ram wise.

####
This downloads a zip. however there is no handling around the process. it'll just show up in your downloads after a few min.

BrianWhitneyAI:
* also has not passed test suite

pgarrison:
I suspect this PR should be merged despite the following potential issue. 

I'm not sure this fully resolves #192. I would expect that from the web a user should be able to download a ZARR that is too big for their RAM but small enough for their hard drive. For example, this 42GB ZARR:`https://allencell.s3.amazonaws.com/aics/emt_timelapse_dataset/data/3500005548_25_raw_converted.ome.zarr/`.

My quick read of the code suggests that this PR attempts to load the entire ZARR into memory [here](https://github.com/AllenInstitute/biofile-finder/pull/204/files#diff-42033498f3c7b2c91f6ee60b3e73ec03e78f563adda620e53cfc0401c315aca6R64) before finally downloading the ZARR. If I am correct and the ZARR linked above fails to download on a regular laptop, one possible option is to download just a few files at a time with the [File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API). This may get around UI issues with downloading tons of files.

toloudis:
also is it cancellable?

BrianWhitneyAI:
@pgarrison @SeanLeRoy I have added StreamSaver. Do either of you have a large S3 file I can double check it is working correctly with? everything looks good for the variance dataset zarrs but they're smaller.

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -15,6 +15,8 @@ import {
     HttpServiceBase,
 } from "../../../core/services";
 import { DownloadFailure } from "../../../core/errors";
+import axios from "axios";
```
SeanLeRoy, line 18: These should go in the third party imports section alphabetically (by `cockatiel` and `electron`

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -66,7 +68,35 @@ export default class FileDownloadServiceElectron
...
+                return this.downloadS3Directory(
+                    fileInfo,
+                    downloadRequestId,
+                    onProgress,
+                    destination
+                );
+            }
+
+            const isLocal = await this.isLocalPath(fileInfo.path);
+            if (isLocal) {
```
SeanLeRoy, line 85: Rather than checking for S3 specifically could we hope `zarrita` supports other cloud infra like Azure or whatever too or is that definitely not possible? If so, seems like this could be simpler and more versatile
Example:
```
const isLocalFile = await this.isLocalPath(fileInfo.path);
if (isLocalFile) {
    return this.copyLocalZarrDirectory(fileInfo, downloadRequestId, onProgress, destination);
}
return this.downloadS3Directory(fileInfo, downloadRequestId, onProgress, destination);
```

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -97,12 +127,79 @@ export default class FileDownloadServiceElectron
...
+            const fullDestinationDir = path.join(destinationDir, fileInfo.name);
+
+            await this.copyDirectory(fileInfo.path, fullDestinationDir, onProgress);
+
+            return {
+                downloadRequestId: fileInfo.id,
+                msg: `Successfully copied Zarr directory ${fileInfo.path} to ${fullDestinationDir}`,
+                resolution: DownloadResolution.SUCCESS,
+            };
+        } catch (err: unknown) {
```
SeanLeRoy, line 165:
```suggestion
        } catch (err) {
```

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -256,7 +358,7 @@ export default class FileDownloadServiceElectron
                             })
                         );
                         await this.deleteArtifact(outFilePath);
-                    } catch (err) {
+                    } catch (err: unknown) {
```
SeanLeRoy, line 361: Do we need this?
```suggestion
                    } catch (err) {
```

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -279,10 +381,10 @@ export default class FileDownloadServiceElectron
                 });
             });
 
-            req.on("error", async (err) => {
+            req.on("error", async (err: unknown) => {
```
SeanLeRoy, line 384: implicit or `Error`?
```suggestion
            req.on("error", async (err) => {
```

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -317,12 +419,92 @@ export default class FileDownloadServiceElectron
...
-        const response = await this.rawPost<string>(url, postBody);
-        return new Blob([response], { type: "application/json" });
+    private async downloadS3Directory(
+        fileInfo: FileInfo,
+        downloadRequestId: string,
+        onProgress?: (transferredBytes: number) => void,
+        destination?: string
+    ): Promise<DownloadResult> {
+        const { bucket, key, region } = this.parseS3Url(fileInfo.path);
+        const destinationDir = destination || (await this.getDefaultDownloadDirectory());
```
SeanLeRoy, line 429: This happens in both the zarr download functions so I'd bring it out into the function that calls them

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -317,12 +419,92 @@ export default class FileDownloadServiceElectron
...
+            console.error(`Failed to download directory: ${err}`);
+            throw new DownloadFailure(
+                `Failed to download directory: ${(err as Error).message}`,
+                downloadRequestId
+            );
+        }
     }
 
-    public cancelActiveRequest(downloadRequestId: string) {
+    private async downloadS3File(
```
SeanLeRoy, line 475: This seems generic, could it be `downloadFile`?

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -317,12 +419,92 @@ export default class FileDownloadServiceElectron
...
+                    continue;
+                }
+
+                const relativePath = path.relative(key, fileKey);
+                const destinationPath = path.join(fullDestinationDir, relativePath);
+
+                // Ensure the subdirectories exist
+                fs.mkdirSync(path.dirname(destinationPath), { recursive: true });
+
+                const fileUrl = `https://${bucket}.s3.${region}.amazonaws.com/${encodeURIComponent(
```
SeanLeRoy, line 454: Don't we have this initial host name from the file's "File Path"?

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -352,4 +534,32 @@ export default class FileDownloadServiceElectron
...
+
+    private parseS3Url(url: string): { bucket: string; key: string; region: string } {
+        const { hostname, pathname } = new URL(url);
+        const [bucket] = hostname.split(".");
+        const key = pathname.slice(1);
+        const region = hostname.split(".")[2];
+        return { bucket, key, region };
+    }
+
+    private async listS3Objects(bucket: string, prefix: string, region: string): Promise<string[]> {
```
SeanLeRoy, line 546: I take it zarrita can't do this for us?

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -352,4 +534,32 @@ export default class FileDownloadServiceElectron
...
+        const url = `https://${bucket}.s3.${region}.amazonaws.com?list-type=2&prefix=${encodeURIComponent(
+            prefix
+        )}`;
+        const response = await axios.get(url);
+        const parsedResult = await parseStringPromise(response.data);
+
+        const contents = parsedResult.ListBucketResult.Contents || [];
+        const keys: string[] = [];
+
+        for (const content of contents) {
```
SeanLeRoy, line 556: FWIW Could be more ~ javascripty ~ using more innate array toolings
```
return contents
    .map(content => content.Key?.[0])
    .filter(key => key === "string");
```

packages/web/src/services/FileDownloadServiceWeb.ts
```diff
@@ -39,6 +45,73 @@ export default class FileDownloadServiceWeb extends HttpServiceBase implements F
...
+
+    private parseS3Url(url: string): { bucket: string; key: string; region: string } {
+        const { hostname, pathname } = new URL(url);
+        const [bucket] = hostname.split(".");
+        const key = pathname.slice(1);
+        const region = hostname.split(".")[2];
+        return { bucket, key, region };
+    }
+
+    private async listS3Objects(bucket: string, prefix: string, region: string): Promise<string[]> {
```
SeanLeRoy, line 97: `parseS3Url` & `listS3Objects` seem like they should be put into the super class for this. You can make the super an abstract class so you can add the methods to it.

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: Zarrita does not have any writing functions to save/write out the zarr files so we have to deal directly with where the file is stored. I think this means that we have to add support for each type of storage. The only test data I have right now is S3 but if we move towards other storage locations we can open a ticket for them and add individual support?  I don't really see a way to universalize directory download, unless there's some dependency we could use that I don't know about...

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: oh whoops, a symptom of trying lots of bits and pieces. will fix

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: There is some merit to the flatmap function of zarrita for listing contents but in dealing with it you have to make a lot of assumptions (I think) so it didn't end up working too well for me.

packages/web/src/services/FileDownloadServiceWeb.ts
BrianWhitneyAI: ooh cool didn't think of that. will do.

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: we make the distinction above of downloading a Http file. This also downloads a Http file just without a lot of the wrapping since we have to do a large amount of then for zarr. would something like downloadSimpleHttpFile be acceptable?

packages/web/src/services/FileDownloadServiceWeb.ts
BrianWhitneyAI: resolved in [cfdf9b4](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/cfdf9b40628c86077ac4ad73f158267bde3b5b49)

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: resolved in [cfdf9b4](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/cfdf9b40628c86077ac4ad73f158267bde3b5b49) by combining logic. needs more testing though once the caching issue is fixed.

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: resolved in [cc3d7a5](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/cc3d7a5d5be38541d1c9db2c770d19809f60a919)

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: resolved in [cfdf9b4](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/cfdf9b40628c86077ac4ad73f158267bde3b5b49)

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: [cfdf9b4](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/cfdf9b40628c86077ac4ad73f158267bde3b5b49)

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: [cfdf9b4](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/cfdf9b40628c86077ac4ad73f158267bde3b5b49)

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: resolved in [cfdf9b4](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/cfdf9b40628c86077ac4ad73f158267bde3b5b49)

BrianWhitneyAI:
> e5ef578

This commit was the addition of Streamsaver. There has been some iteration since but I believe its the same functionality

BrianWhitneyAI:
> I suspect this PR should be merged despite the following potential issue.
> 
> I'm not sure this fully resolves #192. I would expect that from the web a user should be able to download a ZARR that is too big for their RAM but small enough for their hard drive. For example, this 42GB ZARR:`https://allencell.s3.amazonaws.com/aics/emt_timelapse_dataset/data/3500005548_25_raw_converted.ome.zarr/`.
> 
> My quick read of the code suggests that this PR attempts to load the entire ZARR into memory [here](https://github.com/AllenInstitute/biofile-finder/pull/204/files#diff-42033498f3c7b2c91f6ee60b3e73ec03e78f563adda620e53cfc0401c315aca6R64) before finally downloading the ZARR. If I am correct and the ZARR linked above fails to download on a regular laptop, one possible option is to download just a few files at a time with the [File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API). This may get around UI issues with downloading tons of files.

> e5ef578

This commit was the addition of Streamsaver. There has been some iteration since but I believe its the same functionality

packages/web/src/services/FileDownloadServiceWeb.ts
```diff
@@ -1,76 +1,119 @@
...
+
+                const blob = response.data;
+                const stream = blob.stream();
+                await stream.pipeTo(
+                    new WritableStream({
+                        write(chunk) {
+                            writer.write(chunk);
+                        },
+                    })
+                );
```
pgarrison, line 84: Can this be simplified to something like the following?
```suggestion
                await stream.pipeTo(writer);
```

packages/core/services/FileStorageServiceBase/index.ts
```diff
@@ -0,0 +1,36 @@
+import axios from "axios";
+
+import HttpServiceBase from "../HttpServiceBase";
+
+export default class FileStorageServiceBase extends HttpServiceBase {
```
SeanLeRoy, line 5: Changing this to an abstract class would let you move the methods defined in `FileDownloadService` (the interface) thereby letting you not need both a `FileDownloadService` interface + `FileDownloadServiceBase` class but instead just one `FileDownloadService` abstract class.
```suggestion
export default abstract class FileStorageService extends HttpServiceBase {
```

packages/desktop/src/services/FileDownloadServiceElectron.ts
SeanLeRoy: 🆗

packages/web/src/services/FileDownloadServiceWeb.ts
SeanLeRoy: Left another comment checking in about the way this was abstracted

packages/web/src/services/FileDownloadServiceWeb.ts
```diff
@@ -1,23 +1,113 @@
...
+        return this.downloadFile(fileInfo);
+    }
+
+    private async handleZarrFile(fileInfo: FileInfo): Promise<DownloadResult | null> {
+        if (this.isS3Url(fileInfo.path)) {
+            return await this.downloadS3Directory(fileInfo);
+        }
+
+        const localDownloadResult = this.isLocalPath(fileInfo.path, fileInfo);
+        return localDownloadResult; // This will be either a DownloadResult or null
```
SeanLeRoy, line 34: Feels like we should throw an error here since we couldn't download it when null then?

packages/web/src/services/FileDownloadServiceWeb.ts
```diff
@@ -1,23 +1,113 @@
...
+    }
+
+    private isLocalPath(filePath: string, fileInfo: FileInfo): DownloadResult | null {
+        const uriPattern = /^(https?|ftp):\/\/|^[a-zA-Z]:\\/;
+        const isLocal = filePath.startsWith("file://") || !uriPattern.test(filePath);
```
SeanLeRoy, line 39: I don't think the path necessarily needs to start with `file://` ex. `/allen/aics/a/path/to/a/file` is how most of our paths look; unless "file://" is automatically added somewhere else?

SeanLeRoy: I wonder if instead of checking if it is a local file here we could:

For web, throw an error / alert when the file isn't on S3 since we don't support any other locations with a message like "The .zarr file located at <blah> is not accessible to this application. BioFile Finder currently only supports S3 locations & they must be accessible to you the user from this computer."

For desktop, if S3 do the download per usual. If not, try to access it using `fs` and if it works then just do the download. If that also doesn't work, then throw an error saying it isn't accessible to this application and provide the error we get from `fs`.

packages/web/src/services/FileDownloadServiceWeb.ts
```diff
@@ -1,23 +1,113 @@
...
+
+        if (isLocal) {
+            const directoryPath = fileInfo.path;
+
+            const message = `The directory containing the Zarr file is located at: ${directoryPath}.
+Due to security restrictions, the web browser cannot open this location directly. 
+Please navigate to this directory manually, or upload files to a remote address such as S3.`;
+
+            alert(message);
+            console.log(`Local directory path: ${directoryPath}`);
```
SeanLeRoy, line 49: (nit) remove console log

packages/web/src/services/FileDownloadServiceWeb.ts
```diff
@@ -1,23 +1,113 @@
...
+            const downloadResult = await this.handleZarrFile(fileInfo);
+            if (downloadResult) {
+                return downloadResult;
+            }
+        }
+
+        return this.downloadFile(fileInfo);
+    }
+
+    private async handleZarrFile(fileInfo: FileInfo): Promise<DownloadResult | null> {
```
SeanLeRoy, line 28: Maybe something like this? If we aren't using zarrita could this just download arbitrary folders?
```suggestion
    private async handleDirectoryLikeFile(fileInfo: FileInfo): Promise<DownloadResult | null> {
```

packages/web/src/services/FileDownloadServiceWeb.ts
```diff
@@ -1,23 +1,113 @@
...
+        if (this.isS3Url(fileInfo.path)) {
+            return await this.downloadS3Directory(fileInfo);
+        }
+
+        const localDownloadResult = this.isLocalPath(fileInfo.path, fileInfo);
+        return localDownloadResult; // This will be either a DownloadResult or null
+    }
+
+    private isLocalPath(filePath: string, fileInfo: FileInfo): DownloadResult | null {
+        const uriPattern = /^(https?|ftp):\/\/|^[a-zA-Z]:\\/;
```
SeanLeRoy, line 38: Is there any way we could support `s3` protocol files as well? Here is an example: `s3://allencell/aics/emt_timelapse_dataset/data/3500006062_56_all_cells_mask.ome.zarr`

packages/desktop/src/services/FileDownloadServiceElectron.ts
SeanLeRoy: If they both download a file over http what is the difference? Could we re-use the other function?

pgarrison:
I'm starting to think #192 is too large for a single PR. We're simultaneously wrestling with:
* Both desktop and browser
* Local files, remote files on S3 with HTTP URLs, remote files on S3 with `s3://` URLs, remote files on Azure/other hosting with HTTP URLs (I think @toloudis brought up Azure but I don't see that comment any more).
* Efficient file saving pattern for large ZARRs (to zip or to save multiple files? how to stream to a zip?)
* Discovering which files are in a ZARR (currently the S3-specific `listS3Objects`)
Would it be easier to divide this (substantial) feature into smaller parts?

SeanLeRoy:
> I'm starting to think #192 is too large for a single PR. We're simultaneously wrestling with:
> 
> * Both desktop and browser
> * Local files, remote files on S3 with HTTP URLs, remote files on S3 with `s3://` URLs, remote files on Azure/other hosting with HTTP URLs (I think @toloudis brought up Azure but I don't see that comment any more).
> * Efficient file saving pattern for large ZARRs (to zip or to save multiple files? how to stream to a zip?)
> * Discovering which files are in a ZARR (currently the S3-specific `listS3Objects`)
> 
> Would it be easier to divide this (substantial) feature into smaller parts?

This seems like a good idea. The most important use case is an `https` file path leading to a public `s3` .zarr file from the web version of this application. Though it does have to be efficient. If you'd like to separate some stuff out I'd recommend saving the `s3` protocol and `desktop` versions for separate units of work.

packages/web/src/services/FileDownloadServiceWeb.ts
BrianWhitneyAI: How does this work? its hosted on s3 but mounted locally?

package.json
```diff
@@ -45,7 +45,9 @@
     "@types/react-window": "1.8.x",
     "@types/react-window-infinite-loader": "1.0.x",
     "@types/sinon": "10.x",
+    "@types/streamsaver": "^2.0.4",
     "@types/string-natural-compare": "3.0.x",
+    "@types/xml2js": "^0.4.14",
```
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in [027bc6d](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/027bc6dd1c70892b7b14a5f81606886c1f5dd76f) where it is removed since we dont use xml2js anymore.

package.json
```diff
@@ -71,7 +73,7 @@
     "@aics/redux-utils": "0.6.0",
     "@fluentui/react": "8.67.x",
     "@tippyjs/react": "4.2.x",
-    "axios": "0.21.x",
+    "axios": "^0.21.4",
```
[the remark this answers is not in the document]
BrianWhitneyAI: returned to original in [027bc6d](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/027bc6dd1c70892b7b14a5f81606886c1f5dd76f)

package.json
```diff
@@ -91,6 +93,7 @@
     "redux": "4.0.x",
     "redux-logic": "3.x",
     "reselect": "4.0.x",
+    "streamsaver": "^2.0.6",
```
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in [027bc6d](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/027bc6dd1c70892b7b14a5f81606886c1f5dd76f)

package.json
```diff
@@ -45,7 +45,9 @@
     "@types/react-window": "1.8.x",
     "@types/react-window-infinite-loader": "1.0.x",
     "@types/sinon": "10.x",
+    "@types/streamsaver": "^2.0.4",
```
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in[027bc6d](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/027bc6dd1c70892b7b14a5f81606886c1f5dd76f)

packages/core/services/FileStorageServiceBase/index.ts
BrianWhitneyAI: resolved in [10d7d67](https://github.com/AllenInstitute/biofile-finder/pull/204/commits/10d7d67753c85570a72d61de8bc8a191b4be552d)

BrianWhitneyAI:
Ok so as you both mentioned @pgarrison @SeanLeRoy , this PR is getting pretty big. My proposal is the following coverage 

### In this PR:

- Desktop download for s3 and local (basic)
- Manifest of files from S3 - (This is required for any type of downloading multifile)
- Web download for S3 w/ Streamsaver or equivalent

### In other PRS:
 - Local web download
 - S3 Protocol file support for Web and Desktop
 - Byte progress download for Desktop
 - Azure support
 - Cache clearing key

packages/web/src/services/FileDownloadServiceWeb.ts
```diff
@@ -1,76 +1,119 @@
...
-        const keys = await this.listS3Objects(bucket, key, region);
+        const { hostname, key } = this.parseS3Url(fileInfo.path);
+        const keys = await this.listS3Objects(hostname, key);
 
         if (keys.length === 0) {
             throw new Error("No files found in the specified S3 directory.");
         }
 
-        const zip = new JSZip();
+        const fileStream = StreamSaver.createWriteStream(`${fileInfo.name}.zip`);
```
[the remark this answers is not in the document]
BrianWhitneyAI: Confirmed that the zip is just all the files stitched together. The original method I had I tested and extracted the output. With this new system, I have just been looking at the exported zip file size since the feature isnt finished. Trying to unzip it now does not work :(. Looking into other options

BrianWhitneyAI:
This PR has gone through some changes. As it stands we have:
- [ ] Desktop download for s3 and local (basic)
- [ ] Manifest of files from S3 - (This is required for any type of downloading multifile)
- [ ] Web download for S3 w/ Streamsaver or equivalent

We were not able to implement a streamed download for zarr files from a remote address. The current method reads the full file into memory. Some cursory testing has shown that this works effectively for files under 1 GB. This is because of browsers inherent memory limits. 

Additionally this method suffers from having to build the zip before publishing and does not give a conventional status update. We will need to implement some sort of progress tracking or a spinner for the process in the future.

packages/core/services/FileDownloadService/index.ts
```diff
@@ -17,36 +21,76 @@ export interface FileInfo {
...
+     * List components of S3 directory.
      */
-    cancelActiveRequest(downloadRequestId: string): void;
+    public async listS3Objects(hostname: string, prefix: string): Promise<string[]> {
+        const url = `https://${hostname}?list-type=2&prefix=${encodeURIComponent(prefix)}`;
+        const response = await axios.get(url);
+        const parser = new DOMParser();
+        const xmlDoc = parser.parseFromString(response.data, "text/xml");
+
+        const keys: string[] = [];
```
SeanLeRoy, line 85: Could be more succinct / javascripty
```
return xmlDoc
    .getElementsByTagName("Key")
    .map(content => content.textContent || "");
```

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -66,7 +63,35 @@ export default class FileDownloadServiceElectron
...
+                return this.copyLocalZarrDirectory(
+                    fileInfo,
+                    downloadRequestId,
+                    onProgress,
+                    destination
+                );
+            }
+
+            throw new DownloadFailure(
+                `Invalid path for Zarr file: ${fileInfo.path}`,
```
SeanLeRoy, line 90:
```suggestion
                `Unsupported path for ".zarr":  ${fileInfo.path}. Currently only support AWS S3 or locally stored files.`,
```

SeanLeRoy (approved):

aswallace:
I'm a bit concerned that the rebase is indicating 40 files changed, is the branch fully up-to-date?

BrianWhitneyAI:
> I'm a bit concerned that after rebase it's showing 40 files changed, is the branch fully up-to-date? I may be misunderstanding what it's indicating

I think so. I pulled the main origin rebased and pushed the new origin to get the clear cache features.

aswallace (approved):
I don't seem to be able to open the downloaded files in agave, but that may be user error (not super familiar with zarrs or agave yet)

BrianWhitneyAI:
> I don't seem to be able to open the downloaded files in agave, but that may be user error (not super familiar with zarrs or agave yet)

Are you using the variance data set? if so the files are named ome.tiff and their download is a zarr file. for some reason, the folders are appended with .tiff so then Agave can't read them. If you click on the folder from your desktop and remove the tag it loads. I thought this was a pretty fringe case and is technically an issue with the data source so I didnt make any big changes around it.

PR #211, opened 2024-08-12
Fix splitting on non-strings as annotations

SeanLeRoy:
Antoine noticed a bug where the attached dataset wouldn't display values when grouped by "Used for" + "Gene" + "Pixel Size X". Turns out this is because a spot where we were checking to see if a value was a list we didn't cast the value from the DB into a string first and numbers can't be split like that. I didn't see any other spots where we were doing this so its just the one line changed.

Test it out on staging with the attached CSV
[Dataset_1_V14_V3.csv](https://github.com/user-attachments/files/16590927/Dataset_1_V14_V3.csv)

### DIFF
--- a/packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
+++ b/packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
@@ -109,7 +109,9 @@ export default class DatabaseAnnotationService implements AnnotationService {
         const rows = await this.databaseService.query(sqlBuilder.toSQL());
         const rowsSplitByDelimiter = rows
             .flatMap((row) =>
-                isNil(row[annotation]) ? [] : row[annotation].split(DatabaseService.LIST_DELIMITER)
+                isNil(row[annotation])
+                    ? []
+                    : `${row[annotation]}`.split(DatabaseService.LIST_DELIMITER)
             )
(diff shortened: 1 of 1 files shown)


aswallace (approved):

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -109,7 +109,9 @@ export default class DatabaseAnnotationService implements AnnotationService {
         const rows = await this.databaseService.query(sqlBuilder.toSQL());
         const rowsSplitByDelimiter = rows
             .flatMap((row) =>
-                isNil(row[annotation]) ? [] : row[annotation].split(DatabaseService.LIST_DELIMITER)
+                isNil(row[annotation])
+                    ? []
+                    : `${row[annotation]}`.split(DatabaseService.LIST_DELIMITER)
```
kmitcham, line 114: is this safe from annotations that contain ','?  eg, Baseball Teams: ["tom,dick,harry","larry,curly,moe"]

SeanLeRoy: Nope! The hope is to provide an option to choose the delimeter in the future

PR #213, opened 2024-08-13
Remove footer from app and update styling

aswallace:
### Context
We only need the cookie settings button available on the web pages, not on the app itself, so removing the footer from /app per recent UX discussions since it's visually annoying

### Changes
Because of the way react router works, /app is tied up with the `Outlet` component that's also used for webpages, so the easiest way to conditionally hide footer is to check the route.

To do: The footer will eventually need some styling updates but leaving as is for now to prioritize other tasks

### DIFF
--- a/packages/core/services/FileService/DatabaseFileService/index.ts
+++ b/packages/core/services/FileService/DatabaseFileService/index.ts
@@ -80,6 +80,7 @@ export default class DatabaseFileService implements FileService {
     }
 
     public async getCountOfMatchingFiles(fileSet: FileSet): Promise<number> {
+        if (!this.dataSourceNames.length) return 0;
         const select_key = "num_files";
         const sql = fileSet
             .toQuerySQLBuilder()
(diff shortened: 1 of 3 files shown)


packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -110,6 +111,7 @@ export default class DatabaseFileService implements FileService {
      * and potentially starting from a particular file_id and limited to a set number of files.
      */
     public async getFiles(request: GetFilesRequest): Promise<FileDetail[]> {
+        if (!this.dataSourceNames.length) return [];
```
aswallace, line 114: The changes in this file are actually totally unrelated so I can definitely separate it into its own ticket, but I've found that I frequently can't load the app at all (fails on render) unless these two lines are there. Is this just on mine/is there another way around this?

kmitcham (approved):

packages/core/services/FileService/DatabaseFileService/index.ts
aswallace: Separating this into its own ticket bc relates to the issue Tyler is having on Mac

SeanLeRoy (approved):

PR #215, opened 2024-08-14
Bugfix/fix csv import format

SeanLeRoy:
This bugfix came from Antoine where he was noticing `Clone` could not be grouped by, it seems it was failing bc we don't support querying by numbers very well with the newest regex. We should be importing everything as a varchar anyway since we don't support actually filtering by types outside of AICS data yet; so this changes the import strategy to make everything a varchar by default.

The two styling changes are not relevant but they have been _bugging_ me, they fix the gradient at the bottom of the UI to be back to its original shape (constrained to the file details pane) and make it so Windows users don't see a scrollbar in the query group rows

### DIFF
--- a/packages/core/components/FileDetails/FileDetails.module.css
+++ b/packages/core/components/FileDetails/FileDetails.module.css
@@ -58,7 +58,7 @@
     content: " ";
     height: 50px;
     right: 0;
-    max-width: 100%;
+    max-width: var(--file-details-width);
     position: absolute;
     width: 100%;
     z-index: 11;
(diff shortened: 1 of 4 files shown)


aswallace (approved):
NICE those had been bugging me too. 
Not sure if this is related but also having trouble filtering on certain annotations with weird characters, like selecting more than one value when filtering on Annotation by Timepoint

kmitcham (approved):

SeanLeRoy:
> Not sure if this is related but also having trouble filtering on certain annotations with weird characters, like selecting more than one value when filtering on Annotation by Timepoint

I don't think so, but I'll make a note of this to look into

PR #218, opened 2024-08-15
Remove automatic link detection and force users to specify manually

SeanLeRoy:
## Backstory
Antoine requested that some "link-like" annotation values not be displayed as links and not be provided in the open-with menu. We were just assuming any value starting with "http" was a link, but that wasn't specific enough for this use case since these are download links that are just for programmatic use.

## Description
This forces users to supply the data type in the additional metadata CSV alongside the optional descriptions rather than assuming any "http" value is a link

## Testing Data
Here is his image data CSV and the related description CSV:
[image_data.csv](https://github.com/user-attachments/files/16619804/imaging_and_segmentation_data_FOR_Matheus_test.csv)
[description.csv](https://github.com/user-attachments/files/16627513/Imaging_and_segmentation_data_column_definition.csv)

## Open questions
* ~Does this cause us to miss out on an easy opportunity to render links for people~ _**After some more thinking, we could always go back to treating everything as a link, this new way of doing things is actually more conservative so it would be compatible**_
* ~Should we be even more specific like "Open link" and "Download link" as the types or is that an over-complication?~ **_Had another meeting where this was dubbed to be the more future-proof decision._** see https://github.com/AllenInstitute/biofile-finder/pull/218/commits/d083aa20a15876c6a69dc0d9aaa5ceada05790bf

### DIFF
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -14,10 +14,10 @@ export default abstract class DatabaseService {
     public static readonly LIST_DELIMITER = ",";
     // THIS CAN NEVER CHANGE, THIS VALUE IS INCLUDED IN THE EMT DATA RELEASE
     // AS A MARKER FOR WHAT IS AN ACTUAL LINK
-    private static readonly LINK_TYPE = "Link";
+    private static readonly OPEN_FILE_LINK_TYPE = "Open file link";
     private static readonly ANNOTATION_TYPE_SET = new Set([
         ...Object.values(AnnotationType),
-        DatabaseService.LINK_TYPE,
+        DatabaseService.OPEN_FILE_LINK_TYPE,
     ]);
     private currentAggregateSource?: string;
     // Initialize with AICS FMS data source name to pretend it always exists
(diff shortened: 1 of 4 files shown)


packages/core/services/DatabaseService/index.ts
```diff
@@ -12,6 +12,13 @@ import SQLBuilder from "../../entity/SQLBuilder";
 export default abstract class DatabaseService {
     protected readonly SOURCE_METADATA_TABLE = "source_metadata";
     public static readonly LIST_DELIMITER = ",";
+    // THIS CAN NEVER CHANGE, THIS VALUE IS INCLUDED IN THE EMT DATA RELEASE
```
kmitcham, line 15: Maybe expand this with a link to the data release location, and the date?  I'm just thinking the new developer looking at this in 2027, with no context is going to be very confused.   Also, with no context, 'This' is unclear.  
maybe: "Link as a datatype must be hardcoded, and can not change without breaking visibility in the dataset released in 2024 as part of the paper at http://some.s3.url.dataset"

Hmm.  Could we make a release tag of BFF that goes with the paper?  That might make sense if BFF is going to be eternally tied to the paper release.

SeanLeRoy: > Maybe expand this with a link to the data release location, and the date?

Definately, good idea

> Could we make a release tag of BFF that goes with the paper? 

No, but we could include a Git hash reference... unsure anyone would be able to use that in an a helpful way though.

kmitcham (approved):

packages/core/services/DatabaseService/index.ts
toloudis: Also, we could certainly add other string names that mean "OPEN_FILE_LINK_TYPE" in the future, if data creators dislike that version for whatever reason.

toloudis (approved):
Does this code change implement the distinction between "open link" and "download link"?  If so then I am misunderstanding the code changes.  Or is that going in a separate pr?

aswallace (approved):

BrianWhitneyAI (approved):

packages/core/services/DatabaseService/index.ts
SeanLeRoy: I updated the language in the comment as suggested, but I can't include a link to the paper until it is published so I made a note for myself to add that in later.

PR #223, opened 2024-08-21
Feature/Clear-Cache

BrianWhitneyAI:
We have been running into issues with cached data, this adds a key command for clearing the cache.


BrianWhitneyAI:
This is not working yet. The Key down works but just has a log right now rather than an actual clear.

packages/core/hooks/useKeyDown/useKeyDown.ts
```diff
@@ -0,0 +1,23 @@
+import * as React from "react";
+
+export function useKeyDown(targetKeys: string[], handler: (event: KeyboardEvent) => void) {
```
SeanLeRoy, line 3: When there is only one export, making it the default is generally best
```suggestion
export default function useKeyDown(targetKeys: string[], handler: (event: KeyboardEvent) => void) {
```

packages/desktop/src/renderer/index.tsx
```diff
@@ -20,7 +17,7 @@ import FileViewerServiceElectron from "../services/FileViewerServiceElectron";
 import PersistentConfigServiceElectron from "../services/PersistentConfigServiceElectron";
 import NotificationServiceElectron from "../services/NotificationServiceElectron";
 import { GlobalVariableChannels } from "../util/constants";
-
+import { useKeyDown } from "../../../core/hooks/useKeyDown/useKeyDown"; // Import the useKeyDown hook
```
SeanLeRoy, line 20: I _think_ `hooks` contains exports for other hooks so it would be cool to simplify this import like this if so
```suggestion
import { useKeyDown } from "../../../core/hooks";
```

packages/desktop/src/renderer/index.tsx
```diff
@@ -30,7 +27,20 @@ const persistentConfigService = new PersistentConfigServiceElectron();
 const applicationInfoService = new ApplicationInfoServiceElectron();
 const databaseService = new DatabaseServiceElectron();
 const executionEnvService = new ExecutionEnvServiceElectron(notificationService);
-// application analytics/metrics
+
+// Define the KeyDownHandler component inline
+const KeyDownHandler: React.FC<{ clearStore: () => void }> = ({ clearStore }) => {
```
SeanLeRoy, line 32: I think creating a component just to wrap a side-effect is more work than necessary. In `renderFmsFileExplorer` you should be able to add the `useKeyDown` hook call right before the `render`

packages/desktop/src/services/PersistentConfigServiceElectron.ts
```diff
@@ -10,9 +9,7 @@ import { find } from "lodash";
 
 const OPTIONS: Options<Record<string, unknown>> = {
     // Defines a validation schema for data inserted into the persistent storage
-    // if a breaking change is made see migration patterns in elecron-store docs
     schema: {
-        // Deprecated in 6.0.0
```
SeanLeRoy, line 15: Were these comments removed intentionally?

packages/desktop/src/services/PersistentConfigServiceElectron.ts
```diff
@@ -139,15 +132,32 @@ export default class PersistentConfigServiceElectron implements PersistentConfig
     }
 
     public getAll(): PersistedConfig {
+        // Initialize default values
+        const defaultConfig: PersistedConfig = {
```
SeanLeRoy, line 136: Do we need to add this default here? Shouldn't all values be optional since a first-time user wouldn't have any of these values?

packages/desktop/src/services/PersistentConfigServiceElectron.ts
BrianWhitneyAI: nope! added back in [38e7ac2](https://github.com/AllenInstitute/biofile-finder/pull/223/commits/38e7ac251efb44254d2eb9eee39389e2b01ecc1b)

packages/desktop/src/services/PersistentConfigServiceElectron.ts
BrianWhitneyAI: restored to no defaults in [38e7ac2](https://github.com/AllenInstitute/biofile-finder/pull/223/commits/38e7ac251efb44254d2eb9eee39389e2b01ecc1b)

packages/core/hooks/useKeyDown/useKeyDown.ts
BrianWhitneyAI: resolved in [700014c](https://github.com/AllenInstitute/biofile-finder/pull/223/commits/700014c75843ccf83e48c6f8b6ba1b32ad47c55b)

packages/desktop/src/renderer/index.tsx
BrianWhitneyAI: I didn't see an export location in hooks. For the other hooks like useHelpOptions it looks like we import directly.

packages/desktop/src/renderer/index.tsx
BrianWhitneyAI: Does this still apply now that we are reloading within the component?

pgarrison:
Does Ctrl+Shift+R (Cmd+Shfit+R on MacOS) not achieve what you want? The purpose of this PR is a little fuzzy to me.

BrianWhitneyAI:
> Does Ctrl+Shift+R (Cmd+Shfit+R on MacOS) not achieve what you want? The purpose of this PR is a little fuzzy to me.

Not quite. In the desktop version, we use a persistent store to save out configurations in between sessions. Sometimes in development, we save a bad element, after which it is hard to continue developing because the bad property is rehydrated at the start. This PR adds a key command to clear the cache that we have completely.

packages/desktop/src/renderer/index.tsx
```diff
@@ -30,7 +27,20 @@ const persistentConfigService = new PersistentConfigServiceElectron();
...
+    useKeyDown(["Control", "-"], clearStore);
```
aswallace, line 33: Since we're able to customize the app menu options in desktop/src/main/menu, could we expose this command there? Maybe in "View" if possible since that's where Reload, Force reload, and Toggle dev tools are.

Also just to verify, control is the same in both windows and mac right?

packages/desktop/src/renderer/index.tsx
pgarrison: I think so? `window.location.reload()` is such a major side effect (wiping the whole DOM) the behavior is probably the same regardless of whether it's in a component?

packages/desktop/src/renderer/index.tsx
BrianWhitneyAI: control is a common key. There could be some weirdness between OS for sure. I tested it with my macbook and with a external keyboard that built for windows and it seemed to be fine.

BrianWhitneyAI: Open to adding it to View!

pgarrison (approved):
Overall looks good to me!

packages/desktop/src/renderer/index.tsx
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in 
b45a3bc

packages/core/services/DatabaseService/index.ts
```diff
@@ -210,7 +210,7 @@ export default abstract class DatabaseService {
         } catch (err) {
             // Source metadata file may not have been supplied
             // and/or this column may not exist
-            const errMsg = (err as Error).message;
+            const errMsg = typeof err === "string" ? err : err instanceof Error ? err.message : "";
```
SeanLeRoy, line 213: I've never seen an error thrown as a `string`

packages/desktop/src/renderer/index.tsx
SeanLeRoy: Yes, this is still relevant. Essentially it is bad style to create a component for the purpose of a side-effect. Components are for rendering elements.

SeanLeRoy (approved):

packages/desktop/src/renderer/index.tsx
BrianWhitneyAI: Update on this. Hooks have to be called within a component according to this error:

```
Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
    at resolveDispatcher (react.development.js:1476)
    at Object.useEffect (react.development.js:1519)
    at useKeyDown (index.ts:4)
    at renderFmsFileExplorer (index.tsx:108)
    at Module../src/renderer/index.tsx (index.tsx:130)
    at __webpack_require__ (bootstrap:24)
    at startup:6
    at startup:6
```

Issue #235, opened 2024-09-11
Using wrong font


aswallace:
Even though we have `Open Sans` listed in our stylesheet, Chrome and other browsers are actually rendering their default sans fonts (in my case, Helvetica). We're apparently using a broken stylesheet link in `global.css`. 

Network calls to font api:
- Chrome: 400 Bad Request
- Firefox: The resource from <font link> blocked due to MIME type (“text/html”) mismatch 
- Safari: fails with no info (ty Safari)

### Recommended change: 
Replace current links with `"https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300..700;1,300..600&display=swap"` (link suggested by [Google Font API](https://fonts.google.com/selection/embed)) in index.html and global.css

May also require styling fixes, because we've been modifying font styling away from the Figma designs under the (incorrect) assumption that we were using the correct font and that it just LOOKED different on our site. This means should also verify instances of `font-weight` and `font-size` to make sure they match Figma specs

PR #238, opened 2024-09-13
Feature/update special filter UI and create any/none/fuzzy query params

### LINKED ISSUE #53: Rather than requiring all Filename (maybe file ID too) be fuzzy searches make it optional
Lots of search inputs have toggles for how the query is done, a toggle for turning fuzzy search on or off would be good here, requires work in the file explorer-service

### LINKED ISSUE #74: Add "any value" filter (AKA hierarchy filter)
Currently the only way to check if files have any value for a particular annotation is by using hierarchies (group by). Users should also be able to apply this as a filter-- e.g., instead of having to select specific values, should be able to get files that have any value for that annotation.

Relevant for #96. Closely related to #187.

aswallace:
### Description
A Big Ticket! This PR includes:
- ~UX updates to the special filter types (date, numerical range, search box with fuzzy search)~
- Logic for fuzzy searching (transferred over from #154 )
- Logic for any/none filters (include/exclude query params)
   - Corresponding UI for this in annotation filter form
  
closes #53 , closes #74 

~TODO: fix font (#235 )~

### Testing
- New & updated unit tests
- Tested manually with an EC2 instance of FES (the local version has the fuzzy search changes, but hasn't been deployed) and staging version of Mongo.

### Screenshots
EDIT: Updated screenshots 9/30/2024
#### Any/none option
~The radio buttons are a little off center with Helvetica, but I think it will be fixed when we're using the correct Open Sans link~
any value

no value

#### Search box with new fuzzy search toggle (exact/non-fuzzy by default)
fuzzy search

#### Date range with filter type choice group
date range modal

#### Numerical range with filter type choice group
numerical range modal


packages/core/components/ListPicker/index.tsx
```diff
@@ -172,8 +172,7 @@ export default function ListPicker(props: ListPickerProps) {
             </div>
             <div className={styles.footer}>
                 <p>
-                    {/* (item.length -1) to account for buffer in item list. */}
```
aswallace, line 175: We don't actually use the buffer item, so this was returning incorrect -1 value

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -30,11 +41,32 @@ interface AnnotationFilterFormProps {
...
     const allFilters = useSelector(selection.selectors.getFileFilters);
+    const fuzzyFilters = useSelector(selection.selectors.getFuzzyFilters);
+    const anyValueFilters = useSelector(selection.selectors.getIncludeFilters);
+    const noValueFilters = useSelector(selection.selectors.getExcludeFilters);
     const annotationService = useSelector(interaction.selectors.getAnnotationService);
     const [annotationValues, isLoading, errorMessage] = useAnnotationValues(
         props.annotation.name,
         annotationService
     );
+    const usingAnyValueFilter: boolean = React.useMemo(() => {
```
BrianWhitneyAI, line 52: This logic is reused a lot could do something like 
```
const isUsingFilter = (filters: Filter[], annotationName: string) => {
    return filters?.some((filter) => filter.annotationName === annotationName) || false;
};
```

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -85,10 +132,55 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
                         : props.annotation.valueOf(item.value)
                 )
         );
+        removeSpecialFilters();
         dispatch(selection.actions.addFileFilter(filters));
     };
 
+    // Any value/"include"
+    const onSelectAny = () => {
```
BrianWhitneyAI, line 140: Idk if it's worth it since its just two options but there's a lot of reused stuff here. could combine to something like 
```
function applyFilter(filterType: "include" | "exclude") {
    removeSpecialFilters();
    dispatch(selection.actions.removeFuzzyFilter(new FuzzyFilter(props.annotation.name)));
    onDeselectAll();

    if (filterType === "include" && !usingAnyValueFilter) {
        const includeFilter = new IncludeFilter(props.annotation.name);
        dispatch(selection.actions.addIncludeFilter(includeFilter));
    } else if (filterType === "exclude" && !usingNoValueFilter) {
        const excludeFilter = new ExcludeFilter(props.annotation.name);
        dispatch(selection.actions.addExcludeFilter(excludeFilter));
    }
}

const onSelectAny = () => applyFilter("include");
const onSelectNone = () => applyFilter("exclude");
```
Or alternatively moving the dispatch and onDeselect all into removeSpecialFilters() Also afterthought, would there be possibility for more filter types in the future?

packages/core/components/QueryPart/QueryFilter.tsx
```diff
@@ -54,15 +62,36 @@ export default function QueryFilter(props: Props) {
...
+                .map(([annotationName, filters]) => {
+                    const operator = filters.length > 1 ? "ONE OF" : "EQUALS";
+                    const valueDisplay = map(filters, (filter) => filter.displayValue).join(", ");
+                    return {
+                        id: filters[0].name,
+                        title: `${annotationName} ${operator} ${valueDisplay}`,
+                        description: annotationNameToAnnotationMap[annotationName]?.description,
+                    };
+                })
+                .concat(
```
BrianWhitneyAI, line 75: could use spread rather than chaining concats here

packages/core/state/selection/actions.ts
```diff
@@ -75,6 +78,187 @@ export function removeFileFilter(filter: FileFilter | FileFilter[]): RemoveFileF
...
 }
 
+/**
+ * SET_FUZZY_FILTERS
+ *
+ * Intention to set, wholesale, a list of FuzzyFilters into application state. This should not be dispatched
+ * by UI components; dispatch either an ADD_FUZZY_FILTER or REMOVE_FUZZY_FILTER action. Those actions will
+ * trigger the `modifyFuzzyFilters` logic, which will then dispatch this action.
+ */
+export const SET_FUZZY_FILTERS = makeConstant(STATE_BRANCH_NAME, "set-fuzzy-filters");
```
BrianWhitneyAI, line 88: This is another instance where there is so much similarity between fuzzy include and exclude. Is there a considerable refactor that could happen here? seeing a lot of code in triplicate.

packages/core/state/selection/actions.ts
```diff
@@ -75,6 +78,187 @@ export function removeFileFilter(filter: FileFilter | FileFilter[]): RemoveFileF
     };
 }
 
+/**
+ * SET_FUZZY_FILTERS
+ *
+ * Intention to set, wholesale, a list of FuzzyFilters into application state. This should not be dispatched
```
BrianWhitneyAI, line 84: what does wholesale mean here?

BrianWhitneyAI (approved):
Nice! My biggest comments are on the similarity between the include/exclude/fuzzy filter logic that shows up in duplicate or triplicate frequently. Not sure if there is a way around it

packages/core/state/selection/actions.ts
aswallace: Absolutely, I'd ideally like to remove all the duplicate code and streamline the different filter types. It might be worth making it a separate refactor ticket though, since this feature has been open for a long time now (especially fuzzy search)

packages/core/state/selection/actions.ts
aswallace: copy pasta comment from a different filter type. I believe intended to mean "everything at once instead of individual changes"

packages/core/components/AnnotationFilterForm/index.tsx
aswallace: Only minor concern with this solution is we'd be doing extra dispatches if the filter is already applied, which is why I originally had them in the conditionals. Not going to break anything, but would be extra unnecessary calls to state.
Similarly, would prefer not to move onDeselectAll into removeSpecialFilters with how we currently have it, since it's also called in situations where we don't do a deselectAll, so we'd also end up doing extra dispatches.
This whole file can definitely use a substantive refactor though, the inter-function dependencies are pretty messy

packages/core/components/AnnotationFilterForm/index.tsx
aswallace: The "anyValue" naming is a little misleading here, we're actually checking if we're using an `INCLUDE` (any value) filter rather than a `Filter` of any value

packages/core/entity/FileSelection/test/FileSelection.test.ts
```diff
@@ -674,6 +677,9 @@ describe("FileSelection", () => {
             const fileSet1 = new FileSet();
             const fileSet2 = new FileSet({
                 filters: [new FileFilter("foo", "bar")],
+                fuzzyFilters: [new FuzzyFilter("fuzzyfoo1"), new FuzzyFilter("fuzzyfoo2")],
```
kmitcham, line 680: in cases like this, I like using string variables, e.g. fuzzyFilterValue1 = "fuzzyfoo1", or just setting the names to something more meaningful, like FuzzyFilter("fuzzyFilterValue1") so the validation is easier to follow.

packages/core/entity/FileSet/test/FileSet.test.ts
```diff
@@ -29,6 +38,26 @@ describe("FileSet", () => {
...
+            );
+        });
+
+        // Enforce query param order for cacheing efficiency. Same args should register as same query regardless of order
+        it("includes sort after name-only filters in query string", () => {
+            const fileSet = new FileSet({
+                fuzzyFilters: [fuzzyFileName],
+                sort: dateCreatedDescending,
+            });
+            expect(fileSet.toQueryString()).equals("fuzzy=file_name&sort=date_created(DESC)");
```
kmitcham, line 58: would a second test, with the args in the opposite order, be relevant?

packages/core/entity/FileSelection/test/FileSelection.test.ts
aswallace: I was being lazy. Will update

kmitcham (approved):

packages/core/components/AnnotationFilterForm/index.tsx
aswallace: Slightly refactored this into one function in [69f2d81](https://github.com/AllenInstitute/biofile-finder/pull/238/commits/69f2d81cd0bc08021ab4bc01108dcd64a4a5e048). Not 100% sure how much of improvement it is

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -30,17 +41,39 @@ interface AnnotationFilterFormProps {
...
+    const includeFilters = useSelector(selection.selectors.getIncludeFilters);
+    const excludeFilters = useSelector(selection.selectors.getExcludeFilters);
     const annotationService = useSelector(interaction.selectors.getAnnotationService);
     const [annotationValues, isLoading, errorMessage] = useAnnotationValues(
         props.annotation.name,
         annotationService
     );
 
+    // Check if annotation name is already listed in state as an INCLUDE, EXCLUDE, FUZZY, or regular file filter
+    const usingIncludeFilter: boolean = React.useMemo(() => {
```
SeanLeRoy, line 54: type is implicit for all of these

```
const usingIncludeFilter = React.useMemo(() => (
    !!includeFilters?.some(filter => filter.annotationName === props.annotation.name)
), [includeFilters, props.annotation]);
```

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -30,17 +41,39 @@ interface AnnotationFilterFormProps {
...
+        return !!includeFilters?.some((filter) => filter.annotationName === props.annotation.name);
+    }, [includeFilters, props.annotation]);
+    const usingExcludeFilter: boolean = React.useMemo(() => {
```
SeanLeRoy, line 57: I think the language of include and exclude is a bit ambiguous here since we might add the ability to exclude certain annotation values in the future. IMO I think `isExcludingAllAnnotationValues` & `isIncludingAllAnnotationValues` or something along those lines would be clearer.

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -30,17 +41,39 @@ interface AnnotationFilterFormProps {
...
+    const fuzzySearchEnabled: boolean = React.useMemo(() => {
+        return !!fuzzyFilters?.some((filter) => filter.annotationName === props.annotation.name);
+    }, [fuzzyFilters, props.annotation]);
     const filtersForAnnotation = React.useMemo(
         () => allFilters.filter((filter) => filter.name === props.annotation.name),
         [allFilters, props.annotation]
     );
 
+    // Rehydrate filter type from state into UI
+    const defaultFilterType = () => {
```
SeanLeRoy, line 69: Why is this a function? It looks like we use it only once to pass it to `useState` below (which btw I didn't know you could provide a function to?).

packages/core/components/ChoiceGroup/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+        option?: IChoiceGroupOption | undefined
+    ) => void;
+    options: IChoiceGroupOption[];
+    defaultSelectedKey?: string | number;
+}
+
+/**
+ * Custom styled wrapper for default fluentui component
+ */
+export default function BaseCheckbox(props: Props) {
```
SeanLeRoy, line 20: Component name doesn't match file
```suggestion
export default function ChoiceGroup(props: Props) {
```

packages/core/components/NumberRangePicker/index.tsx
```diff
@@ -50,10 +51,11 @@ export default function NumberRangePicker(props: NumberRangePickerProps) {
...
     );
 
+    // Instead of removing filter completely, reset to min and max and submit
     function onResetSearch() {
-        onReset();
-        setSearchMinValue("");
-        setSearchMaxValue("");
+        setSearchMinValue(overallMin);
+        setSearchMaxValue(overallMax);
+        onSearch(`RANGE(${overallMin},${overallMax})`);
```
SeanLeRoy, line 58: Debatable if this is in the scope of this PR but I think its odd we pass a string version of a range filter to the parent component rather than something like `onSearch(min, max)`

packages/core/entity/SimpleFilter/index.ts
SeanLeRoy, line 1: I feel like the decision to make the `ANY`, `EXCLUDE EVERYTHING`, and `FUZZY` filters totally separate has made the state logic of other components much more complicated. I like the idea of each having their own class, but could they each have a "type" property that we could then use in the components to separate out the filters we need? I'm unsure what we gain by having them separated since we can create composed selectors to separate them out of the state branch if need by like:
```
const getAnnotationsFilteredOut = createSelector(getFileFilters, filters) => (
    filters
        .filter(filter => filter.type === FilterType.EXCLUDE)
        .map(filter => filter.annotationName)
))

const getAnnotationsRequired = createSelector(getFileFilters, filters) => (
    filters
        .filter(filter => filter.type === FilterType.ANY)
        .map(filter => filter.annotationName)
))
```

SeanLeRoy (changes requested):
Pausing the merge of this PR to get feedback about the filter type division.

packages/core/components/NumberRangePicker/index.tsx
aswallace: Agreed, although onSearch is a pre-existing function also used for searchboxes so would need some additional changes to accept other args. `RANGE` as a filter value in general could use refactoring, we have to do some additional processing in number-formatter and date-formatter as well

packages/core/entity/SimpleFilter/index.ts
aswallace: I agree, but would need to think about how we'd incorporate fuzzy filters into this, since a filter can be both a regular `FileFilter` with a value AND a fuzzy filter

SeanLeRoy: That could be another type I think

aswallace:
### CHANGES FROM PREVIOUS VERSION
Added a new optional `type` arg to filters in [181067e](https://github.com/AllenInstitute/biofile-finder/pull/238/commits/181067e4d7839c65566dd2ddd8a52b9edc44eb92), so now filters can be any value, exclude (null), fuzzy, or default (regular filter) (open to changing the names of these). This allows us to get rid of a lot of the duplicate logic I had before, especially for redux and the AnnotationFilterForm component.

This means the FileExplorerUrl (our local BFF url parsing class) will NOT have the fuzzy/include/exclude params that I'd added in before, since now all filters are part of the `filter` param but with an extra `type` arg. However, the queries we send to FES do still need the new query params. 
- Because of this, I kept (but simplified) the FuzzyFilter, ExcludeFilter, and IncludeFilter subtypes,  bc even though they don't add new logic, they're a helpful shorthand for creating those filters

packages/core/state/selection/selectors.ts
```diff
@@ -41,6 +41,22 @@ export const isQueryingAicsFms = createSelector(
...
+
+export const getAnnotationsFilteredOut = createSelector([getFileFilters], (filters): FileFilter[] =>
+    filters.filter((filter) => filter.type === FilterType.EXCLUDE)
+);
+
+export const getAnnotationsRequired = createSelector([getFileFilters], (filters): FileFilter[] =>
+    filters.filter((filter) => filter.type === FilterType.ANY)
+);
+
+export const getDefaultFileFilters = createSelector([getFileFilters], (filters): FileFilter[] =>
```
aswallace, line 56: still unsure on the naming for these selectors,

kmitcham (approved):
Seems to still be reasonable.

SeanLeRoy (approved):
Looks great!! I'm psyched we were able to reduce some of the complexity from this.

packages/core/entity/FileExplorerURL/index.ts
```diff
@@ -148,7 +148,10 @@ export default class FileExplorerURL {
                 : undefined,
             filters: unparsedFilters
                 .map((unparsedFilter) => JSON.parse(unparsedFilter))
-                .map((parsedFilter) => new FileFilter(parsedFilter.name, parsedFilter.value)),
+                .map(
+                    (parsedFilter) =>
+                        new FileFilter(parsedFilter.name, parsedFilter.value, parsedFilter.type)
```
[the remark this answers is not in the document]
aswallace: yep! I gave the parent class a default value. Many of our unit tests still use FileFilter(name, value) without a type, so the constructor provides it with `FilterType.DEFAULT` if no arg is given.

Added tests in [d685394](https://github.com/AllenInstitute/biofile-finder/pull/238/commits/d685394f5e7800e206f6cada8585aee6a5d712d9)

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -53,48 +67,64 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
...
-            (item) =>
-                new FileFilter(
-                    props.annotation.name,
-                    isNil(props.annotation.valueOf(item.value))
-                        ? item.value
-                        : props.annotation.valueOf(item.value)
-                )
-        );
-        dispatch(selection.actions.addFileFilter(filters));
+    const _onFilterTypeOptionChange = (option: IChoiceGroupOption | undefined) => {
```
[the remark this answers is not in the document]
aswallace: truly don't remember, I think it might have been legacy. Removed in [d685394](https://github.com/AllenInstitute/biofile-finder/pull/238/commits/d685394f5e7800e206f6cada8585aee6a5d712d9)

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -31,17 +32,30 @@ interface AnnotationFilterFormProps {
...
     const dispatch = useDispatch();
     const allFilters = useSelector(selection.selectors.getFileFilters);
+    const fuzzyFilters = useSelector(selection.selectors.getFuzzyFilters);
     const annotationService = useSelector(interaction.selectors.getAnnotationService);
     const [annotationValues, isLoading, errorMessage] = useAnnotationValues(
         props.annotation.name,
         annotationService
     );
 
+    const fuzzySearchEnabled = React.useMemo(() => {
```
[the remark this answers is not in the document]
aswallace: changed in [d685394](https://github.com/AllenInstitute/biofile-finder/pull/238/commits/d685394f5e7800e206f6cada8585aee6a5d712d9)

BrianWhitneyAI (approved):

PR #242, opened 2024-09-17
Bugfix/fix broken font

### LINKED ISSUE #235: Using wrong font

aswallace:
### Context
Even though we had Open Sans listed in our stylesheet, Chrome and other browsers were actually rendering their default sans fonts (in my case, Helvetica). We were apparently using a broken stylesheet link in `global.css`, so network call was failing. 

closes #235 

### Changes
- Replaced current .css import link with one recommended by the [Google Font API](https://fonts.google.com/selection/embed). 
- Also updated the embed in `docs/index.html` to match Google API rec for good measure. We use fewer styling options in docs, so don't need the full `ital` etc link used in `global.css`
- Updated a couple minor font styling issues, and added the lang tag to our html files.

### Screenshots
Before & after (ignore weird cropping, margins are the same)
Home page before font changes
Home page after font changes

### DIFF
--- a/packages/core/styles/global.css
+++ b/packages/core/styles/global.css
@@ -1,4 +1,4 @@
-@import url('https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0..800&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300..700;1,300..600&display=swap');
 
 html {
     box-sizing: border-box;
--- a/docs/index.html
+++ b/docs/index.html
@@ -9,7 +9,7 @@
         <meta name="author" content="AICS Software">
         <link rel="preconnect" href="https://fonts.googleapis.com">
         <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-        <link href="https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300..700;1,300..600&display=swap" rel="stylesheet">
+        <link href="https://fonts.googleapis.com/css2?family=Open+Sans&family=Raleway:wght@100..600&display=swap" rel="stylesheet">
         <link rel="stylesheet" href="styles.css?v=1.0">
     </head>
 
--- a/docs/index.html
+++ b/docs/index.html
@@ -7,8 +7,9 @@
         <title>BioFile Finder</title>
         <meta name="description" content="BioFile Finder">
         <meta name="author" content="AICS Software">
-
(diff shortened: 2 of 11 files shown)


BrianWhitneyAI (approved):

packages/core/styles/global.css
```diff
@@ -1,4 +1,4 @@
-@import url('https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0..800&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300..700;1,300..600&display=swap');
```
SeanLeRoy, line 1: Doesn't this read as equivalent to 0..800?

SeanLeRoy (approved):
LGTM other than one comment

packages/core/styles/global.css
aswallace: No, 0/1 is the indicator for italics, so the tuples can be interpreted as `{{ital 0.0},{wght 300.0-700.0}}` and  `{{ital 1.0},{wght 300.0-600.0}}` where ital 1.0 = italicized and ital 0.0 = regular. Open Sans only goes down to 300, and we only use up to 700 for regular and 600 for ital, so I limited it to avoid over-importing

aswallace: The double tuple is also necessary regardless of whether we use the same font range for both italics and non-italics. If we just fix the old url so it has the correct range format (https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300..800&display=swap) but don't add in the second tuple, it only imports the normal (non-italic) fonts. So even if we want up to 800 for both, it would need to be https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300..800;1,300..800&display=swap

PR #243, opened 2024-09-18
feature/tracked-zarr-download-desktop

### LINKED ISSUE #236: Download progress for desktop zarr files (remote and local)
The current Zarr download method does not show the progress of the download. This is particularly confusing with multifile where subsequent lower files appear asynchronously.

BrianWhitneyAI:
## Description
The purpose of this PR is to add byte download tracking for S3 Zarr files. This is a companion PR to (#204) which added download support for desktop. That PR was too large to cover all features.

### Related Issues
 Resolves #236

### DIFF
--- a/packages/core/services/FileDownloadService/index.ts
+++ b/packages/core/services/FileDownloadService/index.ts
@@ -1,7 +1,4 @@
 import axios from "axios";
-import * as http from "http";
-import * as https from "https";
-
 import HttpServiceBase from "../HttpServiceBase";
 
 export enum DownloadResolution {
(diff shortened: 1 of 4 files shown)


BrianWhitneyAI:
We have decided to forgo individual local file tracking for now. In the situation that we use a /allen path and want to download locally we should use the FmsFile class's logic for determining cloud paths for all zarrs so that the existing download tracking functions correctly. This might prove more complicated as right now we do not support download paths for paths like `s3://allencell/`

BrianWhitneyAI:
Screenshot 2024-09-23 at 12 16 00 PM

BrianWhitneyAI:
Screenshot 2024-09-23 at 12 17 04 PM

kmitcham (approved):

packages/core/services/FileDownloadService/index.ts
```diff
@@ -1,4 +1,6 @@
 import axios from "axios";
+import * as http from "http";
+import * as https from "https";
```
SeanLeRoy, line 3: `http` & `https` are node libraries so we can't use them in the web version of the app, any logic that requires these libraries has to go into the `desktop` package. Probably easier to find a web friendly way to do this instead.

https://nodejs.org/api/http.html

SeanLeRoy (commented):
LGTM, minus the potential usage of node libraries. Will approve once we resolve that

packages/core/services/FileDownloadService/index.ts
BrianWhitneyAI: resolved in https://github.com/AllenInstitute/biofile-finder/pull/252!

SeanLeRoy (approved):

PR #245, opened 2024-09-23
Feature/default cloudpath zarrs

### LINKED ISSUE #244: Download zarrs from AICS FMS
We have decided to forgo individual local file tracking for now. In the situation that we use a /allen path and want to download locally we should use the FmsFile class's logic for determining cloud paths for all zarrs so that the existing download tracking functions correctly. This might prove more complicated as right now we do not support download paths for paths like `s3://allencell/`

_Originally posted by @BrianWhitneyAI in https://github.com/AllenInstitute/biofile-finder/issues/243#issuecomment-2368907369_

BrianWhitneyAI:
## Description 
 This PR defaults to using the cloudpath for zarr files. If the cloudpath does not exist it will fall back to path.

### Related Issues
Resolves #244.

### DIFF
--- a/packages/core/entity/FileDetail/index.ts
+++ b/packages/core/entity/FileDetail/index.ts
@@ -128,6 +128,7 @@ export default class FileDetail {
         // For AICS files we don't have permission to the bucket nor do we expect to have the /allen
         // drive mounted on the client machine. So we use the NGINX server to serve the file.
 
+        // Always try to use cloudpath for zarrs.
         if (this.path.endsWith(".zarr")) {
             return this.cloudPath;
         }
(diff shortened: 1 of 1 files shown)


BrianWhitneyAI:
Right now we only support S3 downloads and local copying on desktop. The current download path we are using allows for rendering thumbnails but not for download (since it doesnt meet our current S3 method). The cloudpaths we have for those files are permission blocked so we cant use that for download either. If we want to support download from AICS FMS we need to either change permissions of the cloudpaths or look into downloading other formatted paths (RN its a http). This could just point to a s3 file and be easy but would be another step.

PR #252, opened 2024-09-25
feature/tracked-zarr-download-web

### LINKED ISSUE #240: Download Spinner for Zarr Web download
The zarr web download does not report that it is in progress. Adding a spinner and changing the download button text to downloading might be a suitable solution.

BrianWhitneyAI:
## Description 

This PR was originally supposed to add a spinner for when the download was in progress, However in writing that element it became apparent that we could track the download via onProgress and use the in-place infrastructure for progress reporting rather than designing a new feature.

This PR also resolves an issue in FileDownloadService/index.ts where we were using http and https to retrieve file size. These are not available to web so I replaced it with a pure axios call.

Also cancelable zarr downloads for web!

## Relevant Issues
resolves #240

### DIFF
--- a/packages/core/services/FileDownloadService/index.ts
+++ b/packages/core/services/FileDownloadService/index.ts
@@ -1,7 +1,4 @@
 import axios from "axios";
-import * as http from "http";
-import * as https from "https";
-
 import HttpServiceBase from "../HttpServiceBase";
 
 export enum DownloadResolution {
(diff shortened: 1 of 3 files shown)


BrianWhitneyAI:
The caviat to doing it this way is that the file download progress bar will complete as the file is read into memory. (completion when full file is read). But will then take a wile to write out the zip. This leaves time proportional to the size of the file in-between when the progress bar hits 100% and the file is actully downloaded. 

The completion message will still show up after zipping though so I think its ok!

kmitcham (approved):

BrianWhitneyAI:
Screenshot 2024-09-25 at 10 16 10 AM
Screenshot 2024-09-25 at 10 15 49 AM
Screenshot 2024-09-25 at 10 11 06 AM
Screenshot 2024-09-25 at 10 10 03 AM

BrianWhitneyAI:
Screenshot Desctiption:

1) Canceled State
2)Completed State (after zipping)
3) Completed download (before Zipping) - will wait in this state until full sucess
4) in-Progress state

SeanLeRoy (approved):
🙌 LGTM! Good idea

PR #254, opened 2024-09-26
feature/cancelable-desktop-zarr-download

### LINKED ISSUE #253: Download cancelation for Desktop zarrs
The cancelation button does not work for desktop zarrs.

BrianWhitneyAI:
## Description 
This Pr adds the cancelable functionality to S3 desktop downloads as well as error handling around cancelation. It does not support local zarr downloads. If we want to support those we need to track file size locally when it is not provided by the csv/datasource.

### Relevant Issues
This Pr resolves #253

### DIFF
--- a/packages/desktop/src/services/FileDownloadServiceElectron.ts
+++ b/packages/desktop/src/services/FileDownloadServiceElectron.ts
@@ -429,12 +429,24 @@ export default class FileDownloadServiceElectron extends FileDownloadService {
                 throw new Error("No files found in the specified S3 directory.");
             }
 
+            let cancelRequested = false;
+
+            // Register cancellation token for this request
+            this.activeRequestMap[downloadRequestId] = {
+                filePath: fullDestination,
+                cancel: () => {
+                    cancelRequested = true;
+                },
+            };
+
             // Download each file, track its size, and report progress
             for (const fileKey of keys) {
-                if (!fileKey) {
-                    console.warn(`Encountered null or undefined file key. Skipping.`);
-                    continue;
+                // If cancel was requested, cleanup.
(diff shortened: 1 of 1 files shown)


packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -448,15 +460,27 @@ export default class FileDownloadServiceElectron extends FileDownloadService {
                         onProgress(fileDownloadedBytes);
                     }
                 });
+
+                // If cancel was requested, cleanup.
+                if (cancelRequested) {
```
kmitcham, line 465: This checks if cancel was requested during the work from line 448 to 462?  Otherwise we'd just check again at the start of the next item in the 'for'?

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -448,15 +460,27 @@ export default class FileDownloadServiceElectron extends FileDownloadService {
...
+            delete this.activeRequestMap[downloadRequestId];
+
             return {
                 downloadRequestId: fileInfo.id,
                 msg: `Successfully downloaded ${fileInfo.path} to ${fullDestination}`,
                 resolution: DownloadResolution.SUCCESS,
             };
         } catch (err) {
-            console.error(`Failed to download directory: ${err}`);
+            delete this.activeRequestMap[downloadRequestId];
```
kmitcham, line 480: Do the DownloadFailures thrown by cancellation end up in this catch?

kmitcham (approved):

SeanDuHare (approved):

PR #256, opened 2024-09-27
Add "Learn" page

SeanLeRoy:
**Description**
On the splash page we explain the features of the application to users, but not necessarily why they should use it. This is to hopefully aid the splash page with some more contextual information about why some of those features are there and why they might use this application. The FAQ is largely coming from questions asked often during beta tests. Over time we should learn more about what users would like to see here.


packages/web/assets/learn-diagram-csv.png
kmitcham, line 1: nit: Maybe 'Example Color': "red" "blue" and 'Example Size': "large" "medium" 
and file.txt -> largeRed.txt and file2.txt->mediumBlue.txt ?
Just so we don't have the word 'example' in the diagram 6 times?

packages/web/src/components/Learn/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+// Public-facing learning/about page
+export default function Learn() {
+    return (
+        <div className={styles.root}>
+            <div className={styles.banner}>
+                <div className={styles.bannerBlur}>
+                    <div className={classNames(styles.section, styles.titleContainer)}>
+                        <h1>Curate your data your way</h1>
+                        <p>
+                            BioFile Finder helps users effortlessly interact with, extrapolate, and
```
kmitcham, line 20: nit: 'effortlessly' is a very strong claim.  'Conveniently' or 'Easily'?  just a thought

kmitcham (approved):

packages/web/assets/learn-diagram-csv.png
kmitcham: If there are example categories and types that are related to the science but still very clear, that'd be even better, obviously.

aswallace (approved):
left some minor suggestions!

PR #262, opened 2024-10-07
Add include/exclude logic for database service

aswallace:
This adds include/exclude logic for external files that DON'T use FES (e.g., use database service with SQL queries). 
Edit for context: the logic for FES has already been merged via #238 

This does NOT implement fuzzy vs. exact search toggling yet for non-FES sources -- our current logic for the database file service still does fuzzy matching by default. Because of this, I'm conditionally hiding/disabling the fuzzy toggle in the UI element for non-AICS data sources.

### Testing
Tested manually for `any value` filters and `no value` filters using a test csv that contains null values.
There's also some very simple unit tests to make sure things still run with typed filters

### DIFF
--- a/packages/core/components/AnnotationFilterForm/index.tsx
+++ b/packages/core/components/AnnotationFilterForm/index.tsx
@@ -33,6 +33,7 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
     const dispatch = useDispatch();
     const allFilters = useSelector(selection.selectors.getFileFilters);
     const fuzzyFilters = useSelector(selection.selectors.getFuzzyFilters);
+    const isQueryingAicsFms = useSelector(selection.selectors.isQueryingAicsFms);
     const annotationService = useSelector(interaction.selectors.getAnnotationService);
     const [annotationValues, isLoading, errorMessage] = useAnnotationValues(
         props.annotation.name,
@@ -196,6 +197,7 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
                             fuzzySearchEnabled={fuzzySearchEnabled}
                             fieldName={props.annotation.displayName}
                             defaultValue={filtersForAnnotation?.[0]}
+                            hideFuzzyToggle={!isQueryingAicsFms}
                         />
                     );
                 }
--- a/packages/core/components/AnnotationFilterForm/index.tsx
+++ b/packages/core/components/AnnotationFilterForm/index.tsx
@@ -33,7 +33,7 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
     const dispatch = useDispatch();
     const allFilters = useSelector(selection.selectors.getFileFilters);
     const fuzzyFilters = useSelector(selection.selectors.getFuzzyFilters);
-    const isQueryingAicsFms = useSelector(selection.selectors.isQueryingAicsFms);
+    const isQueryingSourceWithFuzzy = useSelector(selection.selectors.isQueryingAicsFms);
     const annotationService = useSelector(interaction.selectors.getAnnotationService);
     const [annotationValues, isLoading, errorMessage] = useAnnotationValues(
         props.annotation.name,
@@ -197,7 +197,7 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
                             fuzzySearchEnabled={fuzzySearchEnabled}
                             fieldName={props.annotation.displayName}
                             defaultValue={filtersForAnnotation?.[0]}
-                            hideFuzzyToggle={!isQueryingAicsFms}
+                            hideFuzzyToggle={!isQueryingSourceWithFuzzy}
                         />
                     );
                 }
--- a/packages/core/components/AnnotationFilterForm/index.tsx
+++ b/packages/core/components/AnnotationFilterForm/index.tsx
@@ -33,7 +33,7 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
     const dispatch = useDispatch();
     const allFilters = useSelector(selection.selectors.getFileFilters);
     const fuzzyFilters = useSelector(selection.selectors.getFuzzyFilters);
-    const isQueryingSourceWithFuzzy = useSelector(selection.selectors.isQueryingAicsFms);
+    const canFuzzySearch = useSelector(selection.selectors.isQueryingAicsFms);
     const annotationService = useSelector(interaction.selectors.getAnnotationService);
     const [annotationValues, isLoading, errorMessage] = useAnnotationValues(
         props.annotation.name,
@@ -197,7 +197,7 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
                             fuzzySearchEnabled={fuzzySearchEnabled}
                             fieldName={props.annotation.displayName}
                             defaultValue={filtersForAnnotation?.[0]}
-                            hideFuzzyToggle={!isQueryingSourceWithFuzzy}
+                            hideFuzzyToggle={!canFuzzySearch}
                         />
                     );
                 }
--- a/packages/core/entity/FileSet/index.ts
+++ b/packages/core/entity/FileSet/index.ts
@@ -198,15 +198,10 @@ export default class FileSet {
         // Transform the map above into SQL comparison clauses
         const sqlBuilder = this.sort ? this.sort.toQuerySQLBuilder() : new SQLBuilder();
 
-        Object.entries(filtersGroupedByAnnotation).forEach(([annotation, appliedFilters]) => {
-            // If an annotation has no filtered values applied, we assume it's a hierarchy or include filter
-            if (appliedFilters.length === 0) {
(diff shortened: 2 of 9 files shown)


packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -33,6 +33,7 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
     const dispatch = useDispatch();
     const allFilters = useSelector(selection.selectors.getFileFilters);
     const fuzzyFilters = useSelector(selection.selectors.getFuzzyFilters);
+    const isQueryingAicsFms = useSelector(selection.selectors.isQueryingAicsFms);
```
kmitcham, line 36: Nit: is it possible for some of the alternative datasources to support fuzzy, while others do not?  in which case the isQueryingAicsFms would be better named 'isQueryingSourceWithFuzzy'  or something similar but less awkward?

kmitcham (approved):

packages/core/components/AnnotationFilterForm/index.tsx
aswallace: Fair point! In the current setup, with all non-FES sources we do fuzzy matching by default on our end, but makes sense to prepare for the possibility

packages/core/entity/FileSet/index.ts
```diff
@@ -202,7 +202,12 @@ export default class FileSet {
         Object.entries(filterValuesByAnnotation).forEach(([annotation, filterValues]) => {
             // If a filter value is `null` then we need to modify the way we approach filtering
             // it in SQL
-            if (filterValues.length === 0) {
+            if (!!this.excludeFilters?.some((filter) => filter.name === annotation)) {
+                sqlBuilder.where(`"${annotation}" IS NULL`);
```
pgarrison, line 206: I think I'm missing some context for this change. If this PR is about adding the include/exclude filters, why does FileSet already have the `excludeFilters` and `includeFilters` variables? And why weren't they used in `toQuerySQLBuilder` before?

aswallace: The main logic for include, exclude, and fuzzy filters was added in #238 and remains the same. However, the emphasis for that ticket was to integrate with FES, which we query via URLs and not through our database service (which is how we query other data sources, e.g., CSVs). 
The `toQuerySQLBuilder` is ONLY used by our database service, not by FES, so to prevent #238 from getting more bloated than it already was, I left it to be a separate ticket.

packages/core/services/AnnotationService/DatabaseAnnotationService/test/DatabaseAnnotationService.test.ts
```diff
@@ -59,6 +59,16 @@ describe("DatabaseAnnotationService", () => {
...
         });
+
+        it("issues a request for annotation values for the first level of the annotation hierarchy with typed filters", async () => {
+            const annotationService = new DatabaseAnnotationService({
+                dataSourceNames: ["e"],
+                databaseService,
+            });
+            const filter = new FileFilter("bar", "barValue", FilterType.ANY);
+            const values = await annotationService.fetchRootHierarchyValues(["foo"], [filter]);
+            expect(values).to.deep.equal(["Cell Line0", "Is Split Scene1", "Whatever2"]);
```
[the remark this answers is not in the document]
aswallace: These tests were essentially copy-pastes of the existing tests in this suite. Will update to clarify values

packages/core/entity/FileSet/index.ts
pgarrison: Okay, thanks. So it looks like `toQueryString` is for FES and `toQuerySQLBuilder` is for database services? 

On line 176, `FileSet.toQueryString` delegates to `filter.toQueryString`: why not follow the same pattern for `toQuerySQLBuilder` and leave it up to the filter class to decide the correct SQL? If I am reading correctly, this would remove the need for the `excludeFilters` and `includeFilters` variables in `FileSet`

aswallace: fwiw I'm quite sure I had a justification for not delegating to the class which I no longer remember. I think it had to do with the way we're mapping the filters to a grouped list for sql that made it seem like it would be complicated... I'll try it though and see if it comes back to me

pgarrison: Whatever you end up doing, I think my feedback for this PR would be to just use a comment to explain the decision

aswallace: I tried to reach a middle ground on this in [11f2e05](https://github.com/AllenInstitute/biofile-finder/pull/262/commits/11f2e05e91f1da314a17794fdda1db8c8c1b23b8). Not 100% sure I fully addressed your concerns though, so lmk if you have more questions

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -33,6 +33,7 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
     const dispatch = useDispatch();
     const allFilters = useSelector(selection.selectors.getFileFilters);
     const fuzzyFilters = useSelector(selection.selectors.getFuzzyFilters);
+    const isQueryingSourceWithFuzzy = useSelector(selection.selectors.isQueryingAicsFms);
```
SeanLeRoy, line 36: This name feels like what `fuzzySearchEnabled` is trying to describe. I think this is meant to describe whether or not we _can_ query the data with `fuzzy` so IMO I think something like `canFuzzySearch` would be clearer

packages/core/entity/SQLBuilder/index.ts
```diff
@@ -91,7 +91,7 @@ export default class SQLBuilder {
 
     public toSQL(): string {
         if (!this.fromStatement) {
-            throw new Error("Unable to build SLQ without a FROM statement");
+            throw new Error("Unable to build SQL without a FROM statement");
```
SeanLeRoy, line 94: 🚀

packages/core/entity/FileSet/index.ts
```diff
@@ -187,28 +187,24 @@ export default class FileSet {
...
         // Transform the map above into SQL comparison clauses
         const sqlBuilder = this.sort ? this.sort.toQuerySQLBuilder() : new SQLBuilder();
 
-        Object.entries(filterValuesByAnnotation).forEach(([annotation, filterValues]) => {
-            // If a filter value is `null` then we need to modify the way we approach filtering
-            // it in SQL
-            if (filterValues.length === 0) {
+        Object.entries(filtersGroupedByAnnotation).forEach(([annotation, appliedFilters]) => {
+            // If an annotation has no filtered values applied, we assume it's a hierarchy or include filter
+            if (appliedFilters.length === 0) {
```
SeanLeRoy, line 203: Can this case happen anymore? If all of the values in the key:value entries of `filtersGroupedByAnnotation` are coming from `this.filters` then there has to be at least one filter per annotation I believe.

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -61,27 +61,30 @@ export default class DatabaseAnnotationService implements AnnotationService {
...
+            {} as { [name: string]: FileFilter[] }
         );
 
         hierarchy
             // Map before filter because index is important to map to the path
             .forEach((annotation, index) => {
                 if (!filtersByAnnotation[annotation]) {
-                    filtersByAnnotation[annotation] = [index < path.length ? path[index] : null];
+                    filtersByAnnotation[annotation] = [
+                        new FileFilter(annotation, index < path.length ? path[index] : null),
```
SeanLeRoy, line 74: Will this result in creating the appropriate `type` of `FileFilter`? Ex. the "else" parse of the ternary here returns a `null` which is essentially trying to create an `IncludeFilter`. If the appropriate `type` isn't present in the `FileFilter`, won't the `toQueryString` return the wrong string?

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -92,14 +95,13 @@ export default class DatabaseAnnotationService implements AnnotationService {
             .from(this.dataSourceNames);
 
         Object.keys(filtersByAnnotation).forEach((annotationToFilter) => {
-            const annotationValues = filtersByAnnotation[annotationToFilter];
-            if (annotationValues[0] === null) {
+            const appliedFilters = filtersByAnnotation[annotationToFilter];
+            // If an annotation has no filtered values applied, we assume it's a hierarchy or include filter
```
SeanLeRoy, line 99: Same comment as above, is this still a case since what was previously a sentinel `null` value is now a whole filter as an `IncludeFilter`

aswallace: This condition is currently still fulfilled because of line 74. However, I can change 74 to be an actual Include filter (currently it's a default type one) and do away with this.

SeanLeRoy: Oh I see yeah I think keeping it consistent so that the IncludeFilter is used whenever this type of thing happens would be ideal

PR #269, opened 2024-10-11
Feature/warning info and error messaging

aswallace:
Updates UX for the system status messages that are displayed to the user to match Lyndsay's [Figma designs.](https://www.figma.com/design/oNQvOAK11YINcbPvNmkgwO/Public-File-Explorer?node-id=3166-33131&node-type=frame&t=VRvCCjbrx5OzU9Ri-0)

Also changes the messaging when a file cannot be reloaded -- moves the warning out of the data upload modal and into a status message to conform to other status messages. Previously, this warning persisted in the modal, even when the file had been re-uploaded. Now the status message goes away once the file is replaced.

## Old styling examples
image
Screenshot 2024-10-11 at 2 27 46 PM
image

## New styling examples
image

## Comparison of data source reload errors: before & after
image
image


packages/core/components/DataSourcePrompt/index.tsx
```diff
@@ -43,12 +44,12 @@ export default function DataSourcePrompt(props: Props) {
     };
 
     const onSubmit = (dataSource: Source, metadataSource?: Source) => {
-        if (sourceToReplace || query) {
+        if (sourceToReplace || query || requiresDataSourceReload) {
```
aswallace, line 47: @SeanLeRoy : I wasn't sure if we still needed the "sourceToReplace" logic here

packages/core/components/StatusMessage/index.tsx
```diff
@@ -45,31 +47,50 @@ export default function StatusMessage() {
...
                             })}
                             actions={cancelButton}
                             key={statusUpdate.processId}
                             messageBarType={statusToTypeMap[status]}
-                            onDismiss={() =>
-                                dispatch(interaction.actions.removeStatus(statusUpdate.processId))
-                            }
+                            styles={{
+                                iconContainer:
+                                    status === ProcessStatus.STARTED
```
SeanLeRoy, line 87: Not sure if this is the line that causes this, but I think users should be able to dismiss messages about files that are downloading. If a user tries to download 20 images they might get blocked by all the messages.

packages/core/state/interaction/actions.ts
```diff
@@ -439,6 +441,29 @@ export function processFailure(processId: string, msg: string): ProcessFailureAc
...
+ * PROCESS ERROR
+ *
+ * Intention to inform the user of a warning or error generated by the system
+ */
+export interface ProcessErrorAction {
+    type: string;
+    payload: StatusUpdate;
+}
+
+export function processError(processId: string, msg: string): ProcessErrorAction {
```
SeanLeRoy, line 454: Same comment about Error vs Failed

SeanLeRoy (approved):
Awesome Anya! I had some minor notes but overall LGTM! Are there any other `alert` or `console.error` or failure type things around we could switch to using this? If you could do a quick check to see just to make sure we are always presenting this awesome UI that would be great. If it ends up being a lot or tedious creating an issue in the backlog works too.

packages/core/components/QuerySidebar/QueryFooter.tsx
```diff
@@ -29,7 +29,7 @@ export default function QueryFooter(props: Props) {
 
     const onCopy = async () => {
         try {
-            navigator.clipboard.writeText(`https://biofile-finder.allencell.org/app?${url}`);
+            await navigator.clipboard.writeText(`https://biofile-finder.allencell.org/app?${url}`);
```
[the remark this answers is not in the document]
aswallace: strangely yes, I was hitting a weird bug where without the await the navigator would claim it had copied the text (send the 'success!' alert) but then nothing was in my clipboard

packages/core/components/StatusMessage/index.tsx
```diff
@@ -23,6 +23,8 @@ const statusToTypeMap = {
     [ProcessStatus.SUCCEEDED]: MessageBarType.success,
     [ProcessStatus.FAILED]: MessageBarType.error,
     [ProcessStatus.NOT_SET]: MessageBarType.info,
+    [ProcessStatus.WARNING]: MessageBarType.severeWarning,
+    [ProcessStatus.ERROR]: MessageBarType.severeWarning,
```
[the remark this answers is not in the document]
aswallace: good question, I went back and forth on this.
Since this component used to only be used for processes in progress, "Failed" was the existing class (w old icon) used for when that process failed to complete. I added error with the new icon from figma in case we wanted to use this modal for general system errors. Could definitely be combined, I'd just need to check with Lyndsay which of the icons we prefer to keep.

Here's failed, error and warning for comparison:
image

aswallace:
> Are there any other alert or console.error or failure type things around we could switch to using this? If you could do a quick check to see just to make sure we are always presenting this awesome UI that would be great. If it ends up being a lot or tedious creating an issue in the backlog works too

I can definitely add the alerts to this one. For the other console errors etc I suspect there might be ones where I'd need some guidance about whether/what to show the user (w UX?) so I might make a separate ticket for that

packages/core/components/QuerySidebar/QueryFooter.tsx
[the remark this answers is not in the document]
SeanLeRoy: Weird, but awesome catch!

packages/core/components/StatusMessage/index.tsx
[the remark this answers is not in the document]
SeanLeRoy: I don't particularly catch which we keep, but I think we should choose betwee ERROR or FAILURE

packages/core/components/StatusMessage/index.tsx
aswallace: Good point. One issue is that the download process currently resends the status message as progress continues, so dismissing once isn't permanent while it's ongoing. Maybe we could have a way of minimizing it/making the messages more compact instead of dismissing fully?

packages/core/components/StatusMessage/index.tsx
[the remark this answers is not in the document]
aswallace: I'm going with error then since it's more generic sounding

packages/core/components/StatusMessage/index.tsx
SeanLeRoy: Huh... that is annoying. I guess if it isn't a new blocker we are creating and it is just an existing potential blocker we aren't fixing then ignoring it for this PR is fine.

packages/core/components/StatusMessage/index.tsx
```diff
@@ -21,8 +21,9 @@ const statusToTypeMap = {
     [ProcessStatus.STARTED]: MessageBarType.info,
     [ProcessStatus.PROGRESS]: MessageBarType.info,
     [ProcessStatus.SUCCEEDED]: MessageBarType.success,
-    [ProcessStatus.FAILED]: MessageBarType.error,
```
kmitcham, line 24: [humor] Failure is not an option

kmitcham (approved):

BrianWhitneyAI:
Hey I looked through the different types of zarr download and it all looks pretty good! The only issues were 1) we dont support download tracking for local zarrs on desktop. Which is not relevant to this PR I think. and 2) that the error message for local zarr paths or unsupported cloud paths on web is not on this format yet [HERE ](https://github.com/AllenInstitute/biofile-finder/blob/5098b6837303ad4a7d6636506e282f6d2440a869/packages/web/src/services/FileDownloadServiceWeb.ts#L63) and [HERE](https://github.com/AllenInstitute/biofile-finder/blob/5098b6837303ad4a7d6636506e282f6d2440a869/packages/web/src/services/FileDownloadServiceWeb.ts#L52). (with photo at bottom)

Desktop:

- Local Zarr Download  ( Download starts but no error message / completion message ) Not supported
- Remote Public Download (Good) - Download Tracked
- Remote Private Download (Good) - network error 
- Canceled download ( good )  - Canceled By User

Web

- Local Path Zarr (Not Good)
- Remote not https aws path ( Good) 
- Remote Path (Good) - Download Tracked 
- Cancel download (Good)   - Canceled By User


Only missing one I saw: 
Screenshot 2024-10-15 at 1 25 33 PM

BrianWhitneyAI (changes requested):
Hey I looked through the different types of zarr download and it all looks pretty good! The only issues were 1) we dont support download tracking for local zarrs on desktop. Which is not relevant to this PR I think. and 2) that the error message for local zarr paths or unsupported cloud paths on web is not on this format yet [HERE ](https://github.com/AllenInstitute/biofile-finder/blob/5098b6837303ad4a7d6636506e282f6d2440a869/packages/web/src/services/FileDownloadServiceWeb.ts#L63) and [HERE](https://github.com/AllenInstitute/biofile-finder/blob/5098b6837303ad4a7d6636506e282f6d2440a869/packages/web/src/services/FileDownloadServiceWeb.ts#L52). (with photo at bottom)

Desktop:

- Local Zarr Download  ( Download starts but no error message / completion message ) Not supported
- Remote Public Download (Good) - Download Tracked
- Remote Private Download (Good) - network error 
- Canceled download ( good )  - Canceled By User

Web

- Local Path Zarr (Not Good)
- Remote not https aws path ( Good) 
- Remote Path (Good) - Download Tracked 
- Cancel download (Good)   - Canceled By User


Only missing one I saw: 
Screenshot 2024-10-15 at 1 25 33 PM

aswallace:
> that the error message for local zarr paths or unsupported cloud paths on web is not on this format yet

[This one](https://github.com/AllenInstitute/biofile-finder/blob/5098b6837303ad4a7d6636506e282f6d2440a869/packages/web/src/services/FileDownloadServiceWeb.ts#L52) is already fixed [here](https://github.com/AllenInstitute/biofile-finder/pull/269/files#diff-42033498f3c7b2c91f6ee60b3e73ec03e78f563adda620e53cfc0401c315aca6L54); the `throw` automatically converts to a status message.

For the other one you linked (`handleLocalZarrFile`), the most consistent solution would be to also throw an error the way `handleZarrFile` does instead of its current return. I've been hesitant to add state calls in there since it seemed like we tend to avoid that in services, but I'm not sure if that's an actual standard or that we just haven't needed to do it yet.

BrianWhitneyAI (approved):
LGTM!

PR #288, opened 2024-10-17
bugfix/thumbnail-bounding-box

BrianWhitneyAI:
quick fix for loading bounding box.


SeanLeRoy (commented):
Did this solve the problem of it the brain images being too small sometimes too?

BrianWhitneyAI:
> Did this solve the problem of it the brain images being too small sometimes too?

The height seems correct but the width is still proportional to the width of the fileinfo sidebar

SeanLeRoy (approved):

aswallace (approved):

PR #297, opened 2024-10-23
Update view info button style

### LINKED ISSUE #284: Update view info button style
https://www.figma.com/design/oNQvOAK11YINcbPvNmkgwO/Public-File-Explorer?node-id=1068-6905&node-type=frame

SeanLeRoy:
## Description
AGAVE is a desktop application which means it requires a download so we want to make it easy for people to find and download it. This changeset upgrades the styling of the button that links to the download page for AGAVE (and in the future for other desktop apps). Resolves #284

### Before
Screen Shot 2024-10-23 at 12 34 28 PM

### After
Screen Shot 2024-10-23 at 12 34 06 PM

### DIFF
--- a/packages/core/hooks/useOpenWithMenuItems/index.tsx
+++ b/packages/core/hooks/useOpenWithMenuItems/index.tsx
@@ -56,7 +56,7 @@ const DESKTOP_APPS = (fileDetails?: FileDetail): { agave: IContextualMenuItem } 
         key: "agave",
         className: styles.desktopMenuItem,
         text: "AGAVE",
-        title: `Open files with AGAVE v1.7.2+`,
+        title: "Open files with AGAVE v1.7.2+",
         href: `agave://?url=${fileDetails?.path}`,
         disabled: !fileDetails?.path,
         target: "_blank",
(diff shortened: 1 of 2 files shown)


packages/core/hooks/useOpenWithMenuItems/index.tsx
```diff
@@ -54,9 +54,9 @@ const WEB_APPS = (fileDetails?: FileDetail): WebApps => ({
 const DESKTOP_APPS = (fileDetails?: FileDetail): { agave: IContextualMenuItem } => ({
     agave: {
         key: "agave",
-        className: styles.agaveLink,
+        className: styles.desktopMenuItem,
         text: "AGAVE",
-        title: `Open files with AGAVE`,
+        title: "Open files with AGAVE v1.7.2+",
```
aswallace, line 59: Do we have any concern with keeping this version number synced?

aswallace (approved):

packages/core/hooks/useOpenWithMenuItems/index.tsx
SeanLeRoy: Nah because unless they remove support for opening from a browser (pretty unlikely) the user just needs a version >= when they added it when was 1.7.2

kmitcham (approved):

Issue #287, opened 2024-10-16
Minor style fixes


aswallace:
List of style fixes from [this document](https://docs.google.com/document/d/1QP_xT7IZOuLjgaHyK2gletNUK1ybCt7B_ayOG34H0zU) that haven't been completed yet.
Most of these are small & should be quick fixes. Details and screenshots are in doc

- [x] Help menu dropdown items are 14px and should be 16px. Also, there is CSS causing little white corners to show.
- [x] Use a different icon for “View Datasets” button on home page
- [ ] ~Discover features items on home: Either add autoplay or left/right buttons with (e.g.) 1/4~ (Skipping since overlaps with #207 )
- [x] Open-source datasets: related publication link should open in new tab
- [x] "Active" style for query naming input should be bright aqua (match search box outline styling)
- [x] Check line spacing for text & subheaders in new data source modal
- [x] Swap caret direction for collapsed query states
- [x] Check fill/size for icons in queries
- [x] Left panel is not using the intended icon for “collapse”
- [x] Add box shadow to individual queries
- [x] Check spacing on left and right panels (see doc/figma for details)
- [x] All buttons (e.g., download/open) should have left/right padding of 19px at all times
- [x] Check if text in last file in list is still cut off
- [ ] ~Agave menu item styling~ (This is covered by #284 )
- [x] Remove icon from file types (csv, json, parquet)
- [x] Change "New query" in collapsed left panel to "Open query panel" or similar, fix icon, change background & font weight
- [x] Various thumbnail svg issues
    - File names on image grid are not white. 
    -  Not rendering border radius
    - Svg looks off-center
    - Svg color conflicts with background when selected
    - Label needs more padding
- [x] Row styling does not extend all the way across

SuyashShukla0007:
can i work on this

PR #302, opened 2024-10-23
Fix stalled thumbnail unit tests

aswallace:
The unit tests for `LazilyRenderedThumbnail` were getting stalled. It seems the `.useEffect` was creating an infinite loop, which is apparently a common issue with React/jest. 

I'm still not sure if I fully understand, but my best interpretation is that `useEffect` is dependent on `file` changes, & calling `setThumbnailPath` technically modifies `file`, so the effect runs again. Then toggling `isLoading` back to true triggers a re-render since the `FileThumbnail` component depends on `isLoading`, which calls `useEffect` again, and the loop starts all over... So our unit tests never fully finished rendering the component. 

## Proposed change
Since we already initialize `isLoading` to true and don't set it to false anywhere else, we can remove setting it to true in the effect. This way, `useEffect` only calls twice (once on initialization, and then once to register the change to `file`) and then stops. 

There's also an unrelated .css change; I missed increasing the font size for thumbnail labels in #287.

## Testing/Reviewing
I verified that unit tests run & finish now, but it would be great to double check that the thumbnail loading state still works as expected.

### DIFF
--- a/packages/core/components/FileList/LazilyRenderedThumbnail.module.css
+++ b/packages/core/components/FileList/LazilyRenderedThumbnail.module.css
@@ -16,7 +16,7 @@
 }
 
 .small-font {
-  font-size: var(--xs-paragraph-size);
+  font-size: var(--s-paragraph-size);
 }
 
 .focused {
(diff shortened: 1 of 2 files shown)


BrianWhitneyAI (approved):
LGTM!

kmitcham (approved):

PR #303, opened 2024-10-24
Eternal load on start

### LINKED ISSUE #285: App landing page data source is set to eternally load

SeanLeRoy:
The spinner was blocking the data source query panel on start for users that both aren't looking at a specific data source yet and aren't on institute wifi. This adds a qualifier to make it only load when we are expecting there to be a data source and isn't one.

Resolves #285


aswallace:
The one other time this seems to happen is if you only have one query open and you delete it

BrianWhitneyAI (approved):

kmitcham (approved):

PR #306, opened 2024-10-24
LOAD button on dataset page bug

### LINKED ISSUE #300: Load button on datasets not loading
Load button opens details panel instead of loading in BFF.

SeanLeRoy:
The "LOAD" button on the open-source datasets page doesn't actually load the dataset. This is because the pointer-events property here is preventing that. Resolves #300

### DIFF
--- a/packages/web/src/components/OpenSourceDatasets/DatasetRow.module.css
+++ b/packages/web/src/components/OpenSourceDatasets/DatasetRow.module.css
@@ -20,7 +20,6 @@
 }
 
 .button-wrapper {
-  pointer-events: none;
   position: absolute;
   top: 0;
   left: 0;
--- a/packages/web/src/components/OpenSourceDatasets/DatasetRow.module.css
+++ b/packages/web/src/components/OpenSourceDatasets/DatasetRow.module.css
@@ -22,12 +22,15 @@
 .button-wrapper {
   position: absolute;
   top: 0;
-  left: 0;
-  width: 100%;
+  right: 0;
   height: 100%;
   display: flex;
   align-items: center;
   justify-content: flex-end;
+}
+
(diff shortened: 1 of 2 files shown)


BrianWhitneyAI (approved):

aswallace (approved):
The only minor tradeoff from this is that the "related publication" link won't be clickable in the dataset row. If we keep it that way, Lyndsay suggested that we should remove the blue link styling from that column (which we agreed should be fine since the link is still clickable from the side panel)

SeanLeRoy:
> The only minor tradeoff from this is that the links in the "related publication" column won't be clickable in the dataset row. If we keep it that way, Lyndsay suggested that we should remove the blue link styling from that column so it doesn't look like it should be clickable (which we agreed should be fine since the link is still clickable from the side panel)

I think it still is? I'm confused about this, can you send me a video or screen share tomorrow?

aswallace:
Staging/prod: link has hoverable state and opens pub in new tab

https://github.com/user-attachments/assets/c7778639-ca49-4c95-9352-bf15320fb281


without pointer events: link is blue but clicking just opens the panel

https://github.com/user-attachments/assets/962e05fe-b95a-4b27-a6ae-be64aa71c859

SeanLeRoy:
> The only minor tradeoff from this is that the links in the "related publication" column won't be clickable in the dataset row. If we keep it that way, Lyndsay suggested that we should remove the blue link styling from that column so it doesn't look like it should be clickable (which we agreed should be fine since the link is still clickable from the side panel)

Fixed in https://github.com/AllenInstitute/biofile-finder/pull/306/commits/e66d5d758df75ab0571a9fcbc448743bb0f0ca0e

PR #313, opened 2024-10-31
Upgrade data source validation

### LINKED ISSUE #272: CSV file format
I was trying to use the tool with CSVs from recent brain science released genetic tools atlas, but it never seems to show me anything after i upload the CSV and i'm not sure why... all the images are in public locations in ZARR format in S3. 

Here are the example CSVs i used, from most comprehensive to more limited examples and columns. 

[Updated_Specimen_Metadata_one.csv](https://github.com/user-attachments/files/17351110/Updated_Specimen_Metadata_one.csv)
[Updated_Specimen_Metadata_small.csv](https://github.com/user-attachments/files/17351111/Updated_Specimen_Metadata_small.csv)
[Updated_Specimen_Metadata_with_ome-zarr_locations.csv](https://github.com/user-attachments/files/17351112/Updated_Specimen_Metadata_with_ome-zarr_locations.csv)

SeanLeRoy:
## Description
We got some reports from users that they were having trouble troubleshooting getting their data into BFF. Specifically that it wasn't clear what was happening when "File Path" exact casing wasn't present, when there were duplicate values, and when there were missing values. This changeset makes it so that when a data source is added a validation step runs from which errors will be reported in the UI if relevant.

## Error Examples

Duplicate & blank values (will also be shown separately when relevant)
Screen Shot 2024-10-30 at 5 55 02 PM

Unable to access data source (like on refresh)
Screen Shot 2024-10-30 at 5 49 50 PM

No "File Path" column detected
Screen Shot 2024-10-30 at 5 28 51 PM

## Related Issue
Resolves #272

### DIFF
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -31,9 +31,9 @@ export default abstract class DatabaseService {
         _format: "csv" | "parquet" | "json"
     ): Promise<Uint8Array>;
 
-    public abstract query(_sql: string): Promise<{ [key: string]: string }[]>;
+    public abstract query(_sql: string): Promise<{ [key: string]: any }[]>;
 
-    protected abstract addDataSource(_dataSource: Source): Promise<void>;
+    protected abstract addDataSource(_dataSource: Source, _skipValidation?: boolean): Promise<void>;
 
     protected abstract execute(_sql: string): Promise<void>;
 
@@ -59,8 +59,10 @@ export default abstract class DatabaseService {
         this.query = this.query.bind(this);
     }
 
-    public async prepareDataSources(dataSources: Source[]): Promise<void> {
-        await Promise.all(dataSources.map(this.addDataSource));
+    public async prepareDataSources(dataSources: Source[], skipValidation = false): Promise<void> {
+        await Promise.all(
+            dataSources.map((dataSource) => this.addDataSource(dataSource, skipValidation))
+        );
 
         // Because when querying multiple data sources column differences can complicate the queries
         // preparing a table ahead of time that is the aggregate of the data sources is most optimal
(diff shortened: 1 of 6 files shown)


packages/core/services/DatabaseService/index.ts
```diff
@@ -92,6 +94,70 @@ export default abstract class DatabaseService {
...
+
+        if (!columnsOnTable.has("File Path")) {
+            let error =
+                '\
+                "File Path" column is missing in the data source. \
+                Check the data source header row for a "File Path" (case-sensitive) column name and try again.';
+
+            // Attempt to find a column with a similar name to "File Path"
+            const columns = Array.from(columnsOnTable);
+            const filePathLikeColumn =
```
kmitcham, line 116: What happens here if there are multiple hits?  eg, a column named 'pathfinder' and a column named 'file'?

packages/core/services/DatabaseService/index.ts
```diff
@@ -92,6 +94,70 @@ export default abstract class DatabaseService {
...
+
+        const errors: string[] = [];
+        const isFilePathEverBlankQueryResult = await this.query(`\
+            SELECT EXISTS (                      \
+                FROM "${name}"                   \
+                WHERE TRIM("File Path") IS NULL  \
+            ) AS result                          \
+        `);
+        if (isFilePathEverBlankQueryResult[0]["result"]) {
+            errors.push('"File Path" column contains empty or purely whitespace values.');
```
kmitcham, line 137: Could we upgrade this to count the errors, or ideally list a value from the problem rows?  If my source has a million rows, I don't want to find the errors by hand.  Does the client have access to the SQL?  I think the sql is a function of the app using a library to be smart about csv.

packages/core/services/DatabaseService/index.ts
```diff
@@ -92,6 +94,70 @@ export default abstract class DatabaseService {
...
+            WHERE TRIM("File Path") IS NOT NULL  \
+        `);
+        const {
+            all: totalFilePathCount,
+            uniques: uniqueFilePathCount,
+        } = uniqueFilePathQueryResult[0];
+        const duplicateFilePathCount = totalFilePathCount - uniqueFilePathCount;
+        if (duplicateFilePathCount > 0) {
+            errors.push(
+                `"File Path" column contains duplicates. Found ${duplicateFilePathCount} duplicate values.`
```
kmitcham, line 154: Same thing here; the more the error message can point at the problem rows, the better.

kmitcham (approved):

packages/core/services/DatabaseService/index.ts
SeanLeRoy: It'll point to `pathfinder` in that case. This is just to get a "best" guess at what they were trying to name "File Path"

packages/core/services/DatabaseService/index.ts
SeanLeRoy: Yeah we could probably list the rows, good idea

PR #314, opened 2024-11-01
feature/web-download-threshold

### LINKED ISSUE #283: Add threshold for web download (2 GB) warning message
In web browsers we have to read the entire zarr image into memory and cannot stream files as zarrs are a full directory. As a result of this we are bound to the browsers inherent memory limit of about 2GB. We should notify users if they are trying to download a file larger than this threshold.

BrianWhitneyAI:
The purpose of this PR is to resolve #283. In web browsers we have to read the entire zarr image into memory and cannot stream files as zarrs are a full directory. As a result of this we are bound to the browsers inherent memory limit of about 2GB. We should notify users if they are trying to download a file larger than this threshold.

### DIFF
--- a/packages/core/components/FileDetails/index.tsx
+++ b/packages/core/components/FileDetails/index.tsx
@@ -14,7 +14,7 @@ import FileThumbnail from "../../components/FileThumbnail";
 import { interaction } from "../../state";
 
 import styles from "./FileDetails.module.css";
-import { MaximumDownloadSizeBrowser } from "../../services/FileDownloadService";
+import { MAX_DOWNLOAD_SIZE_WEB } from "../../services/FileDownloadService";
 
 interface Props {
     className?: string;
@@ -120,7 +120,7 @@ export default function FileDetails(props: Props) {
               isZarr &&
               // The Zarr size is calculated using the same traversal method as downloads
               // meaning that if the size cannot be determined, the download is also not possible.
-              (calculatedSize === null || calculatedSize > MaximumDownloadSizeBrowser))
+              (calculatedSize === null || calculatedSize > MAX_DOWNLOAD_SIZE_WEB))
         : true;
 
     // Prevent triggering multiple downloads accidentally -- throttle with a 1s wait
--- a/packages/core/components/FileDetails/index.tsx
+++ b/packages/core/components/FileDetails/index.tsx
@@ -118,6 +118,8 @@ export default function FileDetails(props: Props) {
         ? processStatuses.some((status) => status.data.fileId?.includes(fileDetails.id)) ||
           (isOnWeb &&
               isZarr &&
+              // The Zarr size is calculated using the same traversal method as downloads
+              // meaning that if the size cannot be determined, the download is also not possible.
               (calculatedSize === null || calculatedSize > MaximumDownloadSizeBrowser))
         : true;
 
--- a/packages/core/components/FileDetails/index.tsx
+++ b/packages/core/components/FileDetails/index.tsx
@@ -14,6 +14,7 @@ import FileThumbnail from "../../components/FileThumbnail";
 import { interaction } from "../../state";
 
 import styles from "./FileDetails.module.css";
+import { MaximumDownloadSizeBrowser } from "../../services/FileDownloadService";
 
 interface Props {
     className?: string;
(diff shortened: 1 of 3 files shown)


BrianWhitneyAI:
Screenshot 2024-11-01 at 4 08 03 PM

SeanLeRoy (commented):
I think this should ideally happen before they even try to download the file. Do you think we could do this fast enough to have to focused (selected for the detail pane) file such that we could then toggle the "Download" option to be disabled?

SeanLeRoy:
Also, we should use the "File Size" metadata field when we can (available on the file object as the "size" property) and only calculate the file size when that isn't provided

BrianWhitneyAI:
I disabled the download button on web for zarrs larger than 2 GB and had the size default to fileDetail.Size. Notably we internally set size to 0 if size is not present.

BrianWhitneyAI:
I left intact the download error message since users could batch select files and download them using the download option, bypassing the file detail download button.

packages/core/components/FileDetails/index.tsx
```diff
@@ -75,20 +75,51 @@ export default function FileDetails(props: Props) {
...
 
     const processStatuses = useSelector(interaction.selectors.getProcessStatuses);
     const openWithMenuItems = useOpenWithMenuItems(fileDetails || undefined);
 
+    // Disable download of large Zarrs ( > 2GB).
+    const isDownloadDisabled = fileDetails
+        ? processStatuses.some((status) => status.data.fileId?.includes(fileDetails.id)) ||
+          (isOnWeb &&
+              isZarr &&
+              (calculatedSize === null || calculatedSize > 2 * 1024 * 1024 * 1024))
```
SeanLeRoy, line 120: How long does it take to calculate various zarrs? Like how long does it take for a 2GB zarr vs a 1TB zarr?

SeanLeRoy:
> download

Can we disable that ahead of time as well?

SeanLeRoy:
> I left intact the download error message since users could batch select files and download them using the download option, bypassing the file detail download button.

Can we disable that ahead of time as well?

packages/core/components/FileDetails/index.tsx
BrianWhitneyAI: The largest I tested with was a 40Gb file and it was pretty instantaneous.

BrianWhitneyAI:
> > I left intact the download error message since users could batch select files and download them using the download option, bypassing the file detail download button.
> 
> Can we disable that ahead of time as well?

What would that look like? just if any large zarr is selected it cannot be downloaded? I think this might be confusing to users who select a bunch of files but are blocked by a single zarr file.

packages/core/components/FileDetails/index.tsx
aswallace: Is this value likely to change/would it make sense to have a MAXIMUM_DOWNLOAD_SIZE const somewhere instead calculating the 2 * 1024 * 1024 * 1024 since we also use it in FileDownloadServiceWeb ?

aswallace (approved):

SeanLeRoy:
> > > I left intact the download error message since users could batch select files and download them using the download option, bypassing the file detail download button.
> > 
> > 
> > Can we disable that ahead of time as well?
> 
> What would that look like? just if any large zarr is selected it cannot be downloaded? I think this might be confusing to users who select a bunch of files but are blocked by a single zarr file.

I think just the focused file is downloaded the moment so you'd just need to calculate that one which you'd already be doing / have done for the filedetails pane

SeanLeRoy (approved):

aswallace (approved):

packages/core/components/FileDetails/index.tsx
```diff
@@ -75,20 +76,53 @@ export default function FileDetails(props: Props) {
...
     const [isThumbnailLoading, setIsThumbnailLoading] = React.useState(true);
     const stackTokens: IStackTokens = { childrenGap: 12 + " " + 20 };
+    const [calculatedSize, setCalculatedSize] = React.useState<number | null>(null);
+
+    const platformDependentServices = useSelector(
+        interaction.selectors.getPlatformDependentServices
+    );
+    const fileDownloadService = platformDependentServices.fileDownloadService;
+    const isOnWeb = useSelector(interaction.selectors.isOnWeb);
+    const isZarr = fileDetails?.path.endsWith(".zarr") || fileDetails?.path.endsWith(".zarr/");
```
SeanLeRoy, line 86: Oh I hadn't realized ".zarr/" would be a possibility. I bet I have to refactor some regex to account for this

PR #321, opened 2024-11-08
Upgrade Data Source Ingestion

### LINKED ISSUE #272: CSV file format

SeanLeRoy:
## Background
AIBS and others have had trouble inputting their data sources since the feedback wasn't being shown to the user most of the time and even when it was it was hard to interpret as actionable. Also, several have noticed that empty file paths would lead to missing rows.

Also, @kmitcham made me realize that the app has enough data to be smarter about the kind of feedback it gives and when it can provide defaults.

## Description
This does 3 things as part of ingesting a data source:
* Normalizes the column names. Makes columns like "file_name" change into "File Name" so we no longer have issues with case-sensitivity
* Checks for uniqueness constraint issues and/or blank values in columns where we need a value. This includes making sure either File Path or File ID are unique.
* Adds default columns for File Name and File ID when not present already. This helps by guaranteeing that "File ID" is always present and is always unique since some users want "File Path" to not always be unique.
    * TODO: Triple check with users that default "File ID" isn't annoying. Unsure about alternatives...

## Testing
I've tested this manually with several different .csvs with various casings, missing values, duplicate values and such. However, I would really appreciate as part of this review that you try this out too!  [It is on staging](https://staging.biofile-finder.allencell.org/app)

### Related Issue
Resolves #272

### DIFF
--- a/packages/core/components/DataSourcePrompt/index.tsx
+++ b/packages/core/components/DataSourcePrompt/index.tsx
@@ -17,10 +17,10 @@ interface Props {
 }
 
 const ADDITIONAL_COLUMN_DETAILS = [
-    'If a "Thumbnail" column (case-sensitive) is present it should contain a web URL to a thumbnail image for the file. ',
-    'If a "File Name" column (case-sensitive) is present it should be the file\'s name (this will replace the "File Name" created by default from the path). ',
-    'If a "File Size" column (case-sensitive) is present it should contain the size of the file in bytes. ',
-    'If an "Uploaded" column (case-sensitive) is present it should contain the date the file was uploaded to the cloud storage and be formatted as YYYY-MM-DD HH:MM:SS.Z where Z is a timezone offset. ',
+    'If a "Thumbnail" column is present it should contain a web URL to a thumbnail image for the file. ',
+    'If a "File Name" column is present it should be the file\'s name (this will replace the "File Name" created by default from the path). ',
+    'If a "File Size" column is present it should contain the size of the file in bytes. ',
+    'If an "Uploaded" column is present it should contain the date the file was uploaded to the cloud storage and be formatted as YYYY-MM-DD HH:MM:SS.Z where Z is a timezone offset. ',
 ];
 
 /**
@@ -81,13 +81,10 @@ export default function DataSourcePrompt(props: Props) {
                 <>
                     <ul className={styles.detailsList}>
                         <li className={styles.details}>
-                            The file must contain a &quot;File Path&quot; column (case-sensitive) &
-                            must be unique by the &quot;File Path&quot; column. Any other columns
-                            are optional and will be available as annotations to query by.
-                        </li>
-                        <li className={styles.details}>
-                            The &quot;File Path&quot; column should contain full path to a file in a
-                            cloud storage.
+                            The file must contain a &quot;File Path&quot; column & must be unique by
+                            the &quot;File Path&quot; column or have a unique &quot;File ID&quot;
+                            column. Any other columns are optional and will be available as
+                            annotations to query by.
                         </li>
                     </ul>
                     <h4 className={styles.details}>Advanced:</h4>
@@ -98,7 +95,7 @@ export default function DataSourcePrompt(props: Props) {
                             as&quot; options.
                         </li>
                         <li className={styles.details}>
-                            These are additional special columns that are optional, but will be
+                            These are additional pre-defined columns that are optional, but will be
                             handled as a special case in the application:
                         </li>
                         <ul className={styles.detailsList}>
--- a/packages/core/state/selection/logics.ts
+++ b/packages/core/state/selection/logics.ts
@@ -45,6 +45,7 @@ import {
     ADD_DATASOURCE_RELOAD_ERROR,
     REMOVE_DATASOURCE_RELOAD_ERROR,
     CHANGE_FILE_FILTER_TYPE,
+    AddDataSourceReloadError,
 } from "./actions";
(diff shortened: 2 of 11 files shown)


packages/web/src/services/DatabaseServiceWeb.ts
```diff
@@ -78,55 +75,39 @@ export default class DatabaseServiceWeb extends DatabaseService {
...
-            } else {
-                // Default to CSV
-                await this.execute(
-                    `CREATE TABLE "${name}" AS FROM read_csv_auto('${name}', header=true, all_varchar=true);`
-                );
-            }
-        } catch (err) {
-            await this.deleteDataSource(name);
-            throw new DataSourcePreparationError((err as Error).message, name);
+        if (type === "parquet") {
```
BrianWhitneyAI, line 102: Could converting the `name` variable directly to a SQL query be vurnrable to probaimatic characters here? I think ";" would break this if its in the name or something like "TABLE" alread in the name.

packages/web/src/services/DatabaseServiceWeb.ts
```diff
@@ -78,55 +75,39 @@ export default class DatabaseServiceWeb extends DatabaseService {
...
+            await this.execute(`CREATE TABLE "${name}" AS FROM parquet_scan('${name}');`);
```
BrianWhitneyAI, line 103: Could converting the name variable directly to a SQL query be vurnrable to probaimatic characters here? I think ";" would break this if its in the name or something like "TABLE" alread in the name.

packages/core/components/DataSourcePrompt/index.tsx
```diff
@@ -17,10 +17,10 @@ interface Props {
...
 
 const ADDITIONAL_COLUMN_DETAILS = [
-    'If a "Thumbnail" column (case-sensitive) is present it should contain a web URL to a thumbnail image for the file. ',
-    'If a "File Name" column (case-sensitive) is present it should be the file\'s name (this will replace the "File Name" created by default from the path). ',
-    'If a "File Size" column (case-sensitive) is present it should contain the size of the file in bytes. ',
-    'If an "Uploaded" column (case-sensitive) is present it should contain the date the file was uploaded to the cloud storage and be formatted as YYYY-MM-DD HH:MM:SS.Z where Z is a timezone offset. ',
+    'If a "Thumbnail" column is present it should contain a web URL to a thumbnail image for the file. ',
+    'If a "File Name" column is present it should be the file\'s name (this will replace the "File Name" created by default from the path). ',
+    'If a "File Size" column is present it should contain the size of the file in bytes. ',
+    'If an "Uploaded" column is present it should contain the date the file was uploaded to the cloud storage and be formatted as YYYY-MM-DD HH:MM:SS.Z where Z is a timezone offset. ',
```
BrianWhitneyAI, line 23: Maybe provide a date example here "(e.g., 2024-11-08 14:30:45.0)"

BrianWhitneyAI (approved):

packages/core/state/selection/logics.ts
```diff
@@ -709,23 +704,29 @@ const replaceDataSourceLogic = createLogic({
...
-        const { payload: dataSourceName, type } = deps.action;
-        const datasourceErrorDefaultMessage = `There was an error loading the data source file
-                        '${dataSourceName}'. Please re-select the data source file or a replacement.
-                    </br>
-                        If this is a local file, the browser&apos;s permissions to access the file
-                        may have expired since last time. If so, consider putting the file in a
-                        cloud storage and providing the URL to avoid this issue in the future.`;
-
-        if (type === ADD_DATASOURCE_RELOAD_ERROR) {
+        const isNewError = deps.action.type === ADD_DATASOURCE_RELOAD_ERROR;
```
aswallace, line 707: I'm running into an issue on staging where no matter what's wrong with the file it only shows the "Lost access to the data source" error. This screenshot is from when I uploaded a csv with no file path
image

SeanLeRoy: This is what you should be seeing and is what I am seeing :( I'll poke around
Screen Shot 2024-11-11 at 9 30 40 AM

packages/web/src/services/DatabaseServiceWeb.ts
SeanLeRoy: Yea this isn't a new feature though, but a fair concern I'll note. I wonder if any users have hit this issue yet

packages/core/state/selection/logics.ts
aswallace: I'm trying again this morning and it's working about half the time. I can't quite nail down what the condition are where it shows the correct error vs the one I have screen captured

SeanLeRoy: Fixed in https://github.com/AllenInstitute/biofile-finder/pull/321/commits/91fc03ad9d70ddf7dbad18723abca82c5b117de1

aswallace (approved):

PR #322, opened 2024-11-11
feature/copy-to-local-cache-request

### LINKED ISSUE #291: File Service moveFiles functionality
We want to allow users to move file on and off NAS cache.  In order to do this we need to add a new method `moveFiles` to our FileService that can handle these requests. This will need to be able to handle large batches of files

BrianWhitneyAI:
## Description 
The purpose of this PR is to add in the functionality for submitting the cache request and resolve #291. This PR does three main things:

1.  Repurposes the `FileDownloadServiceBaseURl` which was unused to `AicsLoadBalancerBaseURl`
2. Adds cacheFiles to the existing logic to build the request.
3. Adds a rawPut function to the existing HttpServiceBase for put functionality.

### Request Format

```
        var url = "http://stg-aics/fss2/v3.0/file/cache";
        var id = "03c3eaf8236b4fa4bb1bba14dad12a02";
        var params = {
            "fileIds": [id]
        };
        var xmlHttp = new XMLHttpRequest();
        xmlHttp.open("PUT", url);
        xmlHttp.setRequestHeader("X-User-Id", "daniel.saelid");
        xmlHttp.setRequestHeader("Content-Type", "application/json")
        xmlHttp.send(JSON.stringify(params));
```

### Testing 
This pr has been tested against staging, but has not been tested against production as it is blocked by CORS policy.


BrianWhitneyAI:
This also doesn't contain any notification content to the user about success/failure (apart from invalid submissions) and the response is unused.  We need to think about what we want to return to users.

BrianWhitneyAI:
> This may be for another PR but we need to make some decisions about attaching user ids to these requests. We have existing functionality to access the users credentials on desktop but not on web.

In [46bae5d](https://github.com/AllenInstitute/biofile-finder/pull/322/commits/46bae5d4f05ef4e1a824dc95ae5365776e24f9c5) we use the existing username to build the request. This only works for desktop so we disabled the option for web in [f61ded5](https://github.com/AllenInstitute/biofile-finder/pull/322/commits/f61ded5b550f1f42c6956edcc0b1f75aa47db266)

packages/core/constants/index.ts
```diff
@@ -11,6 +11,12 @@ export enum FileExplorerServiceBaseUrl {
     PRODUCTION = "https://production.int.allencell.org",
 }
 
+export enum AicsLoadBalancerBaseUrl {
+    LOCALHOST = "http://localhost:8080",
+    STAGING = "http://stg-aics.corp.alleninstitute.org",
+    PRODUCTION = "http://aics.corp.alleninstitute.org",
+}
```
aswallace, line 18: May or may not work for this case, but for the MMS endpoint, we're intending to use a mapping of the current FES url to an MMS url instead of having to intialize the app with an additional URL
https://github.com/AllenInstitute/biofile-finder/blob/8a0f1194a6f0b438c4c4838b1efa72c8dcac8cc0/packages/core/services/FileService/HttpFileService/index.ts#L36-L40

packages/core/state/interaction/logics.ts
```diff
@@ -582,12 +582,22 @@ const setIsSmallScreen = createLogic({
...
-        console.log(`Moving files:`, action.payload.fileDetails);
-        done();
+    async process({ action, getState }: ReduxLogicDeps, _dispatch, done) {
+        try {
+            const httpFileService = interactionSelectors.getHttpFileService(getState());
+            const username = interactionSelectors.getUserName(getState());
+            const fileIds = (action as MoveFilesAction).payload.fileDetails.map((file) => file.id);
+            const cacheStatuses = await httpFileService.cacheFiles(fileIds, username);
+
+            // TODO: What to do with the status
```
SeanLeRoy, line 592: Is this in another issue?

packages/core/state/interaction/logics.ts
```diff
@@ -582,12 +582,22 @@ const setIsSmallScreen = createLogic({
...
 const moveFilesLogic = createLogic({
-    type: MOVE_FILES,
-    process(deps, dispatch, done) {
-        const action = deps.action as MoveFilesAction;
-        console.log(`Moving files:`, action.payload.fileDetails);
-        done();
+    async process({ action, getState }: ReduxLogicDeps, _dispatch, done) {
+        try {
+            const httpFileService = interactionSelectors.getHttpFileService(getState());
+            const username = interactionSelectors.getUserName(getState());
```
SeanLeRoy, line 588: Does this come back as potentially undefined? What is the value on web? I know the UI prevents using this on web, but adding a check for the username here if it is potentially undefined would be good type guarding

packages/core/state/interaction/logics.ts
```diff
@@ -582,12 +582,22 @@ const setIsSmallScreen = createLogic({
...
+            const fileIds = (action as MoveFilesAction).payload.fileDetails.map((file) => file.id);
+            const cacheStatuses = await httpFileService.cacheFiles(fileIds, username);
+
+            // TODO: What to do with the status
+            console.log("Cache statuses:", cacheStatuses);
+        } catch (err) {
+            console.error(`Error encountered while moving files: ${err}`);
```
SeanLeRoy, line 595: This should use the alert system

packages/core/state/interaction/selectors.ts
```diff
@@ -174,17 +186,25 @@ export const getAnnotationService = createSelector(
...
+        getUserName,
+        getAicsLoadBalancerBaseUrl,
+        getFileExplorerServiceBaseUrl,
+        getRefreshKey,
+    ],
+    (applicationVersion, userName, aicsLoadBalancerBaseUrl, fileExplorerBaseUrl) =>
         new DatasetService({
             applicationVersion,
             userName,
+            aicsLoadBalancerBaseUrl: aicsLoadBalancerBaseUrl,
```
SeanLeRoy, line 207: Is this actually used by the dataset service? If so, how?

saeliddp:
> This also doesn't contain any notification content to the user about success/failure (apart from invalid submissions) and the response is unused. We need to think about what we want to return to users.

As of right now, there are 5 possible statuses for each of the files submitted in the request--I'll give suggestions for what to immediately show users based on each

- `DOWNLOAD_STARTED` - spinning progress wheel
- `DOWNLOAD_IN_PROGRESS` - spinning progress wheel
- `DOWNLOAD_COMPLETE` - green check mark indicating cached file
- `FILE_RECORD_NOT_FOUND` - red 'x' indicating failure
- `FILE_NOT_FOUND_ON_CLOUD` - red 'x' indicating failure

For files that haven't been recently sent to the cache endpoint, state can be inferred from the following field combinations:

- `shouldBeInLocal=true`, `localPath=null` - spinning progress wheel (caching in progress)
- `shouldBeInLocal=true`, `localPath=<something>` - green check mark (file is in cache)
- `shouldBeInLocal=false` - some UI element to show that the file is not cached
- `shouldBeInLocal=false` - something to show the file isn't cached

**Note**: the fields I just described above haven't yet been implemented in the ETL, I'm working on that now. The status response from the `cache` endpoint is good to go though.

packages/core/state/interaction/logics.ts
BrianWhitneyAI: it does potentially come back as undefined. In this situation during submission the user is set to 'anonymous'. This should not be reachable for web since it is disabled but in the strange condition that a desktop user is undefined than it will use the `anonymous` user id.

packages/core/constants/index.ts
BrianWhitneyAI: Resolved by https://github.com/AllenInstitute/biofile-finder/pull/339

packages/core/state/interaction/selectors.ts
BrianWhitneyAI: resolved in https://github.com/AllenInstitute/biofile-finder/pull/339

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -127,4 +129,29 @@ export default class HttpFileService extends HttpServiceBase implements FileServ
...
+        fileIds: string[],
+        username?: string
+    ): Promise<{ cacheFileStatuses: { [fileId: string]: string } }> {
+        const requestUrl = `${this.aicsLoadBalancerBaseUrl}/${HttpFileService.BASE_FILE_CACHE_URL}${this.pathSuffix}`;
+        const requestBody = JSON.stringify({ fileIds });
+        const headers = {
+            "Content-Type": "application/json",
+            "X-User-Id": username || "anonymous",
+        };
+        console.log(headers);
```
SeanLeRoy, line 146: console.log

SeanLeRoy (approved):

aswallace (approved):

packages/core/services/FileService/HttpFileService/index.ts
BrianWhitneyAI: resolved in [44b3f16](https://github.com/AllenInstitute/biofile-finder/pull/322/commits/44b3f168a061bf7397716f4d325e0b5a2fde6db0)

packages/core/state/interaction/logics.ts
BrianWhitneyAI: resolved in [44b3f16](https://github.com/AllenInstitute/biofile-finder/pull/322/commits/44b3f168a061bf7397716f4d325e0b5a2fde6db0)

packages/core/state/interaction/logics.ts
BrianWhitneyAI: https://github.com/AllenInstitute/biofile-finder/issues/346

PR #323, opened 2024-11-13
Feature/hide unique id for data sources

SeanLeRoy:
## Context
In the last changeset related to this I added validation at ingestion time of a data source that would check for errors and automatically add some columns. This was largely because users were having a difficult time telling what the issue with their data was in the event of a failure.

## Description
This changeset removes the previously auto-generated "File ID" in favor of creating a hidden UID that the app only uses internally to track a row. This makes it so the user doesn't have to see the "File ID." While I was reviewing this code I noticed I could combine some alter commands that hit the database and see a performance improvement so I adjusted those and add a constant `PreDefinedColumn` to track the pre-defined columns like "File ID." FWIW the reason for a unique ID is because we need some unique identifier for a row and if we can't assume "File Path" is unique (because of a request from AIBS) then we need to find or generate one so now we generate one secretly.

## Testing
[This is available on staging.](https://staging.biofile-finder.allencell.org/app)

### DIFF
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -21,10 +21,10 @@ const PRE_DEFINED_COLUMNS = Object.values(PreDefinedColumn);
  * Service reponsible for querying against a database
  */
 export default abstract class DatabaseService {
-    protected readonly SOURCE_METADATA_TABLE = "source_metadata";
     public static readonly LIST_DELIMITER = ",";
+    public static readonly HIDDEN_UID_ANNOTATION = "hidden_bff_uid";
     // Name of the hidden column BFF uses to uniquely identify rows
-    private static readonly HIDDEN_UID_COLUMN = "hidden_bff_uid";
+    protected readonly SOURCE_METADATA_TABLE = "source_metadata";
     // "Open file link" as a datatype must be hardcoded, and CAN NOT change
     // without BREAKING visibility in the dataset released in 2024 as part
     // of the EMT Data Release paper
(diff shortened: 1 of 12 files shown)


packages/core/services/DatabaseService/index.ts
```diff
@@ -7,21 +7,24 @@ import { Source } from "../../entity/FileExplorerURL";
...
+const PRE_DEFINED_COLUMNS = Object.values(PreDefinedColumn);
 
 /**
  * Service reponsible for querying against a database
  */
 export default abstract class DatabaseService {
-    protected readonly SOURCE_METADATA_TABLE = "source_metadata";
     public static readonly LIST_DELIMITER = ",";
+    public static readonly HIDDEN_UID_ANNOTATION = "hidden_bff_uid";
+    // Name of the hidden column BFF uses to uniquely identify rows
```
aswallace, line 26: tiny nit: comment & line 25 might be swapped?

aswallace (approved):

packages/core/services/DatabaseService/index.ts
SeanLeRoy: Done in https://github.com/AllenInstitute/biofile-finder/pull/323/commits/99d172e35b568cbab19eb35777fcd4f302c219e3

packages/core/services/DatabaseService/index.ts
```diff
@@ -173,51 +176,51 @@ export default abstract class DatabaseService {
...
-            await this.execute(`
+                ADD COLUMN ${DatabaseService.HIDDEN_UID_ANNOTATION} INT
+            `,
+            // Altering tables to add primary keys or serially generated columns
+            // isn't yet supported in DuckDB, so this does a serially generated
+            // column addition manually
+            `
                 UPDATE "${name}"
-                SET "File ID" = CONCAT(SQ.row, '-', '${dataSourceNameWithoutDate}')
+                SET "${DatabaseService.HIDDEN_UID_ANNOTATION}" = SQ.row
```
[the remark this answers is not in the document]
SeanLeRoy: >AIBS doesn't have identical rows when the filepaths collide, do they?

They do unfortunately

PR #324, opened 2024-11-14
Prevent duplicate loading of same description source

SeanLeRoy:
Currently there is a bug where description sources for data sources are being loaded in duplicate causing the queries against them to fail. This prevents that by improving our check to see if we have loaded this file before similar to how we do the actual data sources.

### DIFF
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -30,6 +30,7 @@ export default abstract class DatabaseService {
         ...Object.values(AnnotationType),
         DatabaseService.OPEN_FILE_LINK_TYPE,
     ]);
+    private sourceMetadataName?: string;
     private currentAggregateSource?: string;
     // Initialize with AICS FMS data source name to pretend it always exists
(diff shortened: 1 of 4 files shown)


aswallace (approved):
I'm not sure if I fully grasp what the original bug is/what the conditions to reproduce were. The logic for the PR looks sound to me though

packages/core/services/DatabaseService/index.ts
```diff
@@ -140,18 +141,25 @@ export default abstract class DatabaseService {
     }
 
     public async prepareSourceMetadata(sourceMetadata: Source): Promise<void> {
-        if (sourceMetadata.type && sourceMetadata.uri) {
-            this.deleteDataSource(this.SOURCE_METADATA_TABLE);
-            this.dataSourceToAnnotationsMap.clear();
+        const isOldSource = sourceMetadata.name === this.sourceMetadataName;
```
BrianWhitneyAI, line 144: I always find that things with "old" or "new" sort of get lost with time. Might want to add some context here or consider some different names.

BrianWhitneyAI (approved):

PR #337, opened 2024-11-19
Feature/add better column control

### LINKED ISSUE #329: Enable controlling the columns presented when sharing query
Yoav from AIBS wanted control over how users view their data, specifically what the columns shown are. This seems like something we could include in the query arguments like the groupings, filters, etc. Unsure if we would need to include the size of each column as well... could get tricky with different screens

SeanLeRoy:
## Description
Users have wanted to control which columns are shown in shared queries for a while. This adds them to the query arguments decoded/encoded.

While I was in here I noticed the columns & file view were tracked in multiple pieces. I added an interface and enum (respectively) to capture those pieces of information more simply.

## Testing
Tested against AICS FMS, cloud hosted, and local data sources

## Related Issue
Resolves #329

### DIFF
--- a/packages/core/components/EmptyFileListMessage/index.tsx
+++ b/packages/core/components/EmptyFileListMessage/index.tsx
@@ -8,12 +8,12 @@ export default function EmptyFileListMessage() {
         <div className={styles.emptyFileListContainer}>
             <div className={styles.emptyFileListMessage}>
                 <Icon className={styles.emptySearchIcon} iconName="SearchIssue" />
-                <h2>Sorry! No files found</h2>
+                <h2>No files match your query</h2>
                 <br />
-                <div>We couldn&apos;t find any files matching your request.</div>
+                <div>Double check your filters for any issues.</div>
                 <br />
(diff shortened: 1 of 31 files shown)


packages/core/components/EmptyFileListMessage/index.tsx
```diff
@@ -22,7 +22,7 @@ export default function EmptyFileListMessage() {
                     >
                         &nbsp;GitHub&nbsp;
                     </a>
-                    if you still expect there to be matches present.
+                    if you expect there should be matches present for query.
```
kmitcham, line 25: nit: maybe 'for this query'?

kmitcham (approved):

packages/core/components/EmptyFileListMessage/index.tsx
SeanLeRoy: oof will there ever be a day I don't have a typo

BrianWhitneyAI (approved):

PR #350, opened 2024-11-25
feature/toggle_status_report

### LINKED ISSUE #346: Cache status user notification
We need a way to inform users of the success or failure of a cache request. Right now this is not reported. The best way to do this is to use our existing success reporting with a message like X/X cache requests were successful, and separately raise individual failure messages for files, formatting the reasoning.

BrianWhitneyAI:
## Description 
The Purpose of this PR is to resolve #346 and add a notification system to users about their cache request.

### Succesful Caching
Screenshot 2024-11-25 at 1 05 22 PM

### Individual Failed File
Screenshot 2024-11-25 at 1 09 10 PM

### Failed Request (CORS)
Screenshot 2024-11-25 at 12 55 08 PM

### DIFF
--- a/packages/core/state/interaction/logics.ts
+++ b/packages/core/state/interaction/logics.ts
@@ -621,10 +621,7 @@ const moveFilesLogic = createLogic({
                     status === "DOWNLOAD_STARTED"
                 ) {
                     successfulFiles.push(fileId);
-                } else if (
-                    status === "FILE_RECORD_NOT_FOUND" ||
-                    status === "FILE_NOT_FOUND_ON_CLOUD"
-                ) {
+                } else {
                     failedFiles.push(fileId);
                 }
             });
@@ -636,7 +633,8 @@ const moveFilesLogic = createLogic({
                         "moveFilesSuccess",
                         `${successfulFiles.length} out of ${
                             Object.keys(cacheStatuses).length
-                        } files were successfully cached. Files will be available in the NAS (VAST) after downloads finish asynchronously`
+                        } files were successfully queued for download to NAS (Vast) from cloud. 
+                        Files will be available in the NAS after downloads finish asynchronously`
                     )
                 );
             }
--- a/packages/core/state/interaction/logics.ts
+++ b/packages/core/state/interaction/logics.ts
@@ -2,7 +2,7 @@ import { isEmpty, sumBy, throttle, uniq, uniqueId } from "lodash";
 import { AnyAction } from "redux";
 import { createLogic } from "redux-logic";
 
-import { metadata, ReduxLogicDeps, selection } from "../";
+import { interaction, metadata, ReduxLogicDeps, selection } from "../";
 import {
     DOWNLOAD_MANIFEST,
     DownloadManifestAction,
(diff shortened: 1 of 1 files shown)


packages/core/state/interaction/logics.ts
```diff
@@ -582,17 +582,81 @@ const setIsSmallScreen = createLogic({
...
+                return;
+            }
+
+            const successfulFiles: string[] = [];
+            const failedFiles: string[] = [];
+
+            Object.entries(cacheStatuses).forEach(([fileId, status]) => {
+                if (
+                    status === "DOWNLOAD_COMPLETE" ||
+                    status === "DOWNLOAD_IN_PROGRESS" ||
```
kmitcham, line 620: I have a probably overstated concern about the cache having failures after starting, like the network drops out or something.  I think the back-end has stuff to handle that, though, so this should be fine.

kmitcham (approved):

packages/core/state/interaction/logics.ts
```diff
@@ -582,17 +582,81 @@ const setIsSmallScreen = createLogic({
...
+                        `File "${fileName}" failed to cache. Status: ${cacheStatuses[fileId]}`
+                    )
+                );
+            });
         } catch (err) {
-            throw new Error(`Error encountered while moving files: ${err}`);
+            // Service call itself fails
+            console.log(err);
+            dispatch(
+                interaction.actions.processError("moveFilesFailure", `Failed to cache files.`)
```
aswallace, line 658: outside of this PR's scope, but noting that we may want to revisit how we define/use `processIds` for these status messages across the repo since it's currently somewhat ad hoc

packages/core/state/interaction/logics.ts
```diff
@@ -582,17 +582,81 @@ const setIsSmallScreen = createLogic({
...
-            console.log("Cache statuses:", cacheStatuses);
+            const fileDetails = (action as MoveFilesAction).payload.fileDetails;
+
+            // Map file IDs to file names for easy lookup
+            const fileIdToNameMap = Object.fromEntries(
+                fileDetails.map((file) => [file.id, file.name])
+            );
+
+            // Extract file IDs
+            const fileIds = fileDetails.map((file) => file.id);
```
aswallace, line 598: was unsure about the necessity of the single read `const` when `fileIdToNameMap` seems to duplicate the info, but also understand if the purpose is readability/separation of function

aswallace (approved):

packages/core/state/interaction/logics.ts
```diff
@@ -582,17 +582,81 @@ const setIsSmallScreen = createLogic({
...
+                    status === "DOWNLOAD_IN_PROGRESS" ||
+                    status === "DOWNLOAD_STARTED"
+                ) {
+                    successfulFiles.push(fileId);
+                } else if (
+                    status === "FILE_RECORD_NOT_FOUND" ||
+                    status === "FILE_NOT_FOUND_ON_CLOUD"
+                ) {
+                    failedFiles.push(fileId);
+                }
```
SeanLeRoy, line 629: Are there other statuses possible? I wonder if it would make more sense to just keep a list of "success" statuses and if the status isn't in that list then it is considered a failure

packages/core/state/interaction/logics.ts
```diff
@@ -582,17 +582,81 @@ const setIsSmallScreen = createLogic({
...
+            });
+
+            // Dispatch net success message. (Assuming some files were successful)
+            if (successfulFiles.length > 0) {
+                dispatch(
+                    interaction.actions.processSuccess(
+                        "moveFilesSuccess",
+                        `${successfulFiles.length} out of ${
+                            Object.keys(cacheStatuses).length
+                        } files were successfully cached.`
```
SeanLeRoy, line 639: "were successfully cached" is maybe misleading since this just kicks off the download process or refreshes it. I think something like "files will be available in the NAS after downloads finish asynchronously".

Do you have a message about how the downloads will become available in the modal that appears after clicking the copy option in the context menu?

packages/core/state/interaction/logics.ts
BrianWhitneyAI: These are all of the current status possibilities, as detailed[ here](https://github.com/AllenInstitute/biofile-finder/pull/322#issuecomment-2486626852). In the event of an unknown status, I don't think we should mark it as failed, as it could also be an alternate success. Perhaps a third option or warning?

packages/core/state/interaction/logics.ts
BrianWhitneyAI: I totally agree with the new wording. In regards to when the files will be available, I do not have any information for users. We do know that files are given a 180-day lease which there is a note about in the modal.

packages/core/state/interaction/logics.ts
BrianWhitneyAI: The idea was to make decoding the file id back to file name for error reporting more streamlined. I dont really have an opinion either way .

packages/core/state/interaction/logics.ts
BrianWhitneyAI: Resolved wording in [3726ec5](https://github.com/AllenInstitute/biofile-finder/pull/350/commits/3726ec57d03eed9085095800cc3a287af3a29398)

SeanLeRoy: > files were successfully cached. Files will be available in the NAS (VAST) after downloads finish asynchronously

This reads like the "cache" & "NAS" are different things (as in we have past tense cached it, but it still needs downloading to the NAS). Maybe just keep it vague like: 

"files were successfully queued for download to NAS (Vast) from cloud. Files will be available in the NAS after downloads finish asynchronously"

BrianWhitneyAI: resolved in [60cf69a](https://github.com/AllenInstitute/biofile-finder/pull/350/commits/60cf69a2a3c544e763e29152194146d5022d176e)

packages/core/state/interaction/logics.ts
```diff
@@ -582,17 +582,78 @@ const setIsSmallScreen = createLogic({
...
+                        `moveFileFailure_${fileId}`,
+                        `File "${fileName}" failed to cache. Status: ${cacheStatuses[fileId]}`
+                    )
+                );
+            });
         } catch (err) {
-            throw new Error(`Error encountered while moving files: ${err}`);
+            // Service call itself fails
+            dispatch(
+                interaction.actions.processError("moveFilesFailure", `Failed to cache files.`)
```
SeanLeRoy, line 655: Might be good just to toss the error from the err.message here like
```suggestion
                interaction.actions.processError("moveFilesFailure", `Failed to cache files, details: ${(err as Error).message}.`)
```

SeanDuHare (approved):

PR #353, opened 2024-11-26
Intercept modal dismissal with unsaved changes

### LINKED ISSUE #344: Intercept onDismiss/exit in edit modal
When user tries to exist tab or dismiss the modal (by clicking outside or clicking x/cancel), intercept and render a warning modal letting them know the changes won't be saved

Part of #96

aswallace:
Addresses two small requests for the editing modal: 
- Removes redundant tooltips from buttons
- Adds a warning pop-up if user dismisses the modal after they've already entered annotation data
   - Warning logic needs to be global to the whole edit modal instead of just the form component because dismissing can mean clicking cancel, clicking the 'x' or just clicking outside of the modal
   - The pop-up currently only appears for the 'New annotation' pathway after a user enters data for the new ann. and clicks next (per discussion, anything before that is easy to re-enter so warning is not needed)
   - Uses show/hide with css instead of js conditionals since we need the rendered data to persist

closes #344 

image

### DIFF
--- a/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
+++ b/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
@@ -18,6 +18,7 @@ enum EditStep {
 
 interface NewAnnotationProps {
     onDismiss: () => void;
+    onUnsavedChanges: () => void;
     selectedFileCount: number;
 }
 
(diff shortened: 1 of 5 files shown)


kmitcham (approved):

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
```diff
@@ -137,14 +144,14 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
                     <StackItem>
                         {step === EditStep.EDIT_FILES && (
                             <SecondaryButton
-                                title="Go back"
+                                title=""
```
BrianWhitneyAI, line 147: No title for the button?

BrianWhitneyAI (approved):

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
aswallace: Is intentional. Was UX request since when the title is the same as the text on the button, the tooltip is distracting and doesn't add any info. Planned to make ticket to make titles optional for buttons

PR #338, opened 2024-11-19
[Large] Create UI for metadata editing workflow

aswallace:
This PR creates the main part of the UI for editing file metadata from BFF based on [the Figma designs](https://www.figma.com/design/oNQvOAK11YINcbPvNmkgwO/Public-File-Explorer?node-id=875-4724&node-type=canvas&t=JUN2ieb9MDJMnUxO-0). Does not include any backend logic.

~[Temporarily deployed to staging](https://staging.biofile-finder.allencell.org/app)~

Part of #96 

## Reviewing
Lots of changes with a slightly complicated file structure, so here's suggested review order:
- Entrypoint: added edit to the dropdown menu in [packages/core/hooks/useFileAccessContextMenu.ts](https://github.com/AllenInstitute/biofile-finder/pull/338/files#diff-71bcb3b6133f882ebdd1dd66c282d671a299b33032af1c33744369ec05c99339)
- This opens the modal, which is [packages/core/components/Modal/EditMetadata/index.tsx](https://github.com/AllenInstitute/biofile-finder/pull/338/files#diff-a6ec79eb3ac34e5ecd40a3aead96f6e047d6d3143b78100b3d65728380350a9f)
- The modal renders the actual editing component ([packages/core/components/EditMetadata/index.tsx](https://github.com/AllenInstitute/biofile-finder/pull/338/files#diff-264e2d9d82ffe4d96144ff07cb2dac39ff71aab629579b15dbf783ce1a68d9fc)), which is a wrapper for both the "new" and "existing" annotation pathways 
    - Pathway for editing files' values for an existing annotation: [packages/core/components/EditMetadata/ExistingAnnotationPathway.tsx](https://github.com/AllenInstitute/biofile-finder/pull/338/files#diff-f4052c195644362a2061021276d6bb4298a6a4d91c4cedd54acd4b25a461a490)
    - Pathway for creating a new annotation on the files: [packages/core/components/EditMetadata/NewAnnotationPathway.tsx](https://github.com/AllenInstitute/biofile-finder/pull/338/files#diff-a6c315bc57c9c1a4506e78753a90b1b1a4445bd9f28a80ccdca3e358b3f91b25)
[...]

PR #354, opened 2024-12-06
Add custom input fields for editing metadata

aswallace:
This is a follow-up to #338 that adds custom fields for different data types based on [the Figma designs](https://www.figma.com/design/oNQvOAK11YINcbPvNmkgwO/Public-File-Explorer?node-id=3138-10352&node-type=frame&t=FxEJXP0Ufqmnsazm-0).

### Description
The input type switch is modeled off the way we did it in AnnotationFilterForm and re-uses some of the same components.
- Uses our existing choice group/text field/date field components for boolean/string/date+datetime respectively
- Adds a new "duration" form type
- Creates a new numerical input component to be more stylistically consistent (and replaced usage in our NumberRangePicker component)

### Screenshots
Duration 
new duration form
Boolean
choice group
Date/Date Time
date picking form
Dropdown
combo box
Number
new numerical input
Text/Default
default for any type

### DIFF
--- a/packages/core/components/ComboBox/index.tsx
+++ b/packages/core/components/ComboBox/index.tsx
@@ -24,7 +24,6 @@ interface Props {
     multiSelect?: boolean;
     options: IComboBoxOption[];
     placeholder: string;
-    useComboBoxAsMenuWidth?: boolean;
     onChange?: (option: IComboBoxOption | undefined, value?: string | undefined) => void;
 }
 
@@ -88,7 +87,7 @@ export default function BaseComboBox(props: Props) {
                 callout: styles.comboBoxCallout,
                 optionsContainer: styles.optionsContainer,
             }}
-            useComboBoxAsMenuWidth={props?.useComboBoxAsMenuWidth}
+            useComboBoxAsMenuWidth
         />
     );
 }
--- a/packages/core/components/DurationForm/index.tsx
+++ b/packages/core/components/DurationForm/index.tsx
@@ -17,10 +17,10 @@ interface DurationFormProps {
  */
 export default function DurationForm(props: DurationFormProps) {
     const { onChange } = props;
-    const [days, setDurationDays] = React.useState<string>("");
-    const [hours, setDurationHours] = React.useState<string>("");
-    const [minutes, setDurationMinutes] = React.useState<string>("");
-    const [seconds, setDurationSeconds] = React.useState<string>("");
+    const [days, setDurationDays] = React.useState<string>("0");
+    const [hours, setDurationHours] = React.useState<string>("0");
+    const [minutes, setDurationMinutes] = React.useState<string>("0");
+    const [seconds, setDurationSeconds] = React.useState<string>("0");
     const durationFormatter = annotationFormatterFactory(AnnotationType.DURATION);
 
     React.useEffect(() => {
@@ -39,32 +39,32 @@ export default function DurationForm(props: DurationFormProps) {
                     aria-label="Days"
                     className={styles.inputField}
                     id="durationDays"
+                    label="Days"
                     onChange={(event) => setDurationDays(event?.target?.value || "")}
-                    placeholder="Days..."
                     defaultValue={days}
                 />
                 <NumberField
                     aria-label="Hours"
                     className={styles.inputField}
                     id="durationHours"
+                    label="Hrs"
                     onChange={(event) => setDurationHours(event?.target?.value || "")}
-                    placeholder="Hrs..."
                     defaultValue={hours}
                 />
                 <NumberField
                     aria-label="Minutes"
                     className={styles.inputField}
                     id="durationMinutes"
+                    label="Mins"
                     onChange={(event) => setDurationMinutes(event?.target?.value || "")}
-                    placeholder="Mins..."
                     defaultValue={minutes}
                 />
                 <NumberField
                     aria-label="Seconds"
                     className={styles.inputField}
                     id="durationSeconds"
+                    label="Secs"
                     onChange={(event) => setDurationSeconds(event?.target?.value || "")}
-                    placeholder="Secs..."
                     defaultValue={seconds}
                 />
             </div>
--- a/packages/core/components/DurationForm/index.tsx
+++ b/packages/core/components/DurationForm/index.tsx
@@ -1,0 +1,73 @@
+import * as React from "react";
+
+import NumberField from "../NumberRangePicker/NumberField";
+import annotationFormatterFactory, { AnnotationType } from "../../entity/AnnotationFormatter";
+
+import styles from "./DurationForm.module.css";
+
+interface DurationFormProps {
+    className?: string;
+    defaultValue?: number;
+    onChange: (totalDuration: number) => void;
+    title?: string;
+}
+
+/**
+ * This component renders a simple form for entering durations
+ */
+export default function DurationForm(props: DurationFormProps) {
+    const { onChange } = props;
+    const [days, setDurationDays] = React.useState<string>("");
+    const [hours, setDurationHours] = React.useState<string>("");
+    const [minutes, setDurationMinutes] = React.useState<string>("");
+    const [seconds, setDurationSeconds] = React.useState<string>("");
+    const durationFormatter = annotationFormatterFactory(AnnotationType.DURATION);
+
+    React.useEffect(() => {
+        const durationString = `${Number(days) || 0}D ${Number(hours) || 0}H ${
+            Number(minutes) || 0
+        }M ${Number(seconds) || 0}S`;
+        const totalDurationInMs = Number(durationFormatter.valueOf(durationString));
+        onChange(totalDurationInMs);
+    }, [days, hours, minutes, seconds, durationFormatter, onChange]);
+
+    return (
+        <div>
+            <h3 className={styles.title}>{props?.title}</h3>
+            <div className={styles.inputWrapper}>
+                <NumberField
+                    aria-label="Days"
+                    className={styles.inputField}
+                    id="durationDays"
+                    onChange={(event) => setDurationDays(event?.target?.value || "")}
+                    placeholder="Days..."
+                    defaultValue={days}
+                />
+                <NumberField
+                    aria-label="Hours"
+                    className={styles.inputField}
+                    id="durationHours"
+                    onChange={(event) => setDurationHours(event?.target?.value || "")}
+                    placeholder="Hrs..."
+                    defaultValue={hours}
+                />
+                <NumberField
+                    aria-label="Minutes"
+                    className={styles.inputField}
+                    id="durationMinutes"
+                    onChange={(event) => setDurationMinutes(event?.target?.value || "")}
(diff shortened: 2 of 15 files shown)


lynwilhelm:
@aswallace Great work! Just 3 feedback notes:

1. Duration: The labels are overlapping the inputs. There should be a couple pixels of space in between.
2. Hint text should be consistent with pattern, which is to describe what the user should do using an action word. For example "Type an option name..." So, Text/default should read "Type in value(s)" and Number should say "Type a numerical value..."
3. Can the little calendar be the same color as text so it doesn't look disabled?

Thanks for sharing!

aswallace:
Thank you @lynwilhelm! Addressed feedback in [6c52646](https://github.com/AllenInstitute/biofile-finder/pull/354/commits/6c52646965a65d329630a4d0ffcdb9340c04a135) and updated the screenshots to reflect

packages/core/components/DurationForm/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+                <NumberField
+                    aria-label="Days"
+                    className={styles.inputField}
+                    id="durationDays"
+                    label="Days"
+                    onChange={(event) => setDurationDays(event?.target?.value || "")}
+                    defaultValue={days}
+                />
+                <NumberField
+                    aria-label="Hours"
```
kmitcham, line 47: Is there code somewhere to check that hours < 24, minutes < 60, and seconds < 60?

packages/core/components/EditMetadata/MetadataDetails.tsx
```diff
@@ -41,20 +52,95 @@ export default function EditMetadataDetailsList(props: DetailsListProps) {
...
         _: number | undefined,
         column: IColumn | undefined
     ) {
         const fieldContent = item[column?.fieldName as keyof ValueCountItem] as string;
-        if (!fieldContent) return "[No value] (blank)";
-        if (column?.fieldName === "fileCount") {
+        if (column?.fieldName === "value") {
+            if (!fieldContent) return "[No value] (blank)";
+            else return annotationFormatter.displayValue(fieldContent);
+        } else if (column?.fieldName === "fileCount") {
```
kmitcham, line 68: so fileCount will always have a value?

packages/core/components/DurationForm/index.tsx
aswallace: Not currently, but I can add min/max restrictions on the UI

packages/core/components/EditMetadata/MetadataDetails.tsx
aswallace: it should. I'll add a check for clarity though

packages/core/components/DurationForm/index.tsx
aswallace: added in https://github.com/AllenInstitute/biofile-finder/pull/354/commits/1da0c7f16e2d21319ba164760615e2b8a5c7a1f7

packages/core/components/EditMetadata/MetadataDetails.module.css
```diff
@@ -62,3 +67,64 @@
...
+.date-range-text-field > div {
+  border: 1px solid var(--border-color)
+}
+
+.date-range-text-field i {
+  color: var(--primary-text-color)
+}
+
+.no-padding {
+  padding-bottom: 0 !important;
```
pgarrison, line 125: Leave a comment about why the `!important` is necessary?

packages/core/entity/AnnotationFormatter/duration-formatter.ts
```diff
@@ -19,12 +20,22 @@ export default {
...
         // Since seconds can have values like "3.5S", don't use floor when calculating
-        addUnit(1000, "S", false);
+        addUnit(msInASecond, "S", false);
 
         return display.trim();
     },
 
     valueOf(value: any) {
+        const RANGE_OPERATOR_REGEX = /([0-9]+)D ([0-9]+)H ([0-9]+)M ([0-9]*\.?[0-9]+)S/g;
+        const exec = RANGE_OPERATOR_REGEX.exec(value);
```
pgarrison, line 30: Maybe something like `matches` is  more clear than `exec`?

packages/core/components/NumberRangePicker/NumberField.tsx
```diff
@@ -0,0 +1,39 @@
...
+            <input
+                data-testid={props.id}
+                id={props.id}
+                type="number"
+                value={props.defaultValue}
+                step="any"
+                onChange={props.onChange}
+                placeholder={props?.placeholder}
+                min={props?.min}
+                max={props?.max}
```
pgarrison, line 35: From playing with [this example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) input, it seems that `min`and `max` control what the increment/decrement buttons do, but not values entered via keyboard. More dangerous than values out of bounds are non-numeric values. What happens if the user enters `10D 2H 100M 10S` into the hour field?

packages/core/components/DurationForm/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+                    className={styles.inputField}
+                    id="durationHours"
+                    label="Hrs"
+                    onChange={(event) => setDurationHours(event?.target?.value || "")}
+                    defaultValue={hours}
+                    min={0}
+                    max={24}
```
pgarrison, line 55: I think these `max` values are inclusive? In which case these bounds (here and below) should be 23 and 59

pgarrison (approved):
This UI looks so nice!

packages/core/components/DurationForm/index.tsx
aswallace: good call, I'm used to thinking of max as exclusive. Fixed in [b1d134a](https://github.com/AllenInstitute/biofile-finder/pull/354/commits/b1d134ad9a522f43a9135affb81e83607db8eace)

packages/core/components/NumberRangePicker/NumberField.tsx
aswallace: I may be misunderstanding the question, but users are not able to enter non-numeric values in this type of input (with the possible exception of `e`). I've added additional in-bounds validation in [b1d134a](https://github.com/AllenInstitute/biofile-finder/pull/354/commits/b1d134ad9a522f43a9135affb81e83607db8eace)

packages/core/components/EditMetadata/MetadataDetails.module.css
aswallace: these importants were not, in fact, important. Remnants of UI debugging. Removed in [b1d134a](https://github.com/AllenInstitute/biofile-finder/pull/354/commits/b1d134ad9a522f43a9135affb81e83607db8eace)

packages/core/components/NumberRangePicker/NumberField.tsx
pgarrison: Great! I didn't open BFF, just tested against a generic `<input type="number" />` element

packages/core/entity/AnnotationFormatter/duration-formatter.ts
aswallace: fixed in [761fa46](https://github.com/AllenInstitute/biofile-finder/pull/354/commits/761fa46a0c786142ff324d5007c58dc293e24b0b)

packages/core/components/ComboBox/index.tsx
```diff
@@ -67,7 +66,7 @@ export default function BaseComboBox(props: Props) {
             comboBoxOptionStyles={{
                 rootChecked: styles.comboBoxItemChecked,
             }}
-            useComboBoxAsMenuWidth={props?.useComboBoxAsMenuWidth}
+            useComboBoxAsMenuWidth={true}
```
SeanLeRoy, line 69:
```suggestion
            useComboBoxAsMenuWidth
```

packages/core/components/DurationForm/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+    return (
+        <div>
+            <h3 className={styles.title}>{props?.title}</h3>
+            <div className={styles.inputWrapper}>
+                <NumberField
+                    aria-label="Days"
+                    className={styles.inputField}
+                    id="durationDays"
+                    label="Days"
+                    onChange={(event) => setDurationDays(event?.target?.value || "")}
```
SeanLeRoy, line 43: Seems like this component could exclusively pass around `Number` type values (for cases with leading zeroes it would just not yet return a value AKA `undefined`); NBD to keep this though

packages/core/components/EditMetadata/EditMetadata.module.css
```diff
@@ -81,7 +81,7 @@
 
 .text-field {
   padding-bottom: var(--margin);
-  max-width: 300px;;
+  width: 300px;;
```
SeanLeRoy, line 84:
```suggestion
  width: 300px;
```

packages/core/components/EditMetadata/MetadataDetails.tsx
```diff
@@ -41,20 +52,95 @@ export default function EditMetadataDetailsList(props: DetailsListProps) {
...
+            return <div className={styles.columnRightAlignCell}>{fieldContent || 0}</div>;
         }
         return fieldContent;
     }
 
+    const inputField = () => {
+        switch (props.fieldType) {
+            case AnnotationType.DATE:
+            case AnnotationType.DATETIME:
```
SeanLeRoy, line 77: Is there no time selector for `DATETIME` types or is there always a time selector for `DATE` types? Either of those would be bad no?

packages/core/components/ComboBox/index.tsx
aswallace: This was deliberate; FluentUI was not registering the property correctly. I'm not able to replicate the bug now though, so will fix & keep an eye on it

packages/core/components/EditMetadata/MetadataDetails.tsx
aswallace: Unfortunately yes, the date picker does not provide a time selector. This issue also exists in the AnnotationFilterForm

packages/core/components/DurationForm/index.tsx
aswallace: I agree in theory, but running into some implementation bugs

packages/core/components/EditMetadata/MetadataDetails.tsx
aswallace: I'm opening a separate ticket for this because the FluentUI TimePicker is imo too buggy to use (and the API does not match the component), so making our own is just as good of an option but will hold up this ticket

aswallace: added in [9328a44](https://github.com/AllenInstitute/biofile-finder/pull/354/commits/9328a44ecfedb6b82b1fedfbbaf73679dbb37633)
image

SeanLeRoy (approved):

PR #358, opened 2024-12-11
Feature/bff

imashmash:
This PR adds BFF domain S3 bucket and CloudFront distribution ID
Test:
https://github.com/AllenInstitute/biofile-finder/actions/runs/12282978093

### DIFF
--- a/.github/workflows/aws-deployment.yml
+++ b/.github/workflows/aws-deployment.yml
@@ -17,13 +17,11 @@ env:
   AWS_ACCOUNT_ID: ${{ vars.AWS_PUBLIC_DATA_RELEASES_ACCOUNT_ID }}
   AWS_REGION: ${{ vars.AWS_DEFAULT_REGION }}
   STAGING_CLOUDFRONT_DISTRIBUTION_ID: ${{ secrets.STAGING_CLOUDFRONT_DISTRIBUTION_ID }}
-  STAGING_S3_BUCKET: s3://staging.biofile-finder.allencell.org
+  STAGING_S3_BUCKET: s3://staging.bff.allencell.org
   STAGING_BFF_CLOUDFRONT_DISTRIBUTION_ID: ${{ secrets.STAGING_BFF_CLOUDFRONT_DISTRIBUTION_ID }}
-  STAGING_BFF_S3_BUCKET: s3://staging.bff.allencell.org
   PRODUCTION_CLOUDFRONT_DISTRIBUTION_ID: ${{ secrets.PRODUCTION_CLOUDFRONT_DISTRIBUTION_ID }}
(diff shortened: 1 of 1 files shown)


imashmash:
Branch "feature/bff" is not allowed to deploy to production due to environment protection rules. - that's why it failed

anl (commented):
Are we ever going back to biofile-finder.allencell.org?  I think we may no longer need to publish to that bucket (in fact, we can probably[?] delete that bucket, and just use CloudFront to redirect requests, right?).

gmoynihan88 (approved):

.github/workflows/aws-deployment.yml
```diff
@@ -89,10 +91,13 @@ jobs:
           name: aws-deploy-files
           path: ./packages/web/dist
 
-      # Note that the command below will copy the files to the root of the S3 bucket e.g., s3://biofile-finder.allencell.org/
-      # If you want to copy files to a S3 prefix / subdirectory, you would want something like ${{ env.S3_BUCKET }}/your_prefix below
-      - name: Copy build files to S3 root
-        run: aws s3 sync ./packages/web/dist ${{ env.S3_BUCKET }}
+      # Note that the command below will copy the files to the root of the S3 bucket e.g., s3://bff.allencell.org/
+      # If you want to copy files to a S3 prefix / subdirectory, you would want something like ${{ env.BFF_S3_BUCKET }}/your_prefix below
+      - name: Copy build filed to BFF S3
```
anl, line 96:
```suggestion
      - name: Copy build files to BFF S3
```

.github/workflows/aws-deployment.yml
```diff
@@ -89,10 +91,13 @@ jobs:
...
+        run: aws s3 sync ./packages/web/dist ${{ env.BFF_S3_BUCKET }}
 
-      - name: Invalidate CloudFront cache
+      - name: Invalidate CloudFront biofile-finder cache
```
anl, line 99: Maybe add a comment that this isn't strictly necessary since we aren't writing anything served directly by this distribution, since all it does is issue redirects.

anl (approved):
@pgarrison @aswallace @BrianWhitneyAI @yrkim98 @tyler-foster Could one of you review this PR also?

SeanDuHare (commented):
It doesn’t seem like this PR is directly responsible, but I would prefer this change be entirely reverted if the issue I mentioned in slack below will continue.

https://allencellscience.slack.com/archives/C04DNM9G43S/p1734025634278819?thread_ts=1733852921.584909&channel=C04DNM9G43S&message_ts=1734025634.278819

BrianWhitneyAI (approved):

Issue #341, opened 2024-11-22
Add password logic to edit modal


aswallace:
Users will need to enter a password before being able to access the editing modal.
Create a service user in LK for each password we make and keep a mapping of service user to password

Part of #96

PR #372, opened 2024-12-19
[Large] Create logic for editing metadata

aswallace:
Adds editing logic for **existing** annotations for FMS and database (csv/duckdb) files. This reopens/continues Sean's work on the edit logic from #260 with minor updates to reflect codebase & MMS changes. The commit history will be a mess because of rebases, but the file changes are accurate.

This is a big PR, but the most important changes are the FileService additions in [packages/core/services/FileService/DatabaseFileService/index.ts](https://github.com/AllenInstitute/biofile-finder/pull/372/files#diff-11fe0da689506ce774718e44f9ec28279bb2188046b05345c0541cc8ac3fe854) and [packages/core/services/FileService/HttpFileService/index.ts](https://github.com/AllenInstitute/biofile-finder/pull/372/files#diff-adfa3b8fe62502ba6a003096ddf39d22f931e66c4211032d46dd0abec01ea1b5). The logic for sending batches of edit requests via Redux is in [packages/core/state/interaction/logics.ts](https://github.com/AllenInstitute/biofile-finder/pull/372/files#diff-138bf3638bd4fb462a81c5b387e746dfb4537d8c11c64b6ca8cfce593b8b6573). We also added a `put` to the `HttpFileService`. Most other changes are to interfaces, abstract classes or tests.

### Testing
- Unit tests from old PR
- Successfully edited files in FES staging while pointing to MMS staging (which has the CORS updates)
- Successfully edited metadata from an uploaded csv, and confirmed that edits are included when you download as CSV

### Notes: 

Editing FMS files on web WILL NOT WORK unless you add a valid `"X-User-Id"` to the headers in HttpServiceBase. I didn't commit this because I'm currently using my own user ID locally until we complete #341 

The UI occasionally takes a while to display correct data for edited FMS files, even though the edits go through to MMS immediately. I think this is an FMS/FES syncing issue rather than a BFF thing?


packages/core/state/interaction/logics.ts
```diff
@@ -499,6 +501,105 @@ const openWithLogic = createLogic({
...
+            // Begin editing files in batches
+            for (const batch of batches) {
+                // Asynchronously begin the edit for each file in the batch
+                const promises = batch.map(
+                    (fileId) =>
+                        new Promise<void>(async (resolve, reject) => {
+                            try {
+                                await fileService.editFile(
+                                    fileId,
+                                    annotations,
```
SeanLeRoy, line 573: I see... I think the request to get MMS "editable data" was to for insertion right here where this `annotations` parameter is. However, that is a very slow `GET` fetch so I don't think we should necessarily switch to that. Instead, can we just pass here the annotation we want to replace the values for? Lets check up on the way the `PATCH` method works we might be golden with just switching to that.

packages/core/state/interaction/logics.ts
```diff
@@ -499,6 +501,105 @@ const openWithLogic = createLogic({
...
+        // Break files into batches of 10 File IDs
+        const fileIds = filesSelected.map((file) => file.uid);
+        const batches = chunk(fileIds, 10);
+
+        // Dispatch an event to alert the user of the start of the process
+        const editRequestId = uniqueId();
+        const editProcessMsg = "Editing files in progress.";
+        // TODO: Lyndsay's design did not include a progress bar for editing files
+        // nor a way to cancel the process. This is a placeholder for now.
+        dispatch(processStart(editRequestId, editProcessMsg, noop));
```
SeanLeRoy, line 546: hmm... ig thats fine perhaps drop the TODO then?

packages/core/state/interaction/logics.ts
SeanLeRoy: Alright found the logic in MMS, this is the logical path:
https://github.com/AllenCellSoftware/metadata-management-service/blob/master/src/main/java/org/alleninstitute/aics/metadatamanagementservice/filemetadata/FileMetadataController.java#L92
https://github.com/AllenCellSoftware/metadata-management-service/blob/master/src/main/java/org/alleninstitute/aics/metadatamanagementservice/filemetadata/FileIngestionService.java#L101C13-L101C40
https://github.com/AllenCellSoftware/metadata-management-service/blob/master/src/main/java/org/alleninstitute/aics/metadatamanagementservice/filemetadata/FileIngestionService.java#L171-L205

So based on the last chunk of code there we actually want that boolean block of `replaceOldValues` to happen which does not happen with the `PATCH` and does happen with the `PUT`. Furthermore, it only replaces annotations that are included in the request with new values. So all to get this working that we need to do is for the `annotations` parameter in this BFF code assign `annotations` equal to just the annotation the user modified in the modal with the values they want to replace with from the modal. It seems like that may already be happening so everything might just be all good.

SeanLeRoy (approved):

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -127,4 +151,56 @@ export default class HttpFileService extends HttpServiceBase implements FileServ
...
+                    `Unable to edit file. Failed to find annotation id for annotation ${name}`
+                );
+            }
+            return { annotationId, values };
+        });
+        const requestBody = JSON.stringify({ customMetadata: { annotations } });
+        await this.put(url, requestBody);
+    }
+
+    public async getEditableFileMetadata(
```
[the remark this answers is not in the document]
SeanLeRoy: btw this def seems unused now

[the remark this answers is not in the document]
aswallace: Yes, I temporarily kept both of them in for debugging purposes but at this point can be removed

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -127,4 +151,56 @@ export default class HttpFileService extends HttpServiceBase implements FileServ
...
+            const annotationId = annotationNameToAnnotationMap?.[name].id;
+            if (!annotationId) {
+                throw new Error(
+                    `Unable to edit file. Failed to find annotation id for annotation ${name}`
+                );
+            }
+            return { annotationId, values };
+        });
+        const requestBody = JSON.stringify({ customMetadata: { annotations } });
+        await this.put(url, requestBody);
```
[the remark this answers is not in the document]
aswallace: I'm fairly confident we want `PUT`. `PATCH` does not allow us to replace existing values for annotations. I've been testing on `PUT`, and while it does replace all values for a given annotation, it does not expect all metadata for a file. That is, I can call a put with only the annotation I want to edit, and the rest of the annotations will be maintained

packages/core/state/interaction/logics.ts
aswallace: Yes, currently we are only passing the annotations that we want to edit

aswallace: Just to be clear, this is currently working as is

packages/core/state/interaction/logics.ts
aswallace: removed in [e885706](https://github.com/AllenInstitute/biofile-finder/pull/372/commits/e885706c8a2a107d4fd473cacbeb720e0c5c5088)

packages/core/services/FileService/HttpFileService/index.ts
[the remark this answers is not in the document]
aswallace: removed in [e885706](https://github.com/AllenInstitute/biofile-finder/pull/372/commits/e885706c8a2a107d4fd473cacbeb720e0c5c5088)

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -16,13 +23,30 @@ interface Config extends ConnectionConfig {
...
+        annotationId: number;
+        values: string[];
+    }[];
+    templateId?: number;
+}
+
+const FESBaseUrlToMMSBaseUrlMap = {
+    [FileExplorerServiceBaseUrl.LOCALHOST]: "https://localhost:9060",
+    [FileExplorerServiceBaseUrl.STAGING]: "https://stg-aics.corp.alleninstitute.org",
+    [FileExplorerServiceBaseUrl.PRODUCTION]: "https://stg-aics.corp.alleninstitute.org", // TO DO: unsure if using stg for prod was intentional
```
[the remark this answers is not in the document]
aswallace: fixed in [e885706](https://github.com/AllenInstitute/biofile-finder/pull/372/commits/e885706c8a2a107d4fd473cacbeb720e0c5c5088)

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -184,4 +189,43 @@ export default class DatabaseFileService implements FileService {
...
+        );
+        const sql = `\
+            UPDATE '${tableName}' \
+            SET ${columnAssignments.join(", ")} \
+            WHERE ${DatabaseService.HIDDEN_UID_ANNOTATION} = '${fileId}'; \
+        `;
+        return this.databaseService.execute(sql);
+    }
+
+    public async getEditableFileMetadata(
```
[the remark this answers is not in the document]
aswallace: removed in [e885706](https://github.com/AllenInstitute/biofile-finder/pull/372/commits/e885706c8a2a107d4fd473cacbeb720e0c5c5088)

kmitcham:
Is it legal to delete an annotation?  Or set the value to null?

kmitcham (approved):

aswallace:
> Is it legal to delete an annotation? Or set the value to null?

Technically yes, passing an empty array as the values effectively deletes annotations when you use the MMS `put` endpoint. However, the current UI in BFF does not allow you to click submit on the form without entering values (and just spotted that sending an empty string causes MMS to send back an error, so I'm adding a check for that now)

packages/core/components/StatusMessage/StatusMessage.module.css
```diff
@@ -4,6 +4,7 @@
     width: 40vw; 
     height: 100%;
     left: 30px;
+    /* FluentUI modal z-index is set to 1000000 */
```
BrianWhitneyAI, line 7: Is this comment right seems like 1000 not 1000000?

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -127,4 +141,24 @@ export default class HttpFileService extends HttpServiceBase implements FileServ
...
+    public async editFile(
+        fileId: string,
+        annotationNameToValuesMap: AnnotationNameToValuesMap,
+        annotationNameToAnnotationMap?: Record<string, Annotation>
+    ): Promise<void> {
+        const mmsBaseUrl = FESBaseUrlToMMSBaseUrlMap[this.baseUrl as FileExplorerServiceBaseUrl];
+        const url = `${mmsBaseUrl}/${HttpFileService.BASE_EDIT_FILES_URL}/${fileId}`;
+        const annotations = Object.entries(annotationNameToValuesMap).map(([name, values]) => {
+            const annotationId = annotationNameToAnnotationMap?.[name].id;
+            if (!annotationId) {
```
BrianWhitneyAI, line 154: it might be better to aggregate all missing annotations and throw a single error

packages/core/services/FileService/HttpFileService/test/HttpFileService.test.ts
```diff
@@ -44,6 +44,34 @@ describe("HttpFileService", () => {
...
+            const httpFileService = new HttpFileService({
+                baseUrl,
+                httpClient,
+                downloadService: new FileDownloadServiceNoop(),
+            });
+
+            // Act / Assert
+            try {
+                await httpFileService.editFile("file_id", { ["Color"]: ["red"] });
+                expect(false, "Expected to throw").to.be.true;
```
BrianWhitneyAI, line 66: I think this is redundant?

packages/core/services/HttpServiceBase/index.ts
```diff
@@ -210,6 +210,30 @@ export default class HttpServiceBase {
...
+        } catch (err) {
+            // Specific errors about the failure from services will be in this path
+            if (axios.isAxiosError(err) && err?.response?.data?.message) {
+                throw new Error(JSON.stringify(err.response.data.message));
+            }
+            throw err;
+        }
+
+        if (response.status >= 400 || response.data === undefined) {
+            // by default axios will reject if does not satisfy: status >= 200 && status < 300
```
BrianWhitneyAI, line 230: if the default already rejects why is this a condition?

packages/core/services/HttpServiceBase/index.ts
```diff
@@ -210,6 +210,30 @@ export default class HttpServiceBase {
...
+            response = await retry.execute(() => this.httpClient.put(encodedUrl, body, config));
+        } catch (err) {
+            // Specific errors about the failure from services will be in this path
+            if (axios.isAxiosError(err) && err?.response?.data?.message) {
+                throw new Error(JSON.stringify(err.response.data.message));
+            }
+            throw err;
+        }
+
+        if (response.status >= 400 || response.data === undefined) {
```
BrianWhitneyAI, line 229: Could you have a valid response with no data?

BrianWhitneyAI (approved):

packages/core/components/StatusMessage/StatusMessage.module.css
aswallace: Yes, this is for the status message component. The modal z-index is set by fluent, not us, so this comment is intended to note that the status message will be covered by the modal overlay. I can clarify or remove this comment

packages/core/services/FileService/HttpFileService/test/HttpFileService.test.ts
aswallace: I'm not the author of this one so I may not have the right explanation, but I think it's a method of throwing custom error messages in chai?

packages/core/services/HttpServiceBase/index.ts
aswallace: This is a pattern used by all the http service methods that made sense to me before but which I'm now having trouble defending. My interpretation is that when axios rejects a call, it does not necessarily throw a javascript error object and may bypass the catch, so the status must be checked if we want it to throw something

packages/core/services/HttpServiceBase/index.ts
aswallace: Another case of using the pattern but not necessarily being able to defend. I believe we expect to always receive at least the state of the data from a success response

PR #375, opened 2024-12-20
367: Remove Canonical path and format Cloud path

pgarrison:
# Purpose
With the new cloud-first FMS backend features, the `file_path` from FMS is a cloud path like `staging.files.allencell.org/c18/726/e93/ea3/804/468/34d/873/d6c/435/c6/aicsfiles-test-8c85cfec-bdc3-11ef-9089-0242ac110004.czi`. This PR renders this as: `File Path (Cloud) https://s3.us-west-2.amazonaws.com/staging.files.allencell.org/c18/726/e93/ea3/804/468/34d/873/d6c/435/c6/aicsfiles-test-8c85cfec-bdc3-11ef-9089-0242ac110004.czi`.

In addition, this PR removes the `File Path (Canonical)`. Along with #370, this completes the move from:
```
File Path (Canonical) /allen/aics/software/apps/staging/fss/data/7fb/6fb/b0f/6bd/7d5/1db/a4c/1e4/2a6/2a5/c7/aicsfiles-test-a4cb9af6-bcfa-11ef-8fe8-0242ac110002.czi
File Path (Cloud) https://s3.us-west-2.amazonaws.com/staging.files.allencell.org/c18/726/e93/ea3/804/468/34d/873/d6c/435/c6/aicsfiles-test-8c85cfec-bdc3-11ef-9089-0242ac110004.czi
File Path (Local) /mypath/aics/software/apps/staging/fss/data/7fb/6fb/b0f/6bd/7d5/1db/a4c/1e4/2a6/2a5/c7/aicsfiles-test-a4cb9af6-bcfa-11ef-8fe8-0242ac110002.czi
```
To:
```
File Path (Cloud) https://s3.us-west-2.amazonaws.com/staging.files.allencell.org/c18/726/e93/ea3/804/468/34d/873/d6c/435/c6/aicsfiles-test-8c85cfec-bdc3-11ef-9089-0242ac110004.czi
File Path (Vast) /mypath/aics/software/apps/staging/fss/data/7fb/6fb/b0f/6bd/7d5/1db/a4c/1e4/2a6/2a5/c7/aicsfiles-test-a4cb9af6-bcfa-11ef-8fe8-0242ac110002.czi
```

Closes #367 

# Changes
* Rename (Canonical) to (Cloud)
* Add the prefix `https://s3.us-west-2.amazonaws.com/` to cloud paths

# Testing
* Updated both `FileAnnotationList` tests.
* Manual test in staging:
Screenshot 2024-12-20 at 8 48 29 AM

Issue #293, opened 2024-10-22
Downloads from FMS staging data source fail with 404


tyler-foster:
This is an issue I see when trying to download files that were uploaded via the File Upload App to the staging environment. I'm not seeing this issue in production.

What I see:
<img width="1512" alt="image" src="https://github.com/user-attachments/assets/845f5d56-c4b5-44d5-b813-f092f480fe61">

This is with the "staging" data source selected.

Checking the staging FSS2 endpoint, we can see that this file record does exist and has valid paths:
<img width="1067" alt="image" src="https://github.com/user-attachments/assets/2407451e-2ea5-4722-b1db-4a5ddcd2c632">

The first obvious discrepancy is that the `cloudPath` in the Swagger response is `staging.files.allencell.org/2fc/093/2ce/382/2a2/4aa/148/be2/834/afc/aa/nas_test_file_1.txt`, while in BFF it's listed as `https://s3.us-west-2.amazonaws.com/production.files.allencell.org/allen/aics/software/apps/staging/fss/data/2fc/093/2ce/382/2a2/4aa/148/be2/834/afc/aa/nas_test_file_1.txt`

BrianWhitneyAI:
I think this is resolved

PR #377, opened 2024-12-20
376: Downloads use labkey URL when file is available on Vast

### LINKED ISSUE #376: Downloads should use labkey URL when file is available on Vast
Cloud-first FMS backend changes and existing updates for them (#370, #375) cause downloads to use the S3 URL. But for files available on the Vast, the download will be faster and use fewer resources if we download from the Vast via the labkey URL `http://aics.corp.alleninstitute.org/labkey/fmsfiles/image/allen/...`

pgarrison:
Cloud-first FMS backend changes and existing updates for them (https://github.com/AllenInstitute/biofile-finder/pull/370, https://github.com/AllenInstitute/biofile-finder/pull/375) cause downloads to use the S3 URL. But for files available on the Vast, the download will be faster and use fewer resources if we download from the Vast via the labkey URL `http://aics.corp.alleninstitute.org/labkey/fmsfiles/image/allen/...`

Closes #376 

# Testing
* I added an automated test that confirms the `downloadPath` has the expected format.
* I was unable to test this manually. This is because this PR depends on the latest changes to the ETL pipeline, which are only deployed to staging, but downloading files from staging is broken #293 (perhaps it has never worked).

# Notes
I believe we should merge this despite being unable to manually test it, because we know that the latest ETL pipeline changes will break the current `downloadPath` logic.

### DIFF
--- a/packages/core/entity/FileDetail/index.ts
+++ b/packages/core/entity/FileDetail/index.ts
@@ -148,10 +148,10 @@ export default class FileDetail {
     }
 
     public get downloadPath(): string {
-        // For AICS files we don't have permission to the bucket nor do we expect to have the /allen
-        // drive mounted on the client machine. So we use the NGINX server to serve the file.
-        if (this.path.startsWith("/allen")) {
-            return `http://aics.corp.alleninstitute.org/labkey/fmsfiles/image${this.path}`;
+        // For AICS files that are available on the Vast, users can use the cloud path, but the
+        // download will be faster and not incur egress fees if we download via the local network.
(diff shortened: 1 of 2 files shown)


pgarrison:
@BrianWhitneyAI To-do: review and merge!

SeanLeRoy (approved):

kmitcham (approved):

kmitcham:
Downloads from staging have been problematic; I suspect files uploaded to staging might work for download, but items uploaded to prod I would be surprised if they downloaded from staging

PR #380, opened 2025-01-02
Feature/backwards-etl-compatibility

### LINKED ISSUE #378: Cloud-first FMS branch should be backwards compatible
As currently designed, the cloud-first FMS features will cause breakages if used with the pre-cloud-first FMS ETL pipeline. But the cloud-first FMS ETL pipeline changes will cause breakages if used with the pre-cloud-first BFF. To enable a smooth migration, the BFF features should all be backwards compatible.

This is because the file_path is now the cloud path, which will break key features like opening a file in a viewer or downloading a file.

With this backwards compatibility, the following migration is possible.

1. Announce the new BFF version, and give people some time (a day?) to download it.
2. Deploy the ETL changes.

Without this backwards compatibility, we might have to accept that BFF will temporarily be broken for internal users.

BrianWhitneyAI:
### Description 
The purpose of this PR is to resolve #378 and make it so users of the old ETL can update their version prior to the release of the new ETL 

I have the download working and the cloud path is correctly mapped now but we do not show the File Path (local VAST) since we removed that as a top level annotation and now get it from the new etl as a normal annotation. If we wanted it to be visible we would have to add a custom top level annotation that is only used if the old ETL is in place and maps correctly. I think this is too much just for backwards compatibility. For users this would just mean that they do not see the annotation File Path (Local VAST) until the new ETL rolls out.

also if this is approved we need to open an issue to remove it for the next version.

### DIFF
--- a/packages/core/entity/FileDetail/index.ts
+++ b/packages/core/entity/FileDetail/index.ts
@@ -135,6 +135,11 @@ export default class FileDetail {
     }
 
     public get localPath(): string | null {
+        // REMOVE THIS (BACKWARDS COMPAT)
+        if (this.path.startsWith("/allen")) {
+            return this.path;
+        }
+
         const localPath = this.getFirstAnnotationValue("Local File Path");
         if (localPath === undefined) {
             return null;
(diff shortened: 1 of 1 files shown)


kmitcham (approved):

packages/core/entity/FileDetail/index.ts
```diff
@@ -135,6 +135,11 @@ export default class FileDetail {
     }
 
     public get localPath(): string | null {
+        // REMOVE THIS (BACKWARDS COMPAT)
+        if (this.path.startsWith("/allen")) {
```
hughes036, line 139: `this.path.startsWith("/allen")` means that it is a record from the old ETL?  

A function that describes that would be nice for readability:
`
function isOldEtlPath(path){
  retrun this.path.startsWith("/allen");
}
`

hughes036 (approved):

aswallace (approved):

PR #382, opened 2025-01-06
bugfix/svgicon-build-size

BrianWhitneyAI:
### Description
The thumbnail size is incorrectly defaulting to 15 pixels when app is built (but not in local). SvgIcon/index.tsx sets height as 15 by default if the prop is undefined, and in the dmg files and the deployed web builds it correctly sets this as the svg height, but locally it just leaves the height undefined. There are a few different solutions to this. This PR is meant to be a discussion hub for the best solution.
Screenshot 2025-01-06 at 9 47 24 AM

### Solution
1) remove the defaults (changes on this PR)
2) refine thumbnail parameters like so
image
3) ensure build matches local (not sure how to do this)

### DIFF
--- a/packages/core/components/FileThumbnail/index.tsx
+++ b/packages/core/components/FileThumbnail/index.tsx
@@ -45,10 +45,10 @@ export default function FileThumbnail(props: Props) {
     if (!props.uri) {
         return (
             <SvgIcon
-                height={props.height}
+                height={props?.height || 150}
                 pathData={NO_IMAGE_ICON_PATH_DATA}
                 viewBox="0,1,22,22"
-                width={props.width}
+                width={props?.width || 150}
                 className={classNames(styles.noThumbnail, {
                     [styles.noThumbnailSelected]: props?.selected,
                 })}
--- a/packages/core/components/SvgIcon/index.tsx
+++ b/packages/core/components/SvgIcon/index.tsx
@@ -1,5 +1,5 @@
 import classNames from "classnames";
-import { castArray, defaults } from "lodash";
+import { castArray } from "lodash";
 import * as React from "react";
 
 import styles from "./SvgIcon.module.css";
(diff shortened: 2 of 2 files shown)


aswallace:
I'm fine with both solutions (removing defaults and/or setting the height). 
I do think that either way we should still set the height for that specific SvgIcon to some fixed value in the FileDetails panel, e.g., 150px. If we leave height as undefined (option 1) or set to 100% (I know I initially suggested using a percent for option 2, I retract it), the icon keeps getting bigger as you expand the side panel, which can be pretty annoying.

> ensure build matches local (not sure how to do this)

I'm not quite sure what you mean by this, but if it's just saying that we should test that the .dmg files match local builds then we should be able to continue using the `build-executable` script

kmitcham (approved):

BrianWhitneyAI:
> I'm fine with both solutions (removing defaults and/or setting the height). I do think that either way we should still set the height for that specific SvgIcon to some fixed value in the FileDetails panel, e.g., 150px. If we leave height as undefined (option 1) or set to 100% (I know I initially suggested using a percent for option 2, I retract it), the icon keeps getting bigger as you expand the side panel, which can be pretty annoying.
> 
> > ensure build matches local (not sure how to do this)
> 
> I'm not quite sure what you mean by this, but if it's just saying that we should test that the .dmg files match local builds then we should be able to continue using the `build-executable` script

Its pretty abstract but what I was saying was come up with a way to make what we see in build exactly match what we see locally and then go case by case to fix (since we see the visual discrepancy early in local). I do not have an idea of what this would need.

pgarrison:
> Its pretty abstract but what I was saying was come up with a way to make what we see in build exactly match what we see locally and then go case by case to fix (since we see the visual discrepancy early in local). I do not have an idea of what this would need.

If you're talking about an automated solution (e.g. snapshot testing), that might be more complexity than this project is ready for. But some manual QA testing to find more bugs like this is always good.

aswallace:
> Its pretty abstract but what I was saying was come up with a way to make what we see in build exactly match what we see locally and then go case by case to fix (since we see the visual discrepancy early in local). I do not have an idea of what this would need.

I'm pretty sure this particular discrepancy is a lodash versioning issue. If we don't want to deep-dive into dependency differences in the build scripts and are only trying to address this one bug for now, removing the call to `defaults` should make them behave the same way. An alternative would be to continue using `defaults` but fix the call, which based on [this issue](https://github.com/lodash/lodash/issues/565) and some local testing should just be to add an empty obj as the first arg like
```   
const { className, height, onClick, pathAttrs, pathData, viewBox, width } = defaults(
        {},
        props,
        DEFAULT_PROPS
    );
```

Issue #173, opened 2024-07-15
Unable to build universal deploy for MacOS 🐛


aswallace:
### Initial notes from Sean: 

When `Arm64` mac runners became the default for GitHub it seems it broke our desktop deploys to MacOS by having only `Arm64` arch compatible executables get output as part of the release process. This can be fixed by changing the build options in the `package.json` to target `x86` as a separate deploy or by building a `universal` version. However, this is where the trouble really begins. It seems like for some reason the `universal` and `x86` builds that build on GitHub don't actually produce an `x86` compatible executable, instead I get this error:

<img width="984" alt="image" src="https://github.com/user-attachments/assets/5a0d18bc-2074-4c34-a0d5-4c18ce712285" />

This seems to be an error many others have gotten as well and the routes and various fixes they perform are varied. What is more annoying is it seems the File Upload App does exactly what we do [on my working branch](https://github.com/AllenInstitute/aics-fms-file-explorer-app/compare/main...bugfix/make-macos-build-universal?expand=1) but without the error ([see here](https://github.com/aics-int/aics-file-upload-app/blob/main/package.json#L20-L25). 

Related Issues
* https://github.com/electron/universal/issues/41
* https://github.com/electron-userland/electron-builder/issues/7512
* https://github.com/electron-userland/electron-builder/pull/6983
* https://github.com/electron-userland/electron-builder/issues/6433

Things I tried to so far:
* Upgrading electron & electron-builder
* Upgrading electron-builder to a middle ground version like 22
* Setting `mergeAsar` to false
* Setting the node modules included to exclude certain extensions
(various combos of above, not to say any of these should necessarily be ruled out since it may require the right combo)

### Update: Notes from Anya
I eventually narrowed down the issue to DuckDB. When I remove duckdb as a dependency completely (and noop all duckdb functionality) the universal builds work for both arch. I *think* they've fixed the mac incompatibility issue in 1.0.0 (we're using 0.8.x) BUT when I try to upgrade to 1.0.0, I get a whole bunch of nasty mapbox/node-pre-gyp dependency & webpack errors. Unfortunately, node-pre-gyp is no longer maintained, so I haven't had much luck in finding solutions
Related discussion: 
* https://github.com/duckdb/duckdb/discussions/7067
* https://github.com/mapbox/node-pre-gyp/issues/661
* This is for an older version but vaguely related? https://github.com/duckdb/duckdb/issues/5605
[...]

PR #383, opened 2025-01-08
Update build process for MacOS desktop app

aswallace:
The current build script creates dmg files for the desktop app that, when downloaded from GitHub/shared via Slack, show the following error for M chip Macs: 
```
"BioFile Finder" is damaged and can't be opened. You should eject the disk image. This item is on the disk image "BioFile Finder-8.0.0-arm64.dmg
```

This PR updates `desktop/package.json` to what was used to get the v8.0.0 build working for arm64 (also currently works for x86, though only when run on a computer with an Intel chip).
Also adds info to the dev docs about how we currently build & release for MacOS.

For more details on why we have to do separate builds for Mac, see #173

### DIFF
--- a/dev-docs/04-versioning-and-deployment.md
+++ b/dev-docs/04-versioning-and-deployment.md
@@ -45,8 +45,13 @@ builds of `packages/desktop`, prepare a draft Github release, and upload the bui
 6) [Update the GitHub release](https://github.com/AllenInstitute/biofile-finder/releases) once the Github action in Step 4 is finished, manually edit the Github release which was drafted as part of Step 4. Format its release name with the date (consistent with other release names), add a description of the changes, and optionally
 mark whether the release is "pre-release." If it is marked as "pre-release," it will not be accessible for download through the
 Github pages site.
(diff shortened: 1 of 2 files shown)


BrianWhitneyAI (approved):

dev-docs/04-versioning-and-deployment.md
```diff
@@ -45,8 +45,13 @@ builds of `packages/desktop`, prepare a draft Github release, and upload the bui
 6) [Update the GitHub release](https://github.com/AllenInstitute/biofile-finder/releases) once the Github action in Step 4 is finished, manually edit the Github release which was drafted as part of Step 4. Format its release name with the date (consistent with other release names), add a description of the changes, and optionally
 mark whether the release is "pre-release." If it is marked as "pre-release," it will not be accessible for download through the
 Github pages site.
+7) **Temporary workaround for MacOS**: GitHub Actions currently builds .dmg files that don't work for x86 processors and that are interpreted as "damaged" by new MacOS versions. To get around this, follow the instructions below to [manually build executables](#manually-building-an-executable). You will need to do this separately for both arm64 and x86, and each build process **must be done on a machine that uses the correct processor** (e.g., on a computer with an Intel chip for the x86 build).
```
kmitcham, line 48: I like to include a date in messages like **Temporary workaround for MacOS** so when future people come around it is easier to track down info about the issue.  Obviously the date is also in the commit history, but making it easier to spot is easier.  Also, it might help guilt motivation towards fixing it.

kmitcham (approved):

PR #388, opened 2025-01-09
356: Copy to VAST status indicator

### LINKED ISSUE #356: Copy to Vast: status indicator
# User story
Scientists 1 and 2 are working on a project that needs to use file A on local storage, but File A is only on the cloud. Scientist 1 uses BFF to copy the file to the Vast. The file is 4TB so it isn't immediately available. Scientist 2 goes to BFF to do the same thing: they would like to know if the download is already in progress.

# Design
In “canonical file path field” write “Copying to Vast in progress…” (italics) for in-progress downloads

# Implementation details
* Cloud only: `shouldBeInLocal` is False
* Download in progress: `shouldBeInLocal` is True and eviction date is null
* Download complete: `shouldBeInLocal` is True and eviction date is non-null

pgarrison:
# Purpose
Scientists 1 and 2 are working on a project that needs to use file A on local storage, but File A is only on the cloud. Scientist 1 uses BFF to copy the file to the Vast. The file is 4TB so it isn't immediately available. Scientist 2 goes to BFF to do the same thing: they would like to know if the download is already in progress.

Resolves #356 

# Testing
* New unit test
* Manual test in staging
Screenshot 2025-01-09 at 12 58 57 PM

### DIFF
--- a/packages/core/components/FileDetails/FileAnnotationList.tsx
+++ b/packages/core/components/FileDetails/FileAnnotationList.tsx
@@ -73,15 +73,12 @@ export default function FileAnnotationList(props: FileAnnotationListProps) {
             }
 
             let annotationValue = annotation.extractFromFile(fileDetails);
-            if (annotationValue === Annotation.MISSING_VALUE) {
-                // Nothing to show for this annotation -- skip
-                return accum;
-            }
 
             if (annotation.name === AnnotationName.LOCAL_FILE_PATH) {
-                if (localPath === null) {
-                    // localPath hasn't loaded yet, but it should eventually because there is an
-                    // annotation named AnnotationName.LOCAL_FILE_PATH
+                if (fileDetails && fileDetails.downloadInProgress) {
+                    annotationValue = "Copying to VAST in progress…";
+                } else if (localPath === null) {
+                    // localPath hasn't loaded yet or there is no local path annotation
                     return accum;
                 } else {
                     // Use the user's /allen mount point, if known
@@ -92,6 +89,11 @@ export default function FileAnnotationList(props: FileAnnotationListProps) {
             if (annotation.name === AnnotationName.FILE_PATH) {
                 // Display the full http://... URL
                 annotationValue = fileDetails.cloudPath;
+            }
+
+            if (annotationValue === Annotation.MISSING_VALUE) {
+                // Nothing to show for this annotation -- skip
+                return accum;
             }
 
             return [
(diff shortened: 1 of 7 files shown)


lynwilhelm:
With the italics added, this LGTM!

packages/core/components/FileDetails/FileAnnotationList.tsx
```diff
@@ -94,6 +91,11 @@ export default function FileAnnotationList(props: FileAnnotationListProps) {
                 annotationValue = fileDetails.cloudPath;
             }
 
+            if (annotationValue === Annotation.MISSING_VALUE) {
```
BrianWhitneyAI, line 94: What is the purpose of this?

BrianWhitneyAI (approved):

packages/core/components/FileDetails/FileAnnotationList.tsx
pgarrison: This conditional was moved from line 74 because `annotation.extractFromFile(fileDetails);` returns MISSING_VALUE for the local file path annotation when the download is in progress (because there is not local file path yet).

packages/core/components/FileDetails/FileAnnotationList.tsx
```diff
@@ -94,13 +93,19 @@ export default function FileAnnotationList(props: FileAnnotationListProps) {
...
+            }
+
             return [
                 ...accum,
                 <FileAnnotationRow
                     key={annotation.displayName}
                     className={styles.row}
                     name={annotation.displayName}
                     value={annotationValue}
+                    fmsStateIndicator={fmsStateIndicator}
```
aswallace, line 108: If this is a boolean, it should be possible to pass as just `fmsStateIndicator` without `={fmsStateIndicator}`
It also might be nice to generalize this in case we plan to re-use this styling, e.g. `shouldItalicize`

aswallace (approved):

packages/core/components/FileDetails/FileAnnotationList.tsx
pgarrison: > If this is a boolean, it should be possible to pass as just fmsStateIndicator without ={fmsStateIndicator}
The following are equivalent:
```jsx
<FileAnnotationRow fmsStateIndicator />
<FileAnnotationRow fmsStateIndicator={true} />
```
Since `fmsStateIndicator` is sometimes (usually) false, this suggestion wouldn't work without additional changes.

> It also might be nice to generalize this in case we plan to re-use this styling, e.g. shouldItalicize

I was trying to avoid a name like `shouldItalicize` that focuses on the presentation so that if we decide it should be bold or green or something instead, we don't need to also rename the variable. That said, I don't have any strong preference here.

PR #389, opened 2025-01-10
feature/derive_local_path

### LINKED ISSUE #390: Feature Request: Derive local path from fms ID and show conditionally.
When working with the ETL it became apparent that having an annotation with a unique value for each file puts a lot of stress on the annotation collection. We recently added `local file path` and `Cache Eviction Date` to this collection which caused a breakage of our annotation collection rebuild. 

The solution we came up with is to remove the `local path annotation` and derive it from fms ID and show conditionally based on the presence of `Cache Eviction Date`. Additionally `Cache Eviction Date` will move to a bucketed date time, decreasing the unique values for files and making the annotation more palatable.

BrianWhitneyAI:
### Description 
The purpose of this PR is to resolve #390. When working with the ETL it became apparent that having an annotation with a unique value for each file puts a lot of stress on the annotation collection. We recently added `local file path` and `Cache Eviction Date` to this collection which caused a breakage of our annotation collection rebuild. 

The solution we came up with is to remove the `local path` annotation and derive it from fms ID and show conditionally based on the presence of `Cache Eviction Date`. Additionally `Cache Eviction Date` will move to a bucketed date time, decreasing the unique values for files and making the annotation more palatable.

This PR also removes the previous backwards compatibility.


BrianWhitneyAI:
Philip brought up that we could derive from the Cloud path as well which would be much simpler. I opted against this idea since the cloud path is already derived from fms-ID. Open to a discussion here!

packages/core/entity/FileDetail/index.ts
```diff
@@ -85,12 +87,69 @@ export default class FileDetail {
...
+    }
+
+    private static convertFMSIDToLocalPath(guid: string): string[] {
+        if (!guid) {
+            throw new Error("GUID cannot be null or undefined");
+        }
+
+        const paths: string[] = [];
+
+        while (guid.length > 2) {
```
SeanLeRoy, line 125: This is great, but FWIW lodash can do this whole method in one line for ya:
```

    // Splits the fileId into an array where is character is its own element then chunks them into groups of 3 with the
    // remainder (between 1-3 characters) becoming the final group
    chunk(this.fileId.split(""), 3);
```

packages/core/entity/FileDetail/index.ts
```diff
@@ -85,12 +87,69 @@ export default class FileDetail {
...
+        const idx = safeString.length - numChars;
+        return idx >= 0 ? safeString.slice(idx) : safeString;
+    }
+
     private static convertAicsS3PathToHttpUrl(path: string): string {
         return `${AICS_FMS_S3_URL_PREFIX}${path}`;
     }
 
-    constructor(fileDetail: FmsFile, uniqueId?: string) {
+    constructor(fileDetail: FmsFile, env: Environment, uniqueId?: string) {
```
SeanLeRoy, line 150: I think it'll be cleaner long-term to keep this out of the properties and instead pass it to a getter method `getLocalPath()` - this would replace the property `localPath` (FWIW `localPath` seen below is a property because of the `get` before its method name

SeanLeRoy (changes requested):

packages/core/components/FileList/test/LazilyRenderedRow.test.tsx
```diff
@@ -24,20 +25,23 @@ describe("<LazilyRenderedRow />", () => {
...
-                    file_size: 1,
-                    thumbnail: "",
-                    uploaded: new Date().toISOString(),
-                });
+                return new FileDetail(
+                    {
+                        annotations: [
+                            {
+                                name: "someDateAnnotation",
+                                values: ["2019-05-17T07:43:55.205Z"],
```
kmitcham, line 33: Is supposed to be a date, or a datetime?  I thought I saw something about 'bucketing' these values, and it seems like the precision of a datetime is overkill, and a date would be fine.
I suppose the 'date' bucket might put a load on the cache clearing.  And that this might not be the best place to mention it.

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: oh yeah I definitely just ported it over from FSS and didnt consider.

packages/core/components/FileList/test/LazilyRenderedThumbnail.test.tsx
```diff
@@ -10,41 +10,51 @@ import { initialState } from "../../../state";
...
             if (index === 25) {
-                return new FileDetail({
-                    annotations: [],
-                    file_id: "abc12325",
-                    file_name: "my_image25.czi",
-                    file_path: "some/path/to/my_image25.czi",
-                    file_size: 1,
-                    uploaded: new Date().toISOString(),
-                });
+                return new FileDetail(
```
hughes036, line 47: just formatting?

packages/core/constants/index.ts
```diff
@@ -31,6 +31,12 @@ export const TOP_LEVEL_FILE_ANNOTATIONS = [
         description: "Path to file in the cloud.",
         type: AnnotationType.STRING,
     }),
+    new Annotation({
```
hughes036, line 34: Looks like this is the pattern for dealing with file properties?

packages/core/entity/FileDetail/index.ts
```diff
@@ -85,12 +87,69 @@ export default class FileDetail {
...
+        // Define prefixes based on the environment
+        const prefixes: Record<Environment, string> = {
+            LOCALHOST: "/tmp/fss/local",
+            PRODUCTION: "/allen/programs/allencell/data/proj0",
+            STAGING: "/allen/aics/software/apps/staging/fss/data",
+            TEST: "test",
+        };
+
+        // Get the prefix for the given environment
+        const prefix = prefixes[env];
```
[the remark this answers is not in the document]
hughes036: `envToPrefix` ?

packages/core/entity/FileDetail/index.ts
```diff
@@ -85,12 +87,69 @@ export default class FileDetail {
         return FileDetail.convertAicsS3PathToHttpUrl(`${AICS_FMS_S3_BUCKET}${pathWithoutDrive}`);
     }
 
+    private static generateFilePath(env: Environment, fileName: string, fmsId: string): string {
```
hughes036, line 90: Does generateFilePath have test cov?

packages/core/entity/FileDetail/index.ts
hughes036: less custom code to test FTW

packages/core/entity/FileDetail/index.ts
```diff
@@ -135,34 +194,18 @@ export default class FileDetail {
...
-        if (localPath === undefined) {
-            return null;
-        }
-        return localPath as string;
+        return null;
     }
 
     public get cloudPath(): string {
-        // REMOVE THIS (BACKWARDS COMPAT)
-        if (this.path.startsWith("/allen")) {
```
hughes036, line 152: We still need backward compatibility - should this change (and the block below) use the new heuristic `this.getAnnotation("Cache Eviction Date")`?

packages/core/services/HttpServiceBase/index.ts
```diff
@@ -167,6 +167,19 @@ export default class HttpServiceBase {
         return new RestServiceResponse(cachedResponseData);
     }
 
+    public getEnvironmentFromUrl(): Environment {
```
hughes036, line 170: has unit test cov?

hughes036 (approved):
See questions about backward compatibility.  Lgtm otherwise.

packages/core/entity/FileDetail/index.ts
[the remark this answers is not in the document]
BrianWhitneyAI: Resolved by deriving from cloud path in [3bee712](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/3bee712c637ad2bf7571bed19b960fe11de85598)

packages/core/entity/FileDetail/index.ts
```diff
@@ -85,12 +87,69 @@ export default class FileDetail {
         return FileDetail.convertAicsS3PathToHttpUrl(`${AICS_FMS_S3_BUCKET}${pathWithoutDrive}`);
     }
 
+    private static generateFilePath(env: Environment, fileName: string, fmsId: string): string {
+        if (!fileName || !fmsId) {
```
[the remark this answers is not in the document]
BrianWhitneyAI: Resolved by removing in [3bee712](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/3bee712c637ad2bf7571bed19b960fe11de85598)

packages/core/entity/FileDetail/index.ts
```diff
@@ -85,12 +87,69 @@ export default class FileDetail {
...
+            throw new Error("File name and FMS ID must be provided");
+        }
+
+        // Define prefixes based on the environment
+        const prefixes: Record<Environment, string> = {
```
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in [3bee712](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/3bee712c637ad2bf7571bed19b960fe11de85598)

packages/core/entity/FileDetail/index.ts
```diff
@@ -85,12 +87,69 @@ export default class FileDetail {
...
+        // Generate path segments from FMS ID
+        const pathSegments = this.convertFMSIDToLocalPath(fmsId);
+
+        // Construct the full file path
+        const filePath = `${prefix}/${pathSegments.join("/")}/${fileName}`;
+
+        return filePath;
+    }
+
+    private static convertFMSIDToLocalPath(guid: string): string[] {
```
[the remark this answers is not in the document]
BrianWhitneyAI: Resolved by removing in [3bee712](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/3bee712c637ad2bf7571bed19b960fe11de85598)

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: Resolved by deriving from cloud path in [3bee712](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/3bee712c637ad2bf7571bed19b960fe11de85598)

packages/core/entity/FileDetail/index.ts
```diff
@@ -85,12 +87,69 @@ export default class FileDetail {
...
+        if (!guid) {
```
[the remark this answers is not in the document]
BrianWhitneyAI: resolved by removing in [3bee712](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/3bee712c637ad2bf7571bed19b960fe11de85598)

packages/core/components/FileList/test/LazilyRenderedThumbnail.test.tsx
BrianWhitneyAI: Yep!, adding the Environment adds an indent that GitHub doesn't like.

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: I made getLocalPath as a function in [3bee712](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/3bee712c637ad2bf7571bed19b960fe11de85598). However I left get localPath where it calls getLocaPath() since it has a conditional of Cache eviction date. Lmk what you think about this, open to just removing it and moving the conditional into getLocalPath()

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: Not any more.... removed in [3bee712](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/3bee712c637ad2bf7571bed19b960fe11de85598)

packages/core/services/HttpServiceBase/index.ts
BrianWhitneyAI: removed in [3bee712](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/3bee712c637ad2bf7571bed19b960fe11de85598)

packages/core/constants/index.ts
BrianWhitneyAI: Not sure what the comment is here but my interpretation is this format allows us to create our own annotations in BFF which we later propagate with the file properties.

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: The backwards compatibility is to handle file.path being '/allen' or the updated version where file.path = cloud path. I thought that we were fully past that after the new file rebuild, hence trying to remove it. however seems like there's still a possibility of file.path being /allen as talked about today in #cloud-first-fms so I am in favor of adding it back

BrianWhitneyAI: In regards to `Cache Eviction Date` I don't think that we should use this yet since it has not been backfilled and will only show up on new annotations. We would not reach a state where the file path property was old where we had an eviction date I don't think?

kmitcham:
> Philip brought up that we could derive from the Cloud path as well which would be much simpler. I opted against this idea since the cloud path is already derived from fms-ID. Open to a discussion here!

I don't think we will change how we construct the paths (chunks of the reversed id) but if we did, having that path calculated in fewer places would be better.  As I understand it, that suggests deriving the local path from the cloud path might make the path system easier to change  later.  Of course, if we change the path system ONLY for local storage, then having the path calculated in BFF would be better.

packages/core/entity/FileDetail/index.ts
SeanLeRoy: My main issue with this isn't the usage of `localPath` as a property but that `env` is being passed in as a property to this class when it could just be a parameter to `getLocalPath`

packages/core/entity/FileDetail/index.ts
```diff
@@ -126,56 +144,74 @@ export default class FileDetail {
...
+        const localPrefix = HOST_PREFIXES[this.env];
+        const cloudPrefix = `${AICS_FMS_S3_URL_PREFIX}${AICS_FMS_S3_BUCKETS[this.env]}`;
+
+        if (this.path.startsWith(cloudPrefix)) {
+            const relativePath = this.path.replace(`${cloudPrefix}`, "");
+            return `${localPrefix}${relativePath}`;
         }
 
-        // AICS FMS files' paths are cloud paths
         return this.path;
```
SeanLeRoy, line 192: Is this `null` or the cloud path at this point? If `null` can you return `null` here explicitly & if it is the cloud path can you keep the comment above?

packages/core/entity/FileDetail/index.ts
```diff
@@ -126,56 +144,74 @@ export default class FileDetail {
...
         }
+        ////
 
+        const localPath = this.getLocalPath();
         // For AICS files that are available on the Vast, users can use the cloud path, but the
         // download will be faster and not incur egress fees if we download via the local network.
-        if (this.localPath && this.localPath.startsWith("/allen")) {
-            return `http://aics.corp.alleninstitute.org/labkey/fmsfiles/image${this.localPath}`;
+        if (localPath && localPath.startsWith("/allen") && this.env === Environment.PRODUCTION) {
+            return `http://aics.corp.alleninstitute.org/labkey/fmsfiles/image${localPath}`;
```
SeanLeRoy, line 206: Perhaps if you are introducing env dependent variables to this class then there should be a case for `this.env === Environment.STAGING` where the return is:
`return `http://stg-aics.corp.alleninstitute.org/labkey/fmsfiles/image${localPath}`;`

BrianWhitneyAI: We dont have labkey download for staging files I believe so this function default returns the cloud path for the staging file which can be downloaded

SeanLeRoy (approved):

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: This is for backwards compatibility. Right now its possible this.path is '/allen' in this case we want to return it.

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: for record. Consensus was env is used elsewhere and should remain

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: Resolved in [03c770b](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/03c770bc20447a5d27bbb7dde49360e254854308)

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: oh Im wrong. We already check in localPath for this. setting to null.

BrianWhitneyAI: resilved in [41c48db](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/41c48db32edf9182ac477cd73ffb16de9f0a64f4)

packages/core/entity/FileDetail/index.ts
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in [41c48db](https://github.com/AllenInstitute/biofile-finder/pull/389/commits/41c48db32edf9182ac477cd73ffb16de9f0a64f4)

PR #402, opened 2025-01-30
bugfix/missing-icons

### LINKED ISSUE #399: Icons aren't loading
Both safari and chrome on a MacBook Pro: 

Image

Image

BrianWhitneyAI:
### Description 

The icons for bff are missing and requesting them raises a 404. This PR updates the version so that the request goes to the right address. 

Resolves #399

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
       "dependencies": {
         "@aics/frontend-insights": "0.2.x",
         "@aics/redux-utils": "0.6.0",
-        "@fluentui/react": "8.67.x",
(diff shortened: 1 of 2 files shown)


SeanDuHare (approved):
Seems like you could do “^8.67” to avoid the version being able to drop lower than it was prior to this PR

BrianWhitneyAI:
> Member

Anya mentioned as well that we might want an exact release for stability sake. thoughts @SeanLeRoy?

aswallace (approved):
verified fix works for me locally

PR #404, opened 2025-02-04
add custom build workflow

BrianWhitneyAI:
### Description 

This PR adds a workflow to build a dist for a specific branch for a particular OS. This is useful if you have a bug that is os dependent but only have a dev setup of a specific OS.

### DIFF
--- a/.github/workflows/maunal-build.yml
+++ b/.github/workflows/maunal-build.yml
@@ -1,0 +1,51 @@
+name: manual-build
+
+ on:
+   workflow_dispatch:
+     inputs:
+       branch:
+         description: "Branch to build"
+         required: true
+         default: "main"
+       os:
+         description: "Operating System"
+         required: true
+         type: choice
+         options:
+           - ubuntu-latest
+           - windows-latest
+           - macos-latest
+
+ jobs:
+   build:
+     runs-on: ${{ github.event.inputs.os }}
+
+     steps:
+       - name: Check out Git repository
+         uses: actions/checkout@v4
+         with:
+           ref: ${{ github.event.inputs.branch }}
+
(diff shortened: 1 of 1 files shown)


BrianWhitneyAI:
I originally just appended the save artifacts portion of this workflow to our existing integration. This made the workflow pretty slow and store about 1 GB per run. I think this is too much to do per commit to a branch. This is a compromise to generate the artifacts for a specific situation instead.

SeanLeRoy (approved):

.github/workflows/maunal-build.yml
```diff
@@ -0,0 +1,51 @@
...
+
+ jobs:
+   build:
+     runs-on: ${{ github.event.inputs.os }}
+
+     steps:
+       - name: Check out Git repository
+         uses: actions/checkout@v4
+         with:
+           ref: ${{ github.event.inputs.branch }}
```
pgarrison, line 27:
```suggestion
     inputs:
       os:
         description: "Operating System"
         required: true
         type: choice
         options:
           - ubuntu-latest
           - windows-latest
           - macos-latest

 jobs:
   build:
     runs-on: ${{ github.event.inputs.os }}

     steps:
       - name: Check out Git repository
         uses: actions/checkout@v4
```
The `workflow_dispatch` trigger comes with a branch selection dropdown by default. See [this action](https://github.com/AllenInstitute/biofile-finder/actions/workflows/aws-deployment.yml) for example

pgarrison (approved):

aswallace (approved):
Do we know how this will work with the two macos chip types? & if it will only work for m3 (since historically GH causes us issues with x86), unsure if that needs to be noted anywhere

PR #405, opened 2025-02-04
bugfix/ZEN-wrong-path

### LINKED ISSUE #401: BFF fails to open file in Zen
Sara was working with BFF, and tried to send a file to Zen.  Zen started up, but did not view the file.
She was able to open the file from Zen by manually selecting it based on the displayed BFF file path on the vast.
Slack conversation here: 
https://allencellscience.slack.com/archives/C021CEY96LD/p1738092593969209

BrianWhitneyAI:
### Description 
The purpose of this PR is to resolve #401 and allow scientists to open local files ( VAST ) in ZEN. When we switched to cloud path being our primary path we then pass this path to external apps that try to "open it". This means that apps that need a local path like ZEN cannot open these paths.

### Development Decisions

In order to complete this PR we had to make a few decisions.

1) Desktop apps will default to using local path. This means that any app that users add will automatically use local path.
2) Disable desktop applications for cloud only files.

In the future if we want specific apps to use a cloud path then we need to add them manually to the repo, otherwise it'll default to opening with the local path.

### DIFF
--- a/packages/core/hooks/useOpenWithMenuItems/index.tsx
+++ b/packages/core/hooks/useOpenWithMenuItems/index.tsx
@@ -222,7 +222,9 @@ export default (fileDetails?: FileDetail, filters?: FileFilter[]): IContextualMe
                 key: `open-with-${name}`,
                 text: name,
                 title: `Open files with ${name}`,
-                disabled: (!filters && !fileDetails) || app.filePath.toLowerCase().includes("zen"),
+                disabled:
+                    (!filters && !fileDetails) ||
+                    (app.filePath.toLowerCase().includes("zen") && !fileDetails?.localPath),
                 onClick() {
                     if (filters) {
                         dispatch(interaction.actions.openWith(app, filters));
--- a/packages/core/hooks/useOpenWithMenuItems/index.tsx
+++ b/packages/core/hooks/useOpenWithMenuItems/index.tsx
@@ -222,7 +222,7 @@ export default (fileDetails?: FileDetail, filters?: FileFilter[]): IContextualMe
                 key: `open-with-${name}`,
                 text: name,
                 title: `Open files with ${name}`,
-                disabled: (!filters && !fileDetails) || !fileDetails?.localPath,
+                disabled: (!filters && !fileDetails) || app.filePath.toLowerCase().includes("zen"),
                 onClick() {
                     if (filters) {
                         dispatch(interaction.actions.openWith(app, filters));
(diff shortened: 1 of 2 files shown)


SeanLeRoy:
Is this a draft?

BrianWhitneyAI:
Ok I tested this through Remote Desktop. Files that are cloud only grey out ZEN as an option and files that are in local cache open correctly in ZEN

packages/core/state/interaction/logics.ts
```diff
@@ -1,4 +1,5 @@
 import { isEmpty, sumBy, throttle, uniq, uniqueId } from "lodash";
+import { isAbsolute } from "path";
```
SeanLeRoy, line 2: isn't path a `node` module aka not available for web deploys?

packages/core/state/interaction/logics.ts
```diff
@@ -489,8 +490,14 @@ const openWithLogic = createLogic({
         } else {
             filesToOpen = await fileSelection.fetchAllDetails();
         }
+
+        // Determine if exePath is local
+        const isLocalExecutable = isAbsolute(exePath);
```
SeanLeRoy, line 495: I suppose it is likely we will always store the path as absolute for local executables but it does seem potentially buggy with the way certain systems mount apps - I'm pretty sure `executionEnvService` is only used for local apps anyway right? Couldn't we just assume they are always local?

packages/core/hooks/useOpenWithMenuItems/index.tsx
```diff
@@ -222,7 +222,7 @@ export default (fileDetails?: FileDetail, filters?: FileFilter[]): IContextualMe
                 key: `open-with-${name}`,
                 text: name,
                 title: `Open files with ${name}`,
-                disabled: !filters && !fileDetails,
+                disabled: (!filters && !fileDetails) || !fileDetails?.localPath,
```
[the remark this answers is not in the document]
kmitcham: Is the user experience better with mysterious opening failures or reduced options, which seem to be the choices?  Or do we want to undertake tracking which apps are cloud ready, and showing the appropriate items for the situation?

packages/core/state/interaction/logics.ts
BrianWhitneyAI: ah rats I always forget about this. (And didn't test it on web since I was using a remote workstation)

kmitcham (approved):

SeanLeRoy (approved):

PR #406, opened 2025-02-05
Add logic for creating new annotations

### LINKED ISSUE #371: Logic for adding new annotations/metadata
FMS: 
- Call the POST on the appropriate MMS endpoint to create a new annotation

Non FMS:
- Add logic to add new columns to duckDB?

_Originally posted by @aswallace in https://github.com/AllenInstitute/biofile-finder/issues/96#issuecomment-2359485655_

aswallace:
This adds the redux & endpoint logic for creating new annotations (via MMS for FMS and in DuckDB for non-FMS sources), with some supporting UI modifications.

There are a couple tricky logistical choices here that I am NOT attached to and very open to changing.

### 1. Reporting status of network call
We need to wait to send the file edit request until after the annotation creation request is successfully completed. We can't return the results of dispatches directly (e.g., can't pass the response from createNewAnnotationLogic directly to NewAnnotationPathway) since we aren't using thunk (and can't with our current dependencies)
Instead, we report the status from dispatch to our store. The component watches the store property, and then sends the file edit request when it sees success.

### 2. Making the new annotation available for file editing
Our editing logic checks if annotations exist before allowing the edit to go through, but it checks FES, not MMS. FES will not register/report a new MMS annotation until a file exists that uses that annotation.
I'm currently getting around this by temporarily adding the results of the annotation post to our store. This is messy for a few reasons, but one is that the annotation response from MMS looks * slightly * different than the response from FES and needs to be converted.

### 3. Annotation types
FMS doesn't have an endpoint for requesting the full list of annotation types & their IDs, so currently using a hardcoded map.

## Screenshots of UI changes
Added a description text input
Screenshot 2025-02-06 at 11 24 15 AM

Messages to report status of annotation creation
image
image

resolves #371

### DIFF
--- a/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
+++ b/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
@@ -47,6 +47,8 @@ interface AnnotationStatus {
  */
 export default function NewAnnotationPathway(props: NewAnnotationProps) {
     const dispatch = useDispatch();
+    // Destructure to prevent unnecessary useEffect triggers
+    const { onDismiss, hasUnsavedChanges } = props;
 
     const [step, setStep] = React.useState<EditStep>(EditStep.CREATE_FIELD);
     const [newValues, setNewValues] = React.useState<AnnotationValue | undefined>();
@@ -87,8 +89,8 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
                                 message: `Failed to create annotation: ${e}`,
                             });
                         } finally {
-                            props?.hasUnsavedChanges(false);
-                            props.onDismiss();
+                            hasUnsavedChanges(false);
+                            onDismiss();
                         }
                     }
                 default:
@@ -96,7 +98,7 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
             }
         };
         checkForStatusUpdates();
-    }, [statuses, annotationCreationStatus, dispatch]);
+    }, [annotationCreationStatus, dispatch, hasUnsavedChanges, newFieldName, newValues, onDismiss]);
 
     const addDropdownChip = (evt: React.FormEvent) => {
         evt.preventDefault();
--- a/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
+++ b/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
@@ -100,6 +100,19 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
         checkForStatusUpdates();
     }, [annotationCreationStatus, dispatch, hasUnsavedChanges, newFieldName, newValues, onDismiss]);
 
+    const onChangeAlphanumericField = (
+        e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+        newValue: string | undefined
+    ) => {
+        const regex = /^[\w\-\s]+$/g;
+        // Restricts character entry to alphanumeric
+        if (newValue && !regex.test(newValue)) {
+            e.preventDefault();
+        } else {
+            setNewFieldName(newValue || "");
+        }
+    };
+
     const addDropdownChip = (evt: React.FormEvent) => {
         evt.preventDefault();
         if (
@@ -154,7 +167,7 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
                 required
                 label="New metadata field name"
                 className={styles.textField}
-                onChange={(_, newValue) => setNewFieldName(newValue || "")}
+                onChange={(ev, newValue) => onChangeAlphanumericField(ev, newValue)}
                 placeholder="Add a new field name..."
                 value={newFieldName}
             />
--- a/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
+++ b/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
@@ -54,7 +54,9 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
     const [newValues, setNewValues] = React.useState<AnnotationValue | undefined>();
     const [newFieldName, setNewFieldName] = React.useState<string>("");
     const [newFieldDescription, setNewFieldDescription] = React.useState<string>("");
-    const [newFieldDataType, setNewFieldDataType] = React.useState<AnnotationType | undefined>();
+    const [newFieldDataType, setNewFieldDataType] = React.useState<AnnotationType>(
+        AnnotationType.STRING
+    );
     const [newDropdownOption, setNewDropdownOption] = React.useState<string>("");
     const [dropdownOptions, setDropdownOptions] = React.useState<IComboBoxOption[]>([]);
     const [submissionStatus, setSubmissionStatus] = React.useState<AnnotationStatus | undefined>();
@@ -148,8 +150,8 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
         const annotation = new Annotation({
             annotationDisplayName: newFieldName,
             annotationName: newFieldName,
-            description: newFieldDescription ?? "",
-            type: newFieldDataType ?? AnnotationType.STRING,
+            description: newFieldDescription,
+            type: newFieldDataType,
         });
         // File editing step occurs after dispatch is processed and status is updated
         dispatch(
@@ -196,7 +198,9 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
                             };
                         })}
                         onChange={(option) =>
-                            setNewFieldDataType((option?.key as AnnotationType) || "")
+                            setNewFieldDataType(
+                                (option?.key as AnnotationType) || AnnotationType.STRING
+                            )
                         }
                     />
                     {newFieldDataType === AnnotationType.DROPDOWN && (
--- a/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
+++ b/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
@@ -31,7 +31,7 @@ enum EditStep {
 
 interface NewAnnotationProps {
     onDismiss: () => void;
-    hasUnsavedChanges: (arg: boolean) => void;
+    setHasUnsavedChanges: (arg: boolean) => void;
     selectedFileCount: number;
 }
 
@@ -48,7 +48,7 @@ interface AnnotationStatus {
 export default function NewAnnotationPathway(props: NewAnnotationProps) {
     const dispatch = useDispatch();
     // Destructure to prevent unnecessary useEffect triggers
-    const { onDismiss, hasUnsavedChanges } = props;
+    const { onDismiss, setHasUnsavedChanges } = props;
 
     const [step, setStep] = React.useState<EditStep>(EditStep.CREATE_FIELD);
     const [newValues, setNewValues] = React.useState<AnnotationValue | undefined>();
@@ -91,7 +91,7 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
                                 message: `Failed to create annotation: ${e}`,
                             });
                         } finally {
-                            hasUnsavedChanges(false);
+                            setHasUnsavedChanges(false);
                             onDismiss();
                         }
                     }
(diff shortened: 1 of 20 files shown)


packages/core/components/EditMetadata/NewAnnotationPathway.tsx
```diff
@@ -18,21 +31,74 @@ enum EditStep {
...
     const [dropdownOptions, setDropdownOptions] = React.useState<IComboBoxOption[]>([]);
+    const [submissionStatus, setSubmissionStatus] = React.useState<AnnotationStatus | undefined>();
+
+    const statuses = useSelector(interaction.selectors.getProcessStatuses);
+    const annotationCreationStatus = React.useMemo(
+        () => statuses.find((status) => status.processId === newFieldName),
+        [newFieldName, statuses]
+    );
+    // Check for updates to the annotation submission status
+    React.useEffect(() => {
```
aswallace, line 68: We aren't able to directly chain off of the createAnnotation dispatch (we don't use thunk in this codebase) so instead we need to wait for the status to update in state and then respond accordingly when we see `SUCCEEDED`

kmitcham:
We might want to try again to get a (monthly| after 5 new annotations|some other period) meeting for people who want to govern the annotation list.  Or a least a report of the existing annotations?

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
```diff
@@ -18,21 +31,87 @@ enum EditStep {
...
+        checkForStatusUpdates();
+    }, [annotationCreationStatus, dispatch, hasUnsavedChanges, newFieldName, newValues, onDismiss]);
+
+    const onChangeAlphanumericField = (
+        e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+        newValue: string | undefined
+    ) => {
+        const regex = /^[\w\-\s]+$/g;
+        // Restricts character entry to alphanumeric
+        if (newValue && !regex.test(newValue)) {
```
kmitcham, line 109: aww, no injection attacks?  But what if scientists want to mark an image ";Drop tables;":True?

kmitcham (approved):

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
```diff
@@ -18,21 +31,87 @@ enum EditStep {
...
+    const dispatch = useDispatch();
+    // Destructure to prevent unnecessary useEffect triggers
+    const { onDismiss, hasUnsavedChanges } = props;
+
     const [step, setStep] = React.useState<EditStep>(EditStep.CREATE_FIELD);
-    const [newValues, setNewValues] = React.useState<string | undefined>();
+    const [newValues, setNewValues] = React.useState<AnnotationValue | undefined>();
     const [newFieldName, setNewFieldName] = React.useState<string>("");
+    const [newFieldDescription, setNewFieldDescription] = React.useState<string>("");
     const [newFieldDataType, setNewFieldDataType] = React.useState<AnnotationType | undefined>();
```
SeanLeRoy, line 57: Should we just set a default for this if we `onSubmit` do anyway?

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
```diff
@@ -53,30 +132,56 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
...
+            setSubmissionStatus({
+                status: ProcessStatus.ERROR,
+                message: `Missing ${!newFieldName ? "field name" : "values for file"}`,
+            });
+            return;
+        }
+        const annotation = new Annotation({
+            annotationDisplayName: newFieldName,
+            annotationName: newFieldName,
+            description: newFieldDescription ?? "",
```
SeanLeRoy, line 151: Looks like `newFieldDescription` is never `undefined`

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
```diff
@@ -132,17 +237,38 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
...
+                            {
+                                value: undefined,
+                                fileCount: props.selectedFileCount,
+                            } as ValueCountItem,
+                        ]}
+                        dropdownOptions={dropdownOptions}
+                        onChange={(value) => setNewValues(value)}
+                    />
+                    {submissionStatus && (
+                        <div
```
SeanLeRoy, line 258: _**very**_ nice feedback here

packages/core/entity/Annotation/index.ts
```diff
@@ -14,11 +17,21 @@ export interface AnnotationResponse {
     annotationDisplayName: string;
     annotationName: string;
     description: string;
-    type: string;
+    type: string | AnnotationType;
```
SeanLeRoy, line 20: hmm why potentially `string`? simplifying the type whenever possible seems ideal

packages/core/entity/Annotation/index.ts
```diff
@@ -14,11 +17,21 @@ export interface AnnotationResponse {
...
     isOpenFileLink?: boolean;
     units?: string;
 }
 
+/**
+ * TODO: MMS queries return a different JSON structure than FES
```
SeanLeRoy, line 26: TODO?

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -140,4 +140,13 @@ export default class DatabaseAnnotationService implements AnnotationService {
             }, [] as string[])
             .filter((annotation) => !annotationSet.has(annotation));
     }
+
+    public async createAnnotation(annotation: Annotation): Promise<any> {
+        const tableName = this.dataSourceNames.sort().join(", ");
```
SeanLeRoy, line 145: I think there is a constant in here that tracks the separater for data source names? Worth checking might be totally wrong.

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -140,4 +140,13 @@ export default class DatabaseAnnotationService implements AnnotationService {
...
+        return await this.databaseService.addNewColumn(
```
SeanLeRoy, line 146: Can remove `await` and `async` from this.

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -140,4 +140,13 @@ export default class DatabaseAnnotationService implements AnnotationService {
             }, [] as string[])
             .filter((annotation) => !annotationSet.has(annotation));
     }
+
+    public async createAnnotation(annotation: Annotation): Promise<any> {
```
SeanLeRoy, line 144: Can we narrow the `any` here

packages/core/services/AnnotationService/HttpAnnotationService/index.ts
```diff
@@ -116,4 +118,28 @@ export default class HttpAnnotationService extends HttpServiceBase implements An
...
+     * Creates a new annotation via the metadata-management-service
+     * @param annotation The new annotation to create
+     * @param annotationOptions If not empty, pre-set options for annotations of type Dropdown
+     */
+    public async createAnnotation(
+        annotation: Annotation,
+        annotationOptions: string[] = []
+    ): Promise<AnnotationResponseMms[]> {
+        const requestUrl = `${this.metadataManagementServiceBaseURl}/${HttpAnnotationService.BASE_MMS_ANNOTATION_URL}`;
+        const annotationType = (annotation.type as AnnotationType) || AnnotationType.STRING;
```
SeanLeRoy, line 132: Is `type` ever undefined - I don't think it should be

packages/core/services/DatabaseService/index.ts
```diff
@@ -529,4 +529,23 @@ export default abstract class DatabaseService {
...
+            // Cache is now invalid since we added a column
+            this.dataSourceToAnnotationsMap.delete(datasourceName);
+
+            if (description?.trim() && this.existingDataSources.has(this.SOURCE_METADATA_TABLE)) {
+                await this
+                    .execute(`INSERT INTO "${this.SOURCE_METADATA_TABLE}" ("Column Name", "Description")
+                        VALUES ('${columnName}', '${description}');`);
+            }
+        } catch (err) {
+            throw err;
```
SeanLeRoy, line 548: lol catch to throw it? I think we can just remove this then

packages/core/state/metadata/logics.ts
```diff
@@ -108,6 +114,60 @@ const receiveAnnotationsLogic = createLogic({
...
+        const annotationService = interaction.selectors.getAnnotationService(getState());
+        const applicationVersion = interaction.selectors.getApplicationVersion(getState());
+        if (annotationService instanceof HttpAnnotationService) {
+            if (applicationVersion) {
+                annotationService.setApplicationVersion(applicationVersion);
+            }
+            annotationService.setHttpClient(httpClient);
+        }
+
+        return new Promise<AnnotationResponseMms[]>(async (resolve, reject) => {
```
SeanLeRoy, line 138: Hmmm do we want to `return` here or should we be `awaiting` the `Promise`? I'm uncertain if it actually matters...?

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
```diff
@@ -18,21 +31,87 @@ enum EditStep {
 
 interface NewAnnotationProps {
     onDismiss: () => void;
-    onUnsavedChanges: () => void;
+    hasUnsavedChanges: (arg: boolean) => void;
```
SeanLeRoy, line 34: `hasUnsavedChanges` this is the naming convention of a boolean so it feels weird here to me - it seems like this is just the same thing as `onDismiss` but has an argument so would it make sense to add a parameter of "hasUnsavedChanges: boolean" to `onDismiss`?

packages/core/services/DatabaseService/index.ts
aswallace: Good catch lol I think this is a debugging artifact

packages/core/entity/Annotation/index.ts
aswallace: I think this is actually because DatabaseAnnotationService uses other internally defined types, e.g., `DatabaseService.OPEN_FILE_LINK_TYPE,`

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
aswallace: added a default in [73a8316](https://github.com/AllenInstitute/biofile-finder/pull/406/commits/73a83161edfb1645a108729511b49fe509ca46e7)

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
aswallace: removed in [73a8316](https://github.com/AllenInstitute/biofile-finder/pull/406/commits/73a83161edfb1645a108729511b49fe509ca46e7)

packages/core/state/metadata/logics.ts
aswallace: changed to `await` in [b3264d1](https://github.com/AllenInstitute/biofile-finder/pull/406/commits/b3264d1b8ea39f7b15e1cf385b839c9561ee0688). I think for our dispatches it doesn't matter, but `await` is more consistent with what we do elsewhere

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
aswallace: changed in [b3264d1](https://github.com/AllenInstitute/biofile-finder/pull/406/commits/b3264d1b8ea39f7b15e1cf385b839c9561ee0688), 95% sure it's the right constant

packages/core/entity/Annotation/index.ts
aswallace: changed in [ceede0e](https://github.com/AllenInstitute/biofile-finder/pull/406/commits/ceede0e5f22524b7a9e21df69a046a1c1a810014)

SeanLeRoy (approved):

PR #411, opened 2025-02-05
bugfix/cache-modal-performance

### LINKED ISSUE #409: Bug: Copy to VAST with thousands of files crashes BFF
Steps to reproduce:
* Click a file
* Scroll down
* Wait for more files to load
* Hold shift and click another file
* Scroll down
* Repeat until 5000 files are selected
* Right click on a file and select "Copy to VAST"

Actual behavior: The app becomes unresponsive and goes to a blank white screen

Expected behavior: User can extend expiration dates or copy all these files to the VAST

BrianWhitneyAI:
### Description 

The purpose of this PR is to resolve #409 and make the copy to local cache performant for 10,000+ files. This pr add a chunked response, waiting state and error handling around this process to make it smoother for users.

### Testing
I tested this with the startup query date range that BFF has (with about 7,500 files) with no issues. Still investigating larger selections.

### DIFF
--- a/packages/core/components/Modal/CopyFileManifest/index.tsx
+++ b/packages/core/components/Modal/CopyFileManifest/index.tsx
@@ -90,6 +90,7 @@ export default function CopyFileManifest({ onDismiss }: ModalProps) {
 
     const [fileDetails, setFileDetails] = React.useState<FileDetail[]>([]);
     const [loading, setLoading] = React.useState<boolean>(true);
+    const [error, setError] = React.useState<string | null>(null);
 
     React.useEffect(() => {
         async function fetchFiles() {
@@ -97,7 +98,8 @@ export default function CopyFileManifest({ onDismiss }: ModalProps) {
                 const details = await fileSelection.fetchAllDetails();
                 setFileDetails(details);
             } catch (err: any) {
-                throw new Error(err.message || "Error fetching file details.");
+                console.error("Error fetching file details:", err);
+                setError(err.message || "Error fetching file details.");
             } finally {
                 setLoading(false);
             }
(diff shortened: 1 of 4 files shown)


kmitcham (approved):

packages/core/components/Modal/CopyFileManifest/index.tsx
```diff
@@ -60,8 +59,8 @@ function FileTable({ files, title }: { files: FileDetail[]; title: string }) {
                             </tr>
                         </thead>
                         <tbody>
-                            {files.map((file) => (
-                                <tr key={file.id}>
+                            {files.map((file, index) => (
+                                <tr key={file.id || index}>
```
SeanLeRoy: All files should have an id, whats the context behind this change?

packages/core/components/Modal/CopyFileManifest/index.tsx
```diff
@@ -92,13 +89,20 @@ export default function CopyFileManifest({ onDismiss }: ModalProps) {
...
     React.useEffect(() => {
-        async function fetchDetails() {
-            const details = await fileSelection.fetchAllDetails();
-            setFileDetails(details);
+        async function fetchFiles() {
+            try {
+                const details = await fileSelection.fetchAllDetails();
+                setFileDetails(details);
+            } catch (err: any) {
+                throw new Error(err.message || "Error fetching file details.");
```
SeanLeRoy, line 100: This error won't get seen by a user in a useful manner. Having it be set to the component's state like you had earlier and then displaying the error would make this more useful

SeanLeRoy (commented):
So did the caching fix fix this issue overall then?

pgarrison (approved):
I think your recent changes made some of my comments unnecessary.

I tested this in prod and was able to successfully open the modal with over 6,000 files selected. The #412/#408 bug prevented me from selecting more files.

I also uncovered a UI bug that I suspect was already in BFF. I'll file a new issue

packages/core/components/Modal/CopyFileManifest/index.tsx
BrianWhitneyAI: Resolved in [c69c4c8](https://github.com/AllenInstitute/biofile-finder/pull/411/commits/c69c4c89f925a7b94f5fbcd946d82463bf23d721)

SeanLeRoy (approved):

PR #416, opened 2025-02-07
bugfix/out-of-sync-data-source

### LINKED ISSUE #398: Selected data source does not match actual data source
## To reproduce

1. open BFF
2. select 'staging' data source
3. quit BFF
4. open BFF

You will now see 'production' as the selected data source, yet the displayed data will be from staging.

Also, an interesting note is that it seems like if you install a new version of BFF in between steps 3 and 4, you will still have this issue.

BrianWhitneyAI:
### Description 

The purpose of this PR is to resolve #398 and centralize the data source variable in the main process so that the "Data Source" menu always reflects the current state, allows users to change it, and persists those changes across app restarts, refreshes and cache clears.

### Changes 
- Added an envManager.ts module that loads the persisted environment on startup and updates it (persisting the new value and rebuilding the menu).

- Changed the Data Source menu to be built dynamically (via a function) so that its radio buttons always reflect the current environment.

- Modified the menu click to call updateEnvironment() (which updates state and rebuilds the menu) and send an event. The main process loads the environment before building the menu, and the renderer listens for updates to re‑render the UI.

### Testing

1) Open app and select staging 
2) Refresh / Reload / Clear Cache
3) Check that correct data source is selected and matches data
4) repeat and iterate for different env / data sources


BrianWhitneyAI:
Something funky, github doesn't seem to be tracking indent changes. I tried to adhere to 4 space indents but may have missed some.

kmitcham (approved):

packages/desktop/src/main/envManager.ts
```diff
@@ -0,0 +1,30 @@
...
+import Store from "electron-store";
+import { Environment } from "../util/constants";
+
+const store = new Store();
+
+/**
+ * Loads the persisted environment from storage.
+ * Defaults to Production if no value exists.
+ */
+export function loadEnvironment(): Environment {
```
SeanLeRoy, line 12: Could we use the `PersistedStorageService` instead of this? It seems like the same purpose.

packages/desktop/src/renderer/index.tsx
```diff
@@ -37,16 +36,15 @@ const KeyDownHandler: React.FC<{ clearStore: () => void }> = ({ clearStore }) =>
 // Clears the persistent store but retains `Environment` to prevent misalignment
 // between the data source and the selected menu item in the app.
 const clearPersistentStore = () => {
-    const currentEnvironment = global.environment;
+    const currentEnvironment = global.environment || "PRODUCTION";
     persistentConfigService.clear();
     persistentConfigService.persist({ [PersistedConfigKeys.Environment]: currentEnvironment });
```
SeanLeRoy, line 41: like this seems like a conflict with the other persisted storage access - using one access method (persistentConfigService) seems ideal

BrianWhitneyAI: resolved in [94c1070](https://github.com/AllenInstitute/biofile-finder/pull/416/commits/94c1070949efacef943d7487829be2f23543fab0)

packages/desktop/src/main/index.ts
```diff
@@ -41,6 +49,14 @@ const createMainWindow = () => {
         width: 1200,
     });
 
+    // When the renderer finishes loading, send the current environment.
+    mainWindow.webContents.on("did-finish-load", () => {
```
SeanLeRoy, line 53: Where is this event sent? Do we still need to listen here for this for the `mainWindow` to know when the environment changed?

BrianWhitneyAI: oh whoops forgot to remove. Fixed in [586d76b](https://github.com/AllenInstitute/biofile-finder/pull/416/commits/586d76bc92f89fe875762b56a116527e5be3c6a8)

SeanLeRoy (approved):

PR #421, opened 2025-02-10
bugfix/modal-button-location

### LINKED ISSUE #374: Download manifest modal styling issues
1. I discovered the content overlaps, so I think it just needs a min-height added.
Screenshot 2024-12-19 at 11 14 47 AM

2. Content has alignment issue due to left-right padding. Warning: the filter, grouping, etc modals share the same padding and will be affected. In other words, maybe padding needs to be added globally to all modals and then removed from .ListPicker-module__container--JJf_E???

### LINKED ISSUE #413: Summary of Files to Download to VAST is hidden for long lists
Steps to reproduce:
* Select a group of files that contains over a dozen files already on VAST and over a dozen not on VAST
* Right click a file and choose "Copy to VAST"
* Actual behavior: "Files to Download to VAST" shows no summary of number of files and total size, and the Cancel and Confirm buttons overlap the list of Files to Download to VAST (see screenshot)
* Expected behavior: Below the list of "Files to Download to VAST" and above the Cancel and Confirm buttons is the number of files to download and total size

Image

### LINKED ISSUE #418: Copy to VAST modal: cannot read full filenames
Use case: I want a group of files with similar names to be on the VAST, and one of them isn't, so I want to figure out which one.

Steps to reproduce:
* Select a group of files with identical initial prefixes and final suffixes, where some are on Vast and some aren't.
  * E.g., in prod FMS we have a bunch of `3500003704_Scan_..._outlines.png` files like `3500003704_Scan_12-13-2019-7-49-29-AM_Well_B1_Xmm8.901_Ymm14.837_Ch1_-1um_outlines.png`
* Right click -> Copy to VAST
* Try to figure out which ones are not on the Vast
* Actual behavior: all the file names appear the same, e.g. `350000370...ines.png` (see screenshot)
* Expected behavior: The modal can be resized and/or scrolled horizontally to view the full filenames

Image

BrianWhitneyAI:
### Description 

The purpose of this PR is to propose a solution to our increasing size of body content for modals. We have seen #413 where the buttons overlay the content (Present for other modals as well). This PR adds a scroll bar to the body of all modals if it exceeds a certain size and limits the total body to 55% of the total modal. It could use some love around spacing...

### UPDATES:
[22c97f9](https://github.com/AllenInstitute/biofile-finder/pull/421/commits/22c97f98a94d26f89944348b6075d4f64e0041e9) and [ec58021](https://github.com/AllenInstitute/biofile-finder/pull/421/commits/ec58021d19b2730a38b31170a7dd186b1b672c30) resolve #413 
[1e72d98](https://github.com/AllenInstitute/biofile-finder/pull/421/commits/1e72d98b0e07e5ce8ba5c002e3f957527d572e8d) resolves #374
[ad84480](https://github.com/AllenInstitute/biofile-finder/pull/421/commits/ad844802bf98faafd4dabef90e7a71ecf2dcc125) resolves #418

### DIFF
--- a/packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
+++ b/packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
@@ -1,5 +1,8 @@
 .body-container {
+    padding: 0 1em 0 0;
     position: relative;
+    max-height: 55vh;
+    overflow-y: auto;
 }
 
 .note {
--- a/packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
+++ b/packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
@@ -1,3 +1,7 @@
+:root {
+    --file-size-column-width: 7em;
+}
+
 .modal-container {
     width: 100%;
     max-width: 40vw;
@@ -7,7 +11,7 @@
 
 .body-container {
     width: 100%;
-    padding: 0 1em 0 0;
+    padding: 0 0 0 0;
     position: relative;
     max-height: 55vh;
     overflow-y: hidden;
@@ -84,14 +88,14 @@
 
 .file-table th:last-child,
 .file-table td:last-child {
-    width: 10em;
-    min-width: 10em;
-    max-width: 10em;
+    width: var(--file-size-column-width);
+    min-width: var(--file-size-column-width);
+    max-width: var(--file-size-column-width);
 }
 
 .file-table th:first-child,
 .file-table td:first-child {
-    width: calc(100% - 10em);
+    width: calc(100% - var(--file-size-column-width));
 }
 
 
--- a/packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
+++ b/packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
@@ -1,8 +1,17 @@
+.modal-container {
+    width: 100%;
+    max-width: 40vw;
+    margin: 0 auto;
+    box-sizing: border-box;
+}
+
 .body-container {
+    width: 100%;
     padding: 0 1em 0 0;
     position: relative;
     max-height: 55vh;
     overflow-y: hidden;
+    box-sizing: border-box;
 }
 
 .note {
@@ -11,6 +20,7 @@
 }
 
 .table-container {
+    width: 100%;
     margin-bottom: 0;
 }
 
@@ -21,13 +31,17 @@
 }
 
 .table-wrapper {
+    width: 100%;
     position: relative;
+    overflow-x: hidden;
 }
 
 .file-table-container {
+    width: 100%;
     min-height: 45px;
     max-height: 15vh;
     overflow-y: auto;
+    overflow-x: hidden;
     background-color: var(--secondary-background-color);
     position: relative;
     z-index: 1;
@@ -46,6 +60,7 @@
 
 .file-table {
     width: 100%;
+    table-layout: fixed;
     border-collapse: collapse;
     position: relative;
     z-index: 1;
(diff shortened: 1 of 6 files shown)


BrianWhitneyAI:
Screenshot 2025-02-10 at 10 27 23 AM

SeanLeRoy (approved):
This seems good - approving under the following assumptions:
1) You've checked all the other modals to make sure their scrolling is working as expected
2) The total counts for the second list in the screenshot just don't exist as opposed to are being pushed offscreen

pgarrison:
Echoing Sean's comment, I expect there to be total counts for the Files to Download to VAST section, and ideally the user wouldn't have to scroll to see them.

I think this modal could look great (and work well for users) if the outer scrollbar were removed and the two file lists were made shorter to fit both lists and the total counts on screen at once.

kmitcham (approved):

BrianWhitneyAI:
This is having some performance issues with large selections. not sure what the best solution here is (or if we need to fix it) 

https://github.com/user-attachments/assets/650a54b8-73e6-4e90-81e3-4178b1d53647

SeanLeRoy (approved):
My guess is the number of elements being rendered at once is too much to handle responsive behavior for - in a few places in the app we use a list from fluent that lets you lazily render the list elements which would solve this if that’s the issue

packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
```diff
@@ -63,10 +82,24 @@
...
+.file-table th:last-child,
+.file-table td:last-child {
+    width: 10em;
+    min-width: 10em;
+    max-width: 10em;
+}
+
+.file-table th:first-child,
+.file-table td:first-child {
+    width: calc(100% - 10em);
```
pgarrison, line 94: Since `10em` is used multiple times where it's important to keep it identical, I recommend introducing it as a CSS variable.

Also, 10em might be a tad too large? Just judging by the whitespace in your example movie.

pgarrison (approved):
Thanks, love it!

packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
aswallace: I'd add to this that the right padding on .body-container makes it ever so slightly off-center which I think makes the left margin look smaller than it is

packages/core/components/Modal/CopyFileManifest/index.tsx
```diff
@@ -11,6 +11,19 @@ import { interaction, selection } from "../../../state";
 
 import styles from "./CopyFileManifest.module.css";
 
+function ResponsiveFileName({ fileName }: { fileName: string }) {
+    const lastDot = fileName.lastIndexOf(".");
+    const base = lastDot > 0 ? fileName.slice(0, lastDot) : fileName;
+    const extension = lastDot > 0 ? fileName.slice(lastDot) : "";
```
aswallace, line 17: this will not account for file types like .ome.zarr, although that may not be a major concern

aswallace (approved):

packages/core/components/Modal/CopyFileManifest/index.tsx
BrianWhitneyAI: or other files that have a tag like _outlines.ext that are derivitive. I dont think its a big deal.

packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
BrianWhitneyAI: ah yeah that padding was originally for the scroll bar but is no longer needed.

BrianWhitneyAI: resolved in [82c5463](https://github.com/AllenInstitute/biofile-finder/pull/421/commits/82c546325862e26756bf900003a417604881a9f5)

PR #426, opened 2025-02-11
(Med.)[BFF-403] Add "Go to location" option to FileAccessContextMenu

### LINKED ISSUE #403: Feature req: Add a method to open a local file path in the user's native file explorer
# User Story

As a user, I want to be able to swiftly navigate to a file's location in my native file explorer, should that file be available on a local file system.

# Context

In implementing FMS multifile support, one of the challenges we have encountered is the expectation that user's should be able to select specific programs that they can then use to open a file directly from BFF. This appears to be difficult when using programs that users typically use to launch multifiles, e.g. using Zen to open a `.sldy` file.

Instead of implementing custom logic for handling the opening of specific file types using specific applications, we could instead provide functionality that allows the user to open their native file explorer directly to the location of a locally-available file. From there, they can open the file the way they typically would from their native file system.

# Details
* The option to do this would only be available in the desktop version of the app.
* While not necessarily restricted to internal usage, this would primarily benefit our internal users.
* This feature could potentially only be available if the path to the selected file exists on the user's file system (i.e. it won't appear if the `/allen` drive is not mounted).

# Questions
* Is this already a feature? If so, I have no idea where, and the UX around the feature should be improved.
* How exactly should this be implemented?
   * As a separate context menu item in the file list and a separate button in the file details pane?
   * As an addition to the "Open with >" context menu item / dropdown? E.g. "Open with >" -> "Native file browser"?
       * Would users even understand the term "native file browser"? What wording should be used for this if not?

tyler-foster:
closes #403 

### Description

Adds an item to the file list's context menu, "Go to file location", which will appear if the user is running the app in desktop mode and will be enabled if their selected file has a value for `localPath`. Clicking on it will open the file's Vast path in the user's native file browser.

We can consider this as a little escape hatch for the "Open with..." feature, which gets a bit more difficult to use when working with next generation file formats like OME-Zarr. Users can use it to jump to a file and use their operating system's "open with" feature, or drag and drop it somewhere.

image

---

Is there a setup for testing hooks like this in BFF? I only did manual testing, and I wasn't actually able to test opening something from the Vast, because I don't have it mounted on my Macbook and I can't seem to get BFF to launch correctly on my Ubuntu setup.

I'm still open to suggestions / alternative ideas for the wording!
Skipping the fuss around adding similar functionality to the file details pane, for now.

### DIFF
--- a/packages/core/hooks/useFileAccessContextMenu.ts
+++ b/packages/core/hooks/useFileAccessContextMenu.ts
@@ -1,4 +1,5 @@
 import { ContextualMenuItemType, IContextualMenuItem } from "@fluentui/react";
+import { isNil } from "lodash";
 import * as React from "react";
 import { useDispatch, useSelector } from "react-redux";
 
@@ -74,11 +75,7 @@ export default (filters?: FileFilter[], onDismiss?: () => void) => {
                               key: "go-to-file-location",
                               text: "Go to file location",
                               title: "Show selected file in your native file browser",
-                              disabled:
-                                  fileDetails?.localPath === null ||
-                                  fileDetails?.localPath === undefined ||
-                                  fileSelection.count() > 1 ||
-                                  (!filters && fileSelection.count() === 0),
+                              disabled: isNil(fileDetails?.localPath),
                               iconProps: { iconName: "FolderOpen" },
                               onClick() {
                                   if (fileDetails) {
--- a/packages/core/hooks/useFileAccessContextMenu.ts
+++ b/packages/core/hooks/useFileAccessContextMenu.ts
@@ -1,21 +1,11 @@
 import { ContextualMenuItemType, IContextualMenuItem } from "@fluentui/react";
 import * as React from "react";
 import { useDispatch, useSelector } from "react-redux";
-import { shell } from "electron";
 
 import useOpenWithMenuItems from "./useOpenWithMenuItems";
 import FileDetail from "../entity/FileDetail";
 import FileFilter from "../entity/FileFilter";
 import { interaction, selection } from "../state";
-
-/**
- * Reveals a selected file in the user's native file browser
- */
-const openNativeFileBrowser = (fileDetails: FileDetail) => {
-    if (fileDetails.localPath) {
-        shell.showItemInFolder(fileDetails.localPath);
-    }
-};
 
 /**
  * Custom React hook for creating the file access context menu.
@@ -92,7 +82,9 @@ export default (filters?: FileFilter[], onDismiss?: () => void) => {
                               iconProps: { iconName: "FolderOpen" },
                               onClick() {
                                   if (fileDetails) {
-                                      openNativeFileBrowser(fileDetails);
+                                      dispatch(
+                                          interaction.actions.openNativeFileBrowser(fileDetails)
+                                      );
                                   }
                               },
                           },
--- a/packages/core/hooks/useFileAccessContextMenu.ts
+++ b/packages/core/hooks/useFileAccessContextMenu.ts
@@ -1,11 +1,21 @@
 import { ContextualMenuItemType, IContextualMenuItem } from "@fluentui/react";
 import * as React from "react";
 import { useDispatch, useSelector } from "react-redux";
+import { shell } from "electron";
 
 import useOpenWithMenuItems from "./useOpenWithMenuItems";
 import FileDetail from "../entity/FileDetail";
 import FileFilter from "../entity/FileFilter";
 import { interaction, selection } from "../state";
+
+/**
+ * Reveals a selected file in the user's native file browser
+ */
+const openNativeFileBrowser = (fileDetails: FileDetail) => {
+    if (fileDetails.localPath) {
+        shell.showItemInFolder(fileDetails.localPath);
+    }
+};
 
 /**
  * Custom React hook for creating the file access context menu.
@@ -68,6 +78,26 @@ export default (filters?: FileFilter[], onDismiss?: () => void) => {
                         items: openWithSubMenuItems,
                     },
                 },
+                ...(isQueryingAicsFms && !isOnWeb
+                    ? [
+                          {
+                              key: "go-to-file-location",
+                              text: "Go to file location",
+                              title: "Show selected file in your native file browser",
+                              disabled:
+                                  fileDetails?.localPath === null ||
+                                  fileDetails?.localPath === undefined ||
+                                  fileSelection.count() > 1 ||
+                                  (!filters && fileSelection.count() === 0),
+                              iconProps: { iconName: "FolderOpen" },
+                              onClick() {
+                                  if (fileDetails) {
+                                      openNativeFileBrowser(fileDetails);
+                                  }
+                              },
+                          },
+                      ]
+                    : []),
                 {
                     key: "save-as",
                     text: "Save metadata as",
--- a/packages/desktop/src/services/test/FileViewerServiceElectron.test.ts
+++ b/packages/desktop/src/services/test/FileViewerServiceElectron.test.ts
@@ -1,15 +1,11 @@
 import childProcess from "child_process";
 
-import fs from "fs";
-
 import { expect } from "chai";
 import { createSandbox } from "sinon";
 
 import FileViewerServiceElectron from "../FileViewerServiceElectron";
 import NotificationServiceElectron from "../NotificationServiceElectron";
-import { Environment, RUN_IN_RENDERER } from "../../util/constants";
-import FileDetail from "../../../../core/entity/FileDetail";
-import { FileNotFoundError } from "../../../../core/errors";
+import { RUN_IN_RENDERER } from "../../util/constants";
 
 describe(`${RUN_IN_RENDERER} FileViewerServiceElectron`, () => {
     const sandbox = createSandbox();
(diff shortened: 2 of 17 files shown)


packages/core/hooks/useFileAccessContextMenu.ts
```diff
@@ -68,6 +78,26 @@ export default (filters?: FileFilter[], onDismiss?: () => void) => {
...
+                          {
+                              key: "go-to-file-location",
+                              text: "Go to file location",
+                              title: "Show selected file in your native file browser",
+                              disabled:
+                                  fileDetails?.localPath === null ||
+                                  fileDetails?.localPath === undefined ||
+                                  fileSelection.count() > 1 ||
+                                  (!filters && fileSelection.count() === 0),
+                              iconProps: { iconName: "FolderOpen" },
```
tyler-foster, line 92: I only figured out I could use this icon by luck. I still have no idea where I can find valid Fluent UI icons.

Brian W. showed me this:
https://react.fluentui.dev/?path=/docs/icons-catalog--docs

But it would seem `FolderOpen` is not in their list.
It _does_ appear here but the icon on the list doesn't match what I get...
https://github.com/microsoft/fluentui-system-icons/blob/main/icons_regular.md

aswallace: https://developer.microsoft.com/en-us/fluentui#/styles/web/icons#fabric-react

packages/core/hooks/useFileAccessContextMenu.ts
```diff
@@ -1,12 +1,22 @@
 import { ContextualMenuItemType, IContextualMenuItem } from "@fluentui/react";
 import * as React from "react";
 import { useDispatch, useSelector } from "react-redux";
+import { shell } from "electron";
```
aswallace, line 4: nit: bff imports are generally alphabetical (tho not enforced by linter), so would go in line 2

kmitcham (approved):

packages/core/hooks/useFileAccessContextMenu.ts
aswallace: Where imported from

aswallace (approved):

packages/core/hooks/useFileAccessContextMenu.ts
SeanLeRoy, line 4: `electron` is not available on web so this would break our web deploy. Platform specific dependencies have to go into the platform specific package so `electron` stuff would go into `desktop`

SeanLeRoy (changes requested):

packages/core/hooks/useFileAccessContextMenu.ts
```diff
@@ -68,6 +78,26 @@ export default (filters?: FileFilter[], onDismiss?: () => void) => {
                         items: openWithSubMenuItems,
                     },
                 },
+                ...(isQueryingAicsFms && !isOnWeb
+                    ? [
+                          {
+                              key: "go-to-file-location",
+                              text: "Go to file location",
+                              title: "Show selected file in your native file browser",
+                              disabled:
```
SeanLeRoy, line 87: For things like this where they can only apply to one item we just use the focused item (of which there has to be one by default) so it would make sense to me to remove the restriction on `fileSelection.count() > 1`. This _should_ be disabled or removed from the list of options when rendering on a selected folder though.
`disabled: isNil(fileDetails?.localPath)` should cover you

packages/core/hooks/useFileAccessContextMenu.ts
```diff
@@ -1,12 +1,22 @@
...
 import FileDetail from "../entity/FileDetail";
 import FileFilter from "../entity/FileFilter";
 import { interaction, selection } from "../state";
 
+/**
+ * Reveals a selected file in the user's native file browser
+ */
+const openNativeFileBrowser = (fileDetails: FileDetail) => {
+    if (fileDetails.localPath) {
+        shell.showItemInFolder(fileDetails.localPath);
```
SeanLeRoy, line 16: This will likely need to be dispatched as an action to the redux state where a side-effect would occur that would call the `FileViewerService` which would do this functionality in a platform specific way. For web it would just throw a not implemented error.

lynwilhelm:
Why are there 2 "Open" options in this image?

packages/core/hooks/useFileAccessContextMenu.ts
tyler-foster: Good point! Addressed in 0b8c222201b63d8dd6c926f1a52f6e51f39cd10d , moved files around in 0b8c222201b63d8dd6c926f1a52f6e51f39cd10d

packages/core/hooks/useFileAccessContextMenu.ts
tyler-foster: Good point! Addressed in 0b8c222201b63d8dd6c926f1a52f6e51f39cd10d and 0b8c222201b63d8dd6c926f1a52f6e51f39cd10d

packages/core/hooks/useFileAccessContextMenu.ts
tyler-foster: That's waaaay easier. Done in 866d43b759a52c0e1da9297b489aa2afec56c31f

packages/desktop/src/services/test/FileViewerServiceElectron.test.ts
```diff
@@ -69,4 +73,41 @@ describe(`${RUN_IN_RENDERER} FileViewerServiceElectron`, () => {
...
+                    ],
+                    file_path: "/allen/aics/path/to/file.txt",
+                },
+                Environment.TEST
+            );
+
+            // Act + Assert
+            expect(() => {
+                service.openNativeFileBrowser(fileDetail);
+            }).to.throw(FileNotFoundError);
```
tyler-foster, line 110: @SeanLeRoy @aswallace Help! This unit test doesn't actually work the way it looks like it should. It's currently failing on the GitHub build but passing on my local machine.

I know one issue with it is that the `stub()` I'm using _doesn't really stub what it looks like it should_. You can flip the `.returns()` value all you want and it doesn't cause the test to fail or pass.

Does anyone have better experience with / advice for using Sinon?

packages/web/src/services/FileViewerServiceWeb.ts
```diff
@@ -4,4 +4,8 @@ export default class FileViewerServiceWeb implements FileViewerService {
     public async open(): Promise<void> {
         throw new Error("FileViewerServiceWeb::open is not yet implemented");
     }
+
+    public openNativeFileBrowser(): void {
+        throw Error("ExecutionEnvServiceWeb::openNativeFileBrowser not yet implemented");
```
SeanLeRoy, line 9:
```suggestion
        throw Error("FileViewerServiceWeb::openNativeFileBrowser not yet implemented");
```

packages/desktop/src/services/test/FileViewerServiceElectron.test.ts
```diff
@@ -69,4 +73,41 @@ describe(`${RUN_IN_RENDERER} FileViewerServiceElectron`, () => {
             expect(attemptedToShowError).to.be.true;
         });
     });
+
+    describe("openNativeFileBrowser", () => {
+        let existsSyncStub: SinonStub;
+
+        before(() => {
+            existsSyncStub = stub(fs, "existsSync");
```
SeanLeRoy, line 81: You can use the sandbox established before
```suggestion
            existsSyncStub = sandbox.stub(fs, "existsSync");
```

SeanLeRoy: Also could remove the `after` two lines below this

packages/desktop/src/services/test/FileViewerServiceElectron.test.ts
SeanLeRoy: It _might_ be the way that sinon is running against the other tests - so I'd try the sinon changes I suggested in other comments.

it might also be that the way `fs` is imported in this and in the actual usage is different so perhaps try unifying them

aswallace: Looked at the branch locally, agree that it's an issue with the way `fs` is imported. 
I was able to get the test to pass/fail by changing the fs import to `import fs from "fs"; `

packages/desktop/src/services/test/FileViewerServiceElectron.test.ts
tyler-foster: Done in 8c20038b034e59c10dc4284be039a40b071a5ec1

packages/desktop/src/services/test/FileViewerServiceElectron.test.ts
tyler-foster: Thanks you two. The issue did turn out to be the import, and changing the stub to use the sandbox did simplify the test.

I can the test to pass / fail depending on what I have `existsSync` return now, but unfortunately the unit test is still failing in the GitHub build 😢 . Haven't figured out why yet.

tyler-foster: Hey @SeanLeRoy @aswallace , it's me again. I have bad news! I made the changes you guys suggested previously, and while they helped with my local test setup / mocking situation, it looks like they didn't help get the unit test passing in CI. And [peeking at the failing test run](https://github.com/AllenInstitute/biofile-finder/actions/runs/13840327953/job/39312509313?pr=426), I can confidently say I don't have any clue what the issue is, nor how to recreate it, let alone fix it :(

Thoughts:
* If either of you have input on what could be wrong in the CI test, please let me know!
* The failing test is the one I added. Are we desperate to have this unit test? Is there a way we could re-write it such that it dodges whatever is making it fail? If we're going to ultimately get rid of the desktop version of BFF, do we care much about having great unit tests?
* We could completely abandon this idea (of opening the native file browser directly) and create a slightly less enjoyable feature that allows people to easily copy annotation values (e.g. local file path) to their clipboard (FWIW I have watched users struggle to do this). Then users could navigate to file paths by pasting them into their native file browser.

LMK if you guys have any input. Chris and I would like to get this feature (or something like it that makes navigating to files easier) out so we can finally say that multifile support is done.

aswallace: > If we're going to ultimately get rid of the desktop version of BFF, do we care much about having great unit tests?

I think this is a good point, and since we've already been trying to avoid investing too much time in desktop-only features I'd be on board with moving forward with this change minus the unit test

tyler-foster:
Closing this PR for a few reasons:

* Unit test continues to fail in CI
* Recent changes to the `FileDetails` type seem to have rendered these changes inert / unmergeable
* This was going to be a desktop-only feature, and rumor has it we want to deprecate the desktop version of the app in the longterm anyways.

An alternative idea would be to make it easier for users to copy annotation values. That would apply to more situations beyond just copying file paths, and would work in the web version of the app. Would be low prio work for me, so I wouldn't be making that change too soon, but it's something to consider.

PR #429, opened 2025-02-12
Feature/update dataset table metadata

aswallace:
## Context
New metadata properties + refactors of old code for our [Open Source Datasets page](https://bff.allencell.org/datasets).

#### Old code refactors:
May look like a lot of changes, but it's actually mostly just moving around & consolidating duplicate code.
- DatasetRow and DatasetDetails were both using the same copy-pasted functions to load datasets and redirect the browser. I moved these to their shared parent component.

#### New features:
- The 'Related publication' column defaulted to using the DOI as the link, but our new dataset uses a non-DOI link (and it's likely others will as well), so added a new prop and logic to handle
- Added a new 'Specific query' property & logic (rename welcome) in anticipation of allowing users to provide default query args that will load when the dataset is loaded

### DIFF
--- a/packages/core/entity/FileExplorerURL/index.ts
+++ b/packages/core/entity/FileExplorerURL/index.ts
@@ -13,9 +13,11 @@ export enum FileView {
     LARGE_THUMBNAIL = "3",
 }
 
+export const ACCEPTED_SOURCE_TYPES = ["csv", "json", "parquet"] as const;
+
 export interface Source {
     name: string;
-    type?: "csv" | "json" | "parquet";
+    type?: typeof ACCEPTED_SOURCE_TYPES[number];
     uri?: string | File;
 }
 
@@ -72,8 +74,11 @@ export const DEFAULT_AICS_FMS_QUERY: FileExplorerURLComponents = {
 export const getNameAndTypeFromSourceUrl = (dataSourceURL: string) => {
     const uriResource = dataSourceURL.substring(dataSourceURL.lastIndexOf("/") + 1).split("?")[0];
     const name = `${uriResource} (${new Date().toLocaleDateString()} ${new Date().toLocaleTimeString()})`;
-    let extensionGuess = uriResource.split(".").pop();
-    if (!(extensionGuess === "csv" || extensionGuess === "json" || extensionGuess === "parquet")) {
+    // Returns undefined if can't find a match
+    let extensionGuess = ACCEPTED_SOURCE_TYPES.find(
+        (validSourcetype) => validSourcetype === uriResource.split(".").pop()
+    );
+    if (!extensionGuess) {
         console.warn("Assuming the source is csv since no extension was recognized");
         extensionGuess = "csv";
     }
--- a/packages/web/src/components/OpenSourceDatasets/index.tsx
+++ b/packages/web/src/components/OpenSourceDatasets/index.tsx
@@ -1,14 +1,20 @@
 import * as React from "react";
-import { useDispatch } from "react-redux";
+import { useDispatch, useSelector } from "react-redux";
+import { useNavigate } from "react-router-dom";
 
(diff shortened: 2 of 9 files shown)


packages/web/src/components/OpenSourceDatasets/index.tsx
```diff
@@ -26,6 +35,40 @@ export default function OpenSourceDatasets() {
...
+        if (!datasetDetails) throw new Error("No dataset provided");
+
+        const dataSourceURL = datasetDetails.path;
+        const { name, extensionGuess } = getNameAndTypeFromSourceUrl(dataSourceURL);
+        const url = datasetDetails?.presetQuery;
+        openDatasetInApp(
+            datasetDetails.name,
+            {
+                name,
+                type: extensionGuess as "csv" | "json" | "parquet",
```
BrianWhitneyAI, line 65: What happens if the extension guess is not of type "csv" | "json" | "parquet"?  is that guaranteed? if its guaranteed than why is it a guess?

BrianWhitneyAI (approved):

packages/web/src/components/OpenSourceDatasets/index.tsx
aswallace: (not written by me so this may be slightly incorrect) this was just an formal typechecking guarantee. 

The function that produced `extensionGuess` (`getNameAndTypeFromSourceUrl`) did only return those options, and defaulted to "csv" if the extension didn't match any of those. However, it wasn't originally written in a way that *technically* guaranteed those return values (since extensionGuess was officially initialized as a generic `string` instead of type `"csv" | "json" | "parquet"`). I've rewritten the function  to explicitly type that return values in [a9141bd](https://github.com/AllenInstitute/biofile-finder/pull/429/commits/a9141bd1ebfff05fad8f3bfb662d7d45039d9e37)

packages/core/entity/FileExplorerURL/index.ts
```diff
@@ -72,8 +74,11 @@ export const DEFAULT_AICS_FMS_QUERY: FileExplorerURLComponents = {
 export const getNameAndTypeFromSourceUrl = (dataSourceURL: string) => {
     const uriResource = dataSourceURL.substring(dataSourceURL.lastIndexOf("/") + 1).split("?")[0];
     const name = `${uriResource} (${new Date().toLocaleDateString()} ${new Date().toLocaleTimeString()})`;
```
kmitcham, line 76: I don't understand the repeated date thing, but if its intentional, then it is probably OK?

kmitcham (approved):

packages/core/entity/FileExplorerURL/index.ts
aswallace: One is for date and one is for time; I suspect that the reason they're generated separately is that `toLocaleString()` returns them with `, ` in between and we just want a space apparently?

packages/web/src/components/OpenSourceDatasets/index.tsx
SeanLeRoy: It is implicit from the input we allow users to select - not ideal but helps narrow things downstream

packages/web/src/components/OpenSourceDatasets/index.tsx
```diff
@@ -26,6 +35,40 @@ export default function OpenSourceDatasets() {
         );
     }, [dispatch]);
 
+    const openDatasetInApp = (
+        datasetName: string,
+        source: Source,
+        url?: Partial<FileExplorerURLComponents>
+    ) => {
+        dispatch(
```
SeanLeRoy, line 43: ooo interesting so this lets the state parse out the components provided by the URL while using the react router to migrate us between pages? Smart! I like that we don't have to have an explicit link that forces everything to reload

SeanLeRoy (approved):

PR #431, opened 2025-02-13
restrict names for new annotations and update error messages

aswallace:
NOTE: I'm not sure why there are package.json/package-lock changes. Will look into it when I'm back

When users create new annotations ("fields" in the UI), we first want to check if the name is already in use and provide user feedback.
- For exact matches: show an error and prevent submission 
- For similar but non-exact: show a warning but allow submission 

When users click submit, we first try to create the annotation and if that succeeds, send the file metadata edit request
- Updates messages at bottom of modal to inform about annotation creation status

Styling/phrasing per UX discussions in [this Google doc](https://docs.google.com/document/d/16EfIcEzoSdg7UE7fPn88WQWtmh6fDCOVj9_XsD8J4go/edit?tab=t.0).

#### Field name overlap errors & warnings
Screenshot 2025-02-13 at 10 08 47 AM
Screenshot 2025-02-13 at 10 06 59 AM

#### Status message for annotation creation request
Warning message currently has a width restriction bc of nested components, but can change if needed
Screenshot 2025-02-12 at 4 31 17 PM
Screenshot 2025-02-13 at 10 07 22 AM

### DIFF
--- a/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
+++ b/packages/core/components/EditMetadata/NewAnnotationPathway.tsx
@@ -144,7 +144,7 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
         }
     };
 
-    const getFieldNameValidationErrors = (value: string) => {
+    const validateFieldName = (value: string) => {
         const matchingAnnotation = similarExistingFields.find(
             (annotation) =>
                 annotation.key.toLocaleLowerCase().trim() === value.toLocaleLowerCase().trim()
@@ -247,7 +247,7 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
                     [styles.textFieldError]: hasNameErrors,
                 })}
                 onChange={(ev, newValue) => onChangeAlphanumericField(ev, newValue)}
-                onGetErrorMessage={getFieldNameValidationErrors}
+                onGetErrorMessage={validateFieldName}
                 placeholder="Add a new field name..."
                 validateOnFocusOut
                 value={newFieldName}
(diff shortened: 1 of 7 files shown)


lynwilhelm:
Beautiful!

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
```diff
@@ -122,6 +144,52 @@ export default function NewAnnotationPathway(props: NewAnnotationProps) {
         }
     };
 
+    const getFieldNameValidationErrors = (value: string) => {
```
SeanLeRoy, line 147: This is named as a getter but has no return and instead does some side-effects.

SeanLeRoy (approved):
Awesome!

packages/core/components/EditMetadata/NewAnnotationPathway.tsx
aswallace: tentatively renamed to validateFieldName in [bc3f0b8](https://github.com/AllenInstitute/biofile-finder/pull/431/commits/bc3f0b898289c3fe9f93ca8a80e5121e7780b9b1)

kmitcham (approved):
How accessible are the descriptions of the similar annotations from the error message?  Can the user look up descriptions without exiting the window?

aswallace:
> How accessible are the descriptions of the similar annotations from the error message? Can the user look up descriptions without exiting the window?

Great point, currently they can't look them up without exiting. I agree that it would be useful for users to be able to see those, but it might take some workshopping to figure out how to incorporate that

Issue #282, opened 2024-10-15
Inconsistent Thumbnail rendering for  Zarrs 


BrianWhitneyAI:
When using the specimen_metadata.csv of mouse brains, a mix of the files render. Some always render, some sometimes render, and some never render. We added retry logic in #268 which stabilized the sometimes render files. However we are consistently reaching the following error: 

```
RenderZarrThumbnailURL.ts:128 Failed to render Zarr thumbnail: RangeError: Maximum call stack size exceeded
    at renderZarrThumbnailURL (RenderZarrThumbnailURL.ts:91)
    at async FileDetail.getPathToThumbnail (index.ts:184)
``` 

This indicates that we need to add some sort of smarter rendering policy for zarrs.

will-moore:
For thumbnailing from zarrs, I've recently created a library to help with this at https://github.com/will-moore/ome-zarr.js

This is very much a work in progress, but I'd be happy to get feedback if it appears useful?

It handles OME-Zarr v0.4 and v0.5 images.

This is code from the zarr thumbnail rendering at https://ome.github.io/ome2024-ngff-challenge/ which is pretty reliable except for images where the lowest resolution is massive!

PR #435, opened 2025-02-14
Thumbnail fixes

will-moore:
This might address some of the issues on #282.
Main fixes:

 - Avoids `Math.min(...u16data)` and `Math.max(...u16data)` as these were exceeding max call stack size.
 - Make sure the Z slice is valid where the size of Z is 1

Minor fix:
 - I added handling of missing multiscales.axes as these are not present in OME-Zarr v0.2 data (which I happened to be testing with). I don't know if you want to support this as there's not much of this version "in the wild". I didn't handle OME-Zarr v0.3 yet, where `axes` is `{"t", "c", "z", "y", "x"}`.
 
Suggested change:
 - Quite a lot of zarr data (e.g. generated by `bioformats2raw`) has the zarr image at `path/to/image.zarr/0` so the path doesn't end with `.zarr`. Hopefully not too many non-zarr images have `.zarr` in the path, so it shouldn't get many false positives.

### DIFF
--- a/packages/core/entity/FileDetail/RenderZarrThumbnailURL.ts
+++ b/packages/core/entity/FileDetail/RenderZarrThumbnailURL.ts
@@ -103,7 +103,14 @@ export async function renderZarrThumbnailURL(zarrUrl: string): Promise<string | 
                 // Determine a slice of the image that has the best chance at being a good thumbnail.
                 // X and Y will have full range, while Z will be the middle integer.
                 // Non-spatial dims will default to None.
-                const axes = transformAxes(multiscales[0].axes);
+                const defaultAxes = [
+                    { name: "t", type: "time" },
+                    { name: "c", type: "channel" },
+                    { name: "z", type: "space" },
+                    { name: "y", type: "space" },
+                    { name: "x", type: "space" },
+                ];
+                const axes = transformAxes(multiscales[0].axes || defaultAxes);
                 const zIndex = axes.findIndex((item) => item.name === "z");
                 if (zIndex !== -1) {
                     // if size Z is 1, use Math.floor so we get slice of 0
--- a/packages/core/entity/FileDetail/RenderZarrThumbnailURL.ts
+++ b/packages/core/entity/FileDetail/RenderZarrThumbnailURL.ts
@@ -106,7 +106,8 @@ export async function renderZarrThumbnailURL(zarrUrl: string): Promise<string | 
                 const axes = transformAxes(multiscales[0].axes);
                 const zIndex = axes.findIndex((item) => item.name === "z");
                 if (zIndex !== -1) {
-                    const zSliceIndex = Math.ceil(lowestResolution.shape[zIndex] / 2);
+                    // if size Z is 1, use Math.floor so we get slice of 0
+                    const zSliceIndex = Math.floor(lowestResolution.shape[zIndex] / 2);
                     axes[zIndex].value = zSliceIndex;
                 }
 
@@ -119,11 +120,17 @@ export async function renderZarrThumbnailURL(zarrUrl: string): Promise<string | 
                 const u16data = lowestResolutionView.data as Uint16Array;
 
                 // Normalize Data to improve image visibility.
-                const min = Math.min(...u16data);
-                const max = Math.max(...u16data);
+                let min = Infinity;
+                let max = -Infinity;
+                const dlength = u16data.length;
+                for (let i = 0; i < dlength; i++) {
+                    min = Math.min(min, u16data[i]);
+                    max = Math.max(max, u16data[i]);
+                }
                 const normalizedData = new Uint8Array(u16data.length);
                 for (let i = 0; i < u16data.length; i++) {
-                    normalizedData[i] = Math.round((255 * (u16data[i] - min)) / (max - min));
+                    // use bitwise shift to truncate decimal
+                    normalizedData[i] = ((255 * (u16data[i] - min)) / (max - min)) << 0;
                 }
 
                 // Build a canvas to put image data onto.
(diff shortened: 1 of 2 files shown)


packages/core/entity/FileDetail/RenderZarrThumbnailURL.ts
```diff
@@ -119,11 +127,17 @@ export async function renderZarrThumbnailURL(zarrUrl: string): Promise<string |
                 const u16data = lowestResolutionView.data as Uint16Array;
 
                 // Normalize Data to improve image visibility.
-                const min = Math.min(...u16data);
-                const max = Math.max(...u16data);
+                let min = Infinity;
```
BrianWhitneyAI, line 130: This could be written this way to be more javascripty: 
```                
const { min, max } = u16data.reduce(
    (acc, val) => ({
        min: Math.min(acc.min, val),
        max: Math.max(acc.max, val),
    }),
    { min: Infinity, max: -Infinity }
);
```

packages/core/entity/FileDetail/index.ts
```diff
@@ -262,7 +262,7 @@ export default class FileDetail {
             if (isFileRenderableImage) {
                 return this.downloadPath;
             }
-            if (this.path.endsWith(".zarr")) {
+            if (this.path.includes(".zarr")) {
```
BrianWhitneyAI, line 265: What situation does this improve the codebase? I have seen file paths with a trailing slash, but I dont believe those are readable by the dependent package anyway without some clipping.

packages/core/entity/FileDetail/RenderZarrThumbnailURL.ts
```diff
@@ -103,10 +103,18 @@ export async function renderZarrThumbnailURL(zarrUrl: string): Promise<string |
...
                 // Non-spatial dims will default to None.
-                const axes = transformAxes(multiscales[0].axes);
+                const defaultAxes = [
+                    { name: "t", type: "time" },
+                    { name: "c", type: "channel" },
+                    { name: "z", type: "space" },
+                    { name: "y", type: "space" },
+                    { name: "x", type: "space" },
+                ];
+                const axes = transformAxes(multiscales[0].axes || defaultAxes);
```
BrianWhitneyAI, line 113: Do you happen to have any test files that exhibit this behavior? would love to experiment with some!

BrianWhitneyAI (approved):
This looks really good to me! I tested it on the mouse brain dataset that the original issue was referencing and saw significant improvement! Thank you for contributing!!

packages/core/entity/FileDetail/index.ts
BrianWhitneyAI: ah I see """Quite a lot of zarr data (e.g. generated by bioformats2raw) has the zarr image at path/to/image.zarr/0 so the path doesn't end with .zarr. Hopefully not too many non-zarr images have .zarr in the path, so it shouldn't get many false positives.""" 

I think currently we expect users to provide the top level zarr path. I haven't considered reading in portions of the Zarr, `path/to/image.zarr/0` for example

SeanLeRoy: >Quite a lot of zarr data (e.g. generated by bioformats2raw) has the zarr image at path/to/image.zarr/0
Hmm this is good to know - unfortunate from a consistency standpoint, but this seems fine to include for now

SeanLeRoy (approved):

packages/core/entity/FileDetail/RenderZarrThumbnailURL.ts
toloudis: As an old C++ programmer, I can't resist.  You can do performance tests with these functional constructs that show they are much much slower than a simple for loop with integer index. For times when performance matters, like iterating over long typedarrays.

will-moore: Yes, I didn't test it here but that was my understanding. There is some overhead of creating and calling a function for every pixel. Would be interesting to check...

will-moore:
There's a bunch of sample data at https://idr.github.io/ome-ngff-samples/ with different OME-Zarr versions.

Some are Plates and some are `bioformats2raw` layout.

An example of v0.2 data in bioformats2raw layout, viewed in the ngff-validator shows 'series' of 3 images nested below the `.zarr` https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.2/idr0070A/9838562.zarr

Plates are another case where individual images are nested much below the `.zarr`. E.g. https://ome.github.io/ome-ngff-validator/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0056B/7361.zarr shows the Plate... but the image at `A1` field `0` is nested at https://hms-dbmi.github.io/vizarr/?source=https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0056B/7361.zarr/A/1/0 (viewed in vizarr)

will-moore:
@BrianWhitneyAI Is that Mouse Brain dataset publicly available? It would be nice to look and see why some of the other thumbnails are still failing.

BrianWhitneyAI:
@will-moore This was a CSV that was generated by the Allen Institute for Neural Dynamics, It came up in a collaboration meeting. To my knowledge it is not published, so heres the CSV for it. This was the first place that we saw some of the performance issue you fixed. I believe that the other images are failing due to path format. We haven't implemented a lot of leeway around zarr path naming convention so these files are failing to render.
[SpecimenMetadata2 (1).csv](https://github.com/user-attachments/files/18851831/SpecimenMetadata2.1.csv)

PR #437, opened 2025-02-14
Bugfix/empty csv for manually selected files

### LINKED ISSUE #415: Bug: saving CSV of files
From Jianxu:
When I select a consecutive list of files (e.g., click while holding Shift), and then select “Save Metadata As … CSV”, everything is fine. All information is saved properly.
When I select a random list of files (e.g, click while holding Command / Ctrl), and then select “Save Metadata As … CSV”, then only the column names are downloaded. It is an empty table …

BrianWhitneyAI:
### Description 

The purpose of this pr is to resolve #415. We have a situation with imported data where the query we build to download a csv does not account for single selected files ie. (CTRL + Select) this is because we pass it a list of index ranges. which might look like:
```
                    indexRanges: [
                        { start: 0, end: 0 },
                        { start: 2, end: 2 },
                    ],
```
When we build a query it does not register these ranges since they're not inclusive. However if we pass it `indexRanges: [{ start: 0, end: 2 }]` it works correctly. 

### Changes

- Add conditional to handle single index files 
- Seperate the download method which has gotten quite heavy
- Add tests for this case (Comparing the query)

### DIFF
--- a/packages/core/services/FileService/DatabaseFileService/index.ts
+++ b/packages/core/services/FileService/DatabaseFileService/index.ts
@@ -202,6 +202,8 @@ export default class DatabaseFileService implements FileService {
         sql: string,
         format: "csv" | "json" | "parquet"
     ): Promise<DownloadResult> {
+        // If the file system is accessible we can just have DuckDB write the
+        // output query directly to the system rather than to a buffer then the file
         if (this.downloadService.isFileSystemAccessible) {
             const downloadDir = await this.downloadService.getDefaultDownloadDirectory();
             const separator = navigator.userAgent.toLowerCase().includes("windows") ? "\\" : "/";
--- a/packages/core/services/FileService/DatabaseFileService/index.ts
+++ b/packages/core/services/FileService/DatabaseFileService/index.ts
@@ -130,71 +130,38 @@ export default class DatabaseFileService implements FileService {
             .select(annotations.map((annotation) => `"${annotation}"`).join(", "))
             .from(this.dataSourceNames);
 
+        this.applySelectionFilters(sqlBuilder, selections);
+
+        return this.handleFileDownload(sqlBuilder.toSQL(), format);
+    }
+
+    /**
+     * Processes selections and applies WHERE clause directly to the SQLBuilder.
+     */
+    private applySelectionFilters(sqlBuilder: SQLBuilder, selections: Selection[]): void {
         const uidConditions: string[] = [];
 
         for (const selection of selections) {
             const selectionUIDs: string[] = [];
 
             for (const indexRange of selection.indexRanges) {
+                const subQuery = new SQLBuilder()
+                    .select(`${DatabaseService.HIDDEN_UID_ANNOTATION}`)
+                    .from(this.dataSourceNames)
+                    .offset(indexRange.start)
+                    .limit(indexRange.end - indexRange.start + 1);
+
+                this.applyFiltersAndSorting(subQuery, selection);
+
                 if (indexRange.start === indexRange.end) {
-                    // Single index selection case, fetch the UID directly
-                    const uidQuery = new SQLBuilder()
-                        .select(`${DatabaseService.HIDDEN_UID_ANNOTATION}`)
-                        .from(this.dataSourceNames)
-                        .offset(indexRange.start)
-                        .limit(1); // Select exactly one row
-
-                    if (!isEmpty(selection.filters)) {
-                        uidQuery.where(
-                            Object.entries(selection.filters)
-                                .flatMap(([column, values]) =>
-                                    values.map((v) => SQLBuilder.regexMatchValueInList(column, v))
-                                )
-                                .join(") OR (")
-                        );
-                    }
-
-                    if (selection.sort) {
-                        uidQuery.orderBy(
-                            `"${selection.sort.annotationName}" ${
-                                selection.sort.ascending ? "ASC" : "DESC"
-                            }`
-                        );
-                    }
-
-                    selectionUIDs.push(`(${uidQuery.toSQL()})`);
+                    selectionUIDs.push(`(${subQuery.toSQL()})`);
                 } else {
-                    // Normal range selection
-                    const subQuery = new SQLBuilder()
-                        .select(`${DatabaseService.HIDDEN_UID_ANNOTATION}`)
-                        .from(this.dataSourceNames)
-                        .offset(indexRange.start)
-                        .limit(indexRange.end - indexRange.start + 1);
-
-                    if (!isEmpty(selection.filters)) {
-                        subQuery.where(
-                            Object.entries(selection.filters)
-                                .flatMap(([column, values]) =>
-                                    values.map((v) => SQLBuilder.regexMatchValueInList(column, v))
-                                )
-                                .join(") OR (")
-                        );
-                    }
-                    if (selection.sort) {
-                        subQuery.orderBy(
-                            `"${selection.sort.annotationName}" ${
-                                selection.sort.ascending ? "ASC" : "DESC"
-                            }`
-                        );
-                    }
-
                     uidConditions.push(
                         `${DatabaseService.HIDDEN_UID_ANNOTATION} IN (${subQuery.toSQL()})`
                     );
                 }
             }
 
-            // Add collected UIDs from single-index selections
             if (selectionUIDs.length > 0) {
                 uidConditions.push(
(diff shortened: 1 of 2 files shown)


packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -130,55 +130,95 @@ export default class DatabaseFileService implements FileService {
...
+     * Applies filters and sorting to a query.
+     */
+    private applyFiltersAndSorting(query: SQLBuilder, selection: Selection): void {
+        if (!isEmpty(selection.filters)) {
+            query.where(
+                Object.entries(selection.filters)
+                    .flatMap(([column, values]) =>
+                        values.map((v) => SQLBuilder.regexMatchValueInList(column, v))
+                    )
+                    .join(") OR (")
```
kmitcham, line 187: wow, that is correct syntax, but it looks so wrong.

packages/core/services/FileService/DatabaseFileService/test/DatabaseFileService.test.ts
```diff
@@ -100,4 +101,65 @@ describe("DatabaseFileService", () => {
...
+            return sql.replace(/\s+/g, " ").trim();
+        }
+
+        it("correctly modifies SQLBuilder for single index selections (CTRL selection)", () => {
+            // Arrange
+            const selections = [
+                {
+                    indexRanges: [
+                        { start: 0, end: 0 },
+                        { start: 2, end: 2 },
```
kmitcham, line 131: would it be worthwhile to have a test case for 
                    indexRanges: [
                        { start: 0, end: 0 },
                        { start: 2, end: 6 },
                        ]
 mixing the single and the ranges?  Probably redundant.

kmitcham (approved):

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -130,55 +130,95 @@ export default class DatabaseFileService implements FileService {
...
-            selection.indexRanges.forEach((indexRange) => {
+        this.applySelectionFilters(sqlBuilder, selections);
+
+        return this.handleFileDownload(sqlBuilder.toSQL(), format);
+    }
+
+    /**
+     * Processes selections and applies WHERE clause directly to the SQLBuilder.
+     */
+    private applySelectionFilters(sqlBuilder: SQLBuilder, selections: Selection[]): void {
```
SeanLeRoy, line 141: This looks like it should be a static method. Moreover, it feels like this belongs in the `SQLBuilder` domain since it is scoped as a setter-like side-effect function

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -130,55 +130,95 @@ export default class DatabaseFileService implements FileService {
...
+
+        if (uidConditions.length > 0) {
+            sqlBuilder.whereOr(uidConditions.join(" OR "));
+        }
+    }
 
+    /**
+     * Applies filters and sorting to a query.
+     */
+    private applyFiltersAndSorting(query: SQLBuilder, selection: Selection): void {
```
SeanLeRoy, line 180: Static as well

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -130,55 +130,95 @@ export default class DatabaseFileService implements FileService {
...
-                    );
-                }
-                if (selection.sort) {
-                    subQuery.orderBy(
-                        `"${selection.sort.annotationName}" ${
-                            selection.sort.ascending ? "ASC" : "DESC"
-                        }`
+                this.applyFiltersAndSorting(subQuery, selection);
+
+                if (indexRange.start === indexRange.end) {
```
SeanLeRoy, line 156: So this is the real change right? Can you explain what is happening? I don't think I get what separation `selectionUIDs` & `uidConditions` are supposed to have here. It looks like `selectionUIDs` is a list of conditions as well based on `subQuery.toSQL()`

BrianWhitneyAI: Selection IDs are ranges while UidConditions are unique ID selections (when I write that out I realize I should rename them). The layout of this is largely a result of my debugging process and could probably be simplified. Our Issue was in how we are connecting our  single selection subqueries to the larger query.  originally we had 
```
            SELECT "File Name", "File ID", "Gene"
            FROM "Variance Paper Dataset"
            WHERE (hidden_bff_uid IN (
            
            SELECT hidden_bff_uid
            FROM "Variance Paper Dataset"
            
            
            OFFSET 0
            LIMIT 1
        )) AND (hidden_bff_uid IN (
            
            SELECT hidden_bff_uid
            FROM "Variance Paper Dataset"
            
            
            OFFSET 2
            LIMIT 1
        )) AND (hidden_bff_uid IN (
            
            SELECT hidden_bff_uid
            FROM "Variance Paper Dataset"
            
            
            OFFSET 4
            LIMIT 1
        ))
```
This used AND to add ranges which resulted in an empty csv. We can separate the subqueries out so at first we only add the range selections and then afterwards add the list of single selections giving us something like 
```
            SELECT "File Name", "File ID", "Gene"
            FROM "Variance Paper Dataset"
            WHERE (hidden_bff_uid IN (
            
            SELECT hidden_bff_uid
            FROM "Variance Paper Dataset"
            
            
            OFFSET 0
            LIMIT 6
        ) OR hidden_bff_uid IN (
            
            SELECT hidden_bff_uid
            FROM "Variance Paper Dataset"
            
            
            OFFSET 18
            LIMIT 5
        ) OR hidden_bff_uid IN ((
            
            SELECT hidden_bff_uid
            FROM "Variance Paper Dataset"
            
            
            OFFSET 9
            LIMIT 1
        ), (
            
            SELECT hidden_bff_uid
            FROM "Variance Paper Dataset"
            
            
            OFFSET 11
            LIMIT 1
        ), (
            
            SELECT hidden_bff_uid
            FROM "Variance Paper Dataset"
            
            
            OFFSET 13
            LIMIT 1
        ), (
            
            SELECT hidden_bff_uid
            FROM "Variance Paper Dataset"
            
            
            OFFSET 15
            LIMIT 1
        )))
```

using or for ranges and comma for single selections. There is probably a better way to do this but this was the sort of first way I could submit successful single selection queries.

BrianWhitneyAI: Need to think on this some more I think we can probably just add each individual query to the uidConditions and join them all with "or" rather than this separation of handling.

BrianWhitneyAI: Attempted to update in [94e06d3](https://github.com/AllenInstitute/biofile-finder/pull/437/commits/94e06d3c39d9c7506c6094335070803b6b8d6a66). In doing so I noticed that whereOr is not really implemented.

packages/core/services/FileService/DatabaseFileService/index.ts
BrianWhitneyAI: moved to static in [94e06d3](https://github.com/AllenInstitute/biofile-finder/pull/437/commits/94e06d3c39d9c7506c6094335070803b6b8d6a66)

packages/core/services/FileService/DatabaseFileService/test/DatabaseFileService.test.ts
BrianWhitneyAI: The original bug was only failing with single range selections such as { start: 0, end: 0 }. I think there is some value in splitting up the tests so that each one covers a different case here.

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -130,55 +130,81 @@ export default class DatabaseFileService implements FileService {
...
+
+    /**
+     * Processes selections and applies WHERE clause directly to the SQLBuilder.
+     */
+    public static applySelectionFilters(
+        sqlBuilder: SQLBuilder,
+        selections: Selection[],
+        dataSourceNames: string[]
+    ): void {
+        const subQuerys: string[] = [];
```
SeanLeRoy, line 146: (nit) subQueries

SeanLeRoy (approved):

packages/core/services/FileService/DatabaseFileService/index.ts
BrianWhitneyAI: whoops resolved in [089987b](https://github.com/AllenInstitute/biofile-finder/pull/437/commits/089987b35528093503a7ed3ed12286da367f3234)

PR #438, opened 2025-02-17
feature/custom-mac-builds

### LINKED ISSUE #173: Unable to build universal deploy for MacOS 🐛

BrianWhitneyAI:
### Description 

The purpose of this PR is to resolve #173 and add a way to build Intel chipset versions. This pr just adds it to the manual build for now. Once its tested we can add it to our automatic builds/releases.

### DIFF
--- a/.github/workflows/manual-build.yml
+++ b/.github/workflows/manual-build.yml
@@ -14,6 +14,7 @@ on:
       mac_arch:
         description: "Target Mac Architecture (only used for macos-latest)"
         required: false
+        default: arm64
         type: choice
         options:
           - x64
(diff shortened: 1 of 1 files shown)


.github/workflows/manual-build.yml
```diff
@@ -33,13 +40,14 @@ jobs:
           github_token: ${{ secrets.github_token }}
           package_root: "packages/desktop"
           release: false  # No release, just build
+          Invalid workflow file: .github/workflows/manual-build.yml#L43
```
pgarrison, line 43: Is this a typo? If not I think it needs a comment explaining it

pgarrison (approved):

aswallace (approved):

PR #446, opened 2025-02-21
Feature/Allow expanding or collapsing all folders

### LINKED ISSUE #334: Improve comparing multiple file lists
### Description
- Design, Discovery, & Research
- Unable to easily compare between files when grouped, requires many clicks to compare two groupings of files. In the below example to compare these three file lists this required ~13 clicks. AIBS would like to do this with dozens of file lists. We also talked about having a breadcrumb style presentation of the groupings when all expanded to save space and more compactly show the difference between groups.
Screen Shot 2024-11-18 at 2 27 01 PM

***
_Below to be filled out by UX during kickoff Q&A to the best of current understanding_
### Size
S
### Scope
_What will we do? What will we NOT do?_
### User needs statement(s)
As a user I want to be able to compare 5+ lists of files without having to click 10+ times.
### Additional notes
_Knowns/unknowns, stakeholders, time/dependency considerations, deliverables_

aswallace:
## Context
AIBS requested the ability to expand (or collapse) ALL the folders in the directory tree at once. This adds that capability. 
resolves #334 

## Changes
- For folders ONLY (e.g., not files), adds menu options to expand all or collapse all
- Adds redux logic to calculate all file paths in the directory tree
- Adds logic to trigger the loading state, since can take time to find all paths

## Testing
- Added unit tests for new redux logic
- Tested manually on dev builds with multiple groupings applied

## Screenshots
Menu (web)
image

Loading state (uses existing UX)
image

### DIFF
--- a/packages/core/components/StatusMessage/StatusMessage.module.css
+++ b/packages/core/components/StatusMessage/StatusMessage.module.css
@@ -17,7 +17,7 @@
     display: none;
 }
 
-.centeringParent {
+.centering-parent {
     display: flex;
     font-size: var(--l-paragraph-size);
     font-weight: 500;
--- a/packages/core/hooks/useFileAccessContextMenu.ts
+++ b/packages/core/hooks/useFileAccessContextMenu.ts
@@ -14,7 +14,7 @@ import { interaction, selection } from "../state";
  * previously saved applications. Can be supplied an array of filters to use
  * to find files to access instead of the currently selected files.
  */
-export default (filters?: FileFilter[], onDismiss?: () => void, isFolder?: boolean) => {
+export default (folderFilters?: FileFilter[], onDismiss?: () => void) => {
     const dispatch = useDispatch();
     const isOnWeb = useSelector(interaction.selectors.isOnWeb);
     const fileSelection = useSelector(selection.selectors.getFileSelection);
@@ -22,7 +22,7 @@ export default (filters?: FileFilter[], onDismiss?: () => void, isFolder?: boole
 
     const [fileDetails, setFileDetails] = React.useState<FileDetail>();
 
-    const openWithSubMenuItems = useOpenWithMenuItems(fileDetails, filters);
+    const openWithSubMenuItems = useOpenWithMenuItems(fileDetails, folderFilters);
 
     fileSelection.fetchFocusedItemDetails().then((fileDetails) => {
         setFileDetails(fileDetails);
@@ -33,7 +33,7 @@ export default (filters?: FileFilter[], onDismiss?: () => void, isFolder?: boole
             evt.preventDefault();
 
             const contextMenuItems: IContextualMenuItem[] = [
-                ...(!isFolder
+                ...(!folderFilters
                     ? []
                     : [
                           {
@@ -67,10 +67,10 @@ export default (filters?: FileFilter[], onDismiss?: () => void, isFolder?: boole
                               iconProps: {
                                   iconName: "OpenInNewWindow",
                               },
-                              disabled: !filters && fileSelection.count() === 0,
+                              disabled: !folderFilters && fileSelection.count() === 0,
                               onClick() {
-                                  if (filters) {
-                                      dispatch(interaction.actions.openWithDefault(filters));
+                                  if (folderFilters) {
+                                      dispatch(interaction.actions.openWithDefault(folderFilters));
                                   } else if (fileDetails) {
                                       dispatch(
                                           interaction.actions.openWithDefault(undefined, [
@@ -84,7 +84,7 @@ export default (filters?: FileFilter[], onDismiss?: () => void, isFolder?: boole
                 {
                     key: "open-with",
                     text: "Open with",
-                    disabled: !filters && fileSelection.count() === 0,
+                    disabled: !folderFilters && fileSelection.count() === 0,
                     iconProps: {
                         iconName: "OpenInNewWindow",
                     },
@@ -95,7 +95,7 @@ export default (filters?: FileFilter[], onDismiss?: () => void, isFolder?: boole
                 {
                     key: "save-as",
                     text: "Save metadata as",
-                    disabled: !filters && fileSelection.count() === 0,
+                    disabled: !folderFilters && fileSelection.count() === 0,
                     iconProps: {
                         iconName: "Saveas",
                     },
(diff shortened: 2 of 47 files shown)


packages/core/components/StatusMessage/StatusMessage.module.css
```diff
@@ -17,7 +17,7 @@
     display: none;
 }
 
-.centeringParent {
+.centering-parent {
```
kmitcham, line 20: should this file change to match:
packages/core/components/StatusMessage/StatusMessage.module.css:.centeringParent {

packages/core/state/selection/actions.ts
```diff
@@ -515,6 +515,60 @@ export function toggleFileFolderCollapse(fileFolder: FileFolder): ToggleFileFold
...
+    return {
+        type: COLLAPSE_ALL_FILE_FOLDERS,
+    };
+}
+
+/**
+ * EXPAND_ALL_FILE_FOLDERS
+ * Intention to toggle the given file folder's collapsed state
+ */
+export const EXPAND_ALL_FILE_FOLDERS = makeConstant(STATE_BRANCH_NAME, "expand-all-file-folders");
```
kmitcham, line 541: nit: should this match the format/indentation of line 522?

kmitcham (approved):

packages/core/state/selection/actions.ts
aswallace: I believe this was done by our linter; e.g., it only performs the line break/indentation when we pass some column width

packages/core/components/StatusMessage/StatusMessage.module.css
aswallace: I'm not quite sure if I understand the suggestion

SeanLeRoy: I think Kevin is referencing that in the component it is "centeringParent" if so - the reason this is a different casing has to do with how the css module is loaded in. In JS/TS the css module names need to be the camel case version of whatever this case is (hyphen-case?)

packages/core/hooks/useFileAccessContextMenu.ts
```diff
@@ -14,7 +14,7 @@ import { interaction, selection } from "../state";
  * previously saved applications. Can be supplied an array of filters to use
  * to find files to access instead of the currently selected files.
  */
-export default (filters?: FileFilter[], onDismiss?: () => void) => {
+export default (filters?: FileFilter[], onDismiss?: () => void, isFolder?: boolean) => {
```
SeanLeRoy, line 17: Is `isFolder` `true` in all the same spots there are non-`undefined` `filters`? I'm wondering if that could be the heuristic and if so should we rename `filters` to be `folderFilters` & remove `isFolder`?

packages/core/state/selection/actions.ts
```diff
@@ -515,6 +515,60 @@ export function toggleFileFolderCollapse(fileFolder: FileFolder): ToggleFileFold
...
+ * Differentiated from isLoading in directory hierarchy node sub-states
+ */
+export const SET_IS_LOADING_FULL_TREE = makeConstant(STATE_BRANCH_NAME, "set-is-loading-full-tree");
+
+export interface SetIsLoadingFullTreeAction {
+    payload: boolean;
+    type: string;
+}
+
+export function setIsLoadingFullTree(isLoadingFullTree: boolean): SetIsLoadingFullTreeAction {
```
SeanLeRoy, line 565: I'm not sure I understand why it is necessary to know if the full tree is loading vs the node?

packages/core/state/selection/logics.ts
```diff
@@ -340,6 +344,53 @@ const toggleFileFolderCollapse = createLogic({
...
+                if (!!hierarchy.length && pathSoFar.length === hierarchy.length - 1) continue;
+
+                const childNodes = await annotationService.fetchHierarchyValuesUnderPath(
+                    hierarchy,
+                    [...pathSoFar, value],
+                    selectedFileFilters
+                );
+                if (childNodes.length) {
+                    // Not a leaf
+                    unpackAllFileFolders(childNodes, [...pathSoFar, value]);
```
SeanLeRoy, line 375: What is the performance of this looking like? I wonder if this should at every level dispatch a `setOpenFileFolders` action to save the folders it has found so far. However, that would interfere with the `isLoadingFullTree` flag you have (which I asked about in another comment).

packages/core/state/selection/logics.ts
```diff
@@ -340,6 +344,53 @@ const toggleFileFolderCollapse = createLogic({
...
+                hierarchy,
+                selectedFileFilters
+            );
+            await unpackAllFileFolders(rootHierarchyValues, []);
+        }
+        dispatch(setOpenFileFolders(fileFoldersToOpen));
+        dispatch(setIsLoadingFullTree(false) as AnyAction);
+        done();
+    },
+    type: [COLLAPSE_ALL_FILE_FOLDERS, EXPAND_ALL_FILE_FOLDERS],
```
SeanLeRoy, line 391: I think adding `COLLAPSE_ALL_FILE_FOLDERS` has complicated this logic and is unnecessary because unlike `EXPAND_ALL_FILE_FOLDERS` there is no async action that needs to transpire. Instead, for `COLLAPSE_ALL_FILE_FOLDERS` you can just have a reducer that sets `openFileFolders: []` in the state.

packages/core/state/selection/logics.ts
aswallace: My tests were generally taking less than 30 seconds, but I can see it taking longer. Will try this

packages/core/state/selection/logics.ts
aswallace: resolved in [c16c1ae](https://github.com/AllenInstitute/biofile-finder/pull/446/commits/c16c1ae522161412502360a675bab15dbacbf44f)

packages/core/state/selection/logics.ts
aswallace: version of this done in [c16c1ae](https://github.com/AllenInstitute/biofile-finder/pull/446/commits/c16c1ae522161412502360a675bab15dbacbf44f)

packages/core/state/selection/actions.ts
aswallace: removed in [c16c1ae](https://github.com/AllenInstitute/biofile-finder/pull/446/commits/c16c1ae522161412502360a675bab15dbacbf44f)

packages/core/hooks/useFileAccessContextMenu.ts
aswallace: refactored in [aceac39](https://github.com/AllenInstitute/biofile-finder/pull/446/commits/aceac39cfceb07295c2cc3aa1358dd1fd161957d)

SeanLeRoy (approved):

PR #447, opened 2025-02-24
bugfix/renderer-crash

### LINKED ISSUE #439: Desktop app crashes on refresh
Steps to reproduce
* Open BFF desktop app (I'm on 8.1.1)
* On a Mac, hit `Cmd+R` (`Ctrl+R` on Windows, I believe)
* Application crashes (blank white screen)

So far I have been able to reproduce this with 100% consistency.

BrianWhitneyAI:
### Description 

The purpose of this PR is to resolve #439 and fix the broken reload. Currently if you reload the app on desktop it'll crash but another reload will fix it. This PR just adds some retry logic on the renderer. I could not identify the exact problem but it seems to be an access issue with DuckDB.

### DIFF
--- a/packages/desktop/src/main/index.ts
+++ b/packages/desktop/src/main/index.ts
@@ -26,6 +26,10 @@ app.setAppUserModelId("org.aics.alleninstitute.fileexplorer");
 // Prevent window from being garbage collected
 let mainWindow: BrowserWindow | undefined;
 
+// Maximum number of renderer reload attempts
+const MAX_RELOAD_ATTEMPTS = 3;
+let reloadAttempts = 0;
+
 // register handlers called via ipc between renderer and main
 const registerIpcHandlers = () => {
     ExecutionEnvServicElectron.registerIpcHandlers();
@@ -49,7 +53,26 @@ const createMainWindow = () => {
         width: 1200,
     });
 
+    // Listen for renderer crashes and attempt a reload.
+    mainWindow.webContents.on("crashed", (event, killed) => {
+        console.error("Renderer process crashed. Killed:", killed);
+        if (reloadAttempts < MAX_RELOAD_ATTEMPTS) {
+            reloadAttempts++;
+            console.log(`Reloading renderer... attempt ${reloadAttempts}/${MAX_RELOAD_ATTEMPTS}`);
+            setTimeout(() => {
+                if (mainWindow) {
+                    mainWindow.reload();
+                }
+            }, 1000);
+        } else {
+            console.error("Maximum reload attempts reached. Quitting application.");
+            app.quit();
+        }
+    });
+
     mainWindow.once("ready-to-show", () => {
+        // Reset reload counter once the renderer is successfully loaded.
+        reloadAttempts = 0;
         if (mainWindow) {
             mainWindow.show();
         }
--- a/packages/desktop/src/main/index.ts
+++ b/packages/desktop/src/main/index.ts
@@ -26,9 +26,8 @@ app.setAppUserModelId("org.aics.alleninstitute.fileexplorer");
 // Prevent window from being garbage collected
 let mainWindow: BrowserWindow | undefined;
 
-// Maximum number of renderer reload attempts
-const MAX_RELOAD_ATTEMPTS = 3;
-let reloadAttempts = 0;
+// Track if a reload attempt has already been made
+let hasReloaded = false;
 
 // register handlers called via ipc between renderer and main
 const registerIpcHandlers = () => {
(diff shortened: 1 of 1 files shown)


packages/desktop/src/main/index.ts
```diff
@@ -49,7 +53,26 @@ const createMainWindow = () => {
...
+    mainWindow.webContents.on("crashed", (event, killed) => {
+        console.error("Renderer process crashed. Killed:", killed);
+        if (reloadAttempts < MAX_RELOAD_ATTEMPTS) {
+            reloadAttempts++;
+            console.log(`Reloading renderer... attempt ${reloadAttempts}/${MAX_RELOAD_ATTEMPTS}`);
+            setTimeout(() => {
+                if (mainWindow) {
+                    mainWindow.reload();
+                }
+            }, 1000);
```
BrianWhitneyAI, line 66: maybe we want less than a second here?

packages/desktop/src/main/index.ts
```diff
@@ -49,7 +53,26 @@ const createMainWindow = () => {
...
+        } else {
+            console.error("Maximum reload attempts reached. Quitting application.");
+            app.quit();
+        }
+    });
+
     mainWindow.once("ready-to-show", () => {
+        // Reset reload counter once the renderer is successfully loaded.
+        reloadAttempts = 0;
```
aswallace, line 75: This does not appear to be resetting correctly. I tried just resetting the app three times in succession (each time waiting until it loaded successfully) and the count keeps adding instead of going back to zero

packages/desktop/src/main/index.ts
```diff
@@ -26,6 +26,10 @@ app.setAppUserModelId("org.aics.alleninstitute.fileexplorer");
 // Prevent window from being garbage collected
 let mainWindow: BrowserWindow | undefined;
 
+// Maximum number of renderer reload attempts
+const MAX_RELOAD_ATTEMPTS = 3;
```
aswallace, line 30: Why specifically 3? If we know that for standard (non broken) refreshes it only takes 1 re-reload to work, for any other condition up to 3 probably means something is really broken and we just want it to error?

aswallace (changes requested):

packages/desktop/src/main/index.ts
BrianWhitneyAI: no real reason, 3 is just a standard I've done before for retries

packages/desktop/src/main/index.ts
```diff
@@ -49,7 +53,26 @@ const createMainWindow = () => {
...
+            reloadAttempts++;
+            console.log(`Reloading renderer... attempt ${reloadAttempts}/${MAX_RELOAD_ATTEMPTS}`);
+            setTimeout(() => {
+                if (mainWindow) {
+                    mainWindow.reload();
+                }
+            }, 1000);
+        } else {
+            console.error("Maximum reload attempts reached. Quitting application.");
+            app.quit();
```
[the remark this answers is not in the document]
BrianWhitneyAI: ah yes agreed

packages/desktop/src/main/index.ts
BrianWhitneyAI: hmm ok I will look into this

packages/desktop/src/main/index.ts
SeanLeRoy: I agree lets just fail out after one. Are you seeing it ever continue to fail after one?

packages/desktop/src/main/index.ts
```diff
@@ -49,7 +53,26 @@ const createMainWindow = () => {
         width: 1200,
     });
 
+    // Listen for renderer crashes and attempt a reload.
+    mainWindow.webContents.on("crashed", (event, killed) => {
```
SeanLeRoy, line 57: kinda crazy `crashed` is an event

packages/desktop/src/main/index.ts
SeanLeRoy: Hmm why wait at all?

packages/desktop/src/main/index.ts
SeanLeRoy: btw if/once you convert it to 1 retry i'd make this a boolean flag rather than a count for readability.

packages/desktop/src/main/index.ts
BrianWhitneyAI: The idea was to account for async tasks that might not have finished/ happened in the wrong order.

packages/desktop/src/main/index.ts
BrianWhitneyAI: yeah I'm thanking chat gpt on finding that one. Couldn't set up my own event correctly for this case.

packages/desktop/src/main/index.ts
BrianWhitneyAI: resolved in [45ea20a](https://github.com/AllenInstitute/biofile-finder/pull/447/commits/45ea20a5e2e2f88bd759b9fa4f2922928e7ee2a8)

packages/desktop/src/main/index.ts
BrianWhitneyAI: resolved in [45ea20a](https://github.com/AllenInstitute/biofile-finder/pull/447/commits/45ea20a5e2e2f88bd759b9fa4f2922928e7ee2a8) just use the 1did-finish-load` instead

packages/desktop/src/main/index.ts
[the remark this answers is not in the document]
BrianWhitneyAI: resolved in [45ea20a](https://github.com/AllenInstitute/biofile-finder/pull/447/commits/45ea20a5e2e2f88bd759b9fa4f2922928e7ee2a8)

packages/desktop/src/main/index.ts
SeanLeRoy: What tasks? If we have crashed can they still be running?

SeanLeRoy (approved):

aswallace (approved):

PR #453, opened 2025-02-26
New download page

SeanDuHare:
New download page.

TODO:
* Test it out
* Do bare minimum styling
* Remove old GH page version from repo
* Remove old GH page from repo settings

### DIFF
--- a/packages/core/hooks/useHelpOptions/index.ts
+++ b/packages/core/hooks/useHelpOptions/index.ts
@@ -117,17 +117,13 @@ export default function useHelpOptions(
                       },
                   },
               ]),
-        ...(isOnWeb
-            ? []
-            : [
-                  {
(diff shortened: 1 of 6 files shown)


packages/core/hooks/useHelpOptions/index.ts
```diff
@@ -117,17 +117,13 @@ export default function useHelpOptions(
...
-            : [
-                  {
-                      key: "download-newest-version",
-                      text: "Download Newest Version",
-                      title: "Opens the BioFile Finder download page in a new window",
-                      href: "https://alleninstitute.github.io/biofile-finder/",
-                      target: "_blank",
-                  },
-              ]),
+        {
```
SeanLeRoy, line 120: This needs to be hidden for non-aics uesrs

PR #455, opened 2025-02-27
feature/robust-cache-messages

### LINKED ISSUE #417: Copy to VAST success message is misleading when extending expiration dates
Steps to reproduce:
* Select a file that is already on the VAST
* Right click -> Copy to VAST
* Check that the file is going to have its expiration date extended
* Click confirm
* Actual behavior: Success message says that "files were successfully queued for download" (see screenshot)
* Expected behavior: Success message says that the VAST reservation was extended

Image

It's a little less clear what should happen when some files are queued for download and some have their expiration extended. Perhaps the modal just mentions the ones queued for download? E.g., if 10 files are selected and 4 need to be downloaded, the text is "4 out of 4 files were successfully queued for download..."

BrianWhitneyAI:
### Description 
The purpose of this PR is to resolve #417 and clear up some misleading reporting about the cache action. This PR separates the net success message into two separate messages for the two condition (while keeping messages separate). it also handles plural/singular cases.

### DIFF
--- a/packages/core/state/interaction/logics.ts
+++ b/packages/core/state/interaction/logics.ts
@@ -616,7 +616,8 @@ const copyFilesLogic = createLogic({
                 return;
             }
 
-            const successfulFiles: string[] = [];
+            const newlyCachedFiles: string[] = [];
+            const extendedExpirationFiles: string[] = [];
             const failedFiles: string[] = [];
 
             Object.entries(cacheStatuses).forEach(([fileId, status]) => {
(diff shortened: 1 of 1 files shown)


BrianWhitneyAI:
Screenshot 2025-02-27 at 12 42 23 PM

packages/core/state/interaction/logics.ts
```diff
@@ -625,21 +626,44 @@ const copyFilesLogic = createLogic({
                     status === "DOWNLOAD_IN_PROGRESS" ||
                     status === "DOWNLOAD_STARTED"
                 ) {
-                    successfulFiles.push(fileId);
+                    const file = fileDetails.find((f) => f.id === fileId);
+                    const isAlreadyCached = file?.annotations.some(
+                        ({ name, values }) =>
+                            name === "Should Be in Local Cache" && values[0] === true
```
aswallace, line 632: we might want to use the `AnnotationName.SHOULD_BE_IN_LOCAL` constant here and elsewhere (saw that the full string is also used in other places) if possible

kmitcham (approved):

SeanLeRoy (approved):

PR #458, opened 2025-02-28
bugfix/empty-local-path

BrianWhitneyAI:


packages/core/entity/Annotation/AnnotationName.ts
```diff
@@ -6,7 +6,7 @@ export default {
     FILE_NAME: "file_name", // a file attribute (top-level prop on file documents in MongoDb)
     FILE_SIZE: "file_size", // a file attribute (top-level prop on file documents in MongoDb)
     FILE_PATH: "file_path", // a file attribute (top-level prop on file documents in MongoDb)
-    LOCAL_FILE_PATH: "File Path (Local VAST)", // (optional) annotation for FMS files on the local NAS
+    LOCAL_FILE_PATH: "file_local_path", // (optional) annotation for FMS files on the local NAS
```
SeanLeRoy, line 9: I don't understand the need for this? When does "File Path (Local VAST)" come into play?

PR #464, opened 2025-03-07
Feature/metadata editing/restrict annotations

### LINKED ISSUE #341: Add password logic to edit modal

### LINKED ISSUE #343: Restrict editable annotations
Make sure we're limiting which annotations users can add/edit.

### New annotations
- [x] Add fuzzy search when users initially input new field name so that we can verify it won't collide with an existing annotation

### Existing annotations
- [x] Add hardcoded list of annotations that we won't allow users to edit (e.g., fields added by labkey, cacheing)
- [x] Make sure list renders as disabled in edit modal when searching available annotations

Part of #96

SeanLeRoy:
## Changes
Anya added a ton of metadata editing logic this is the last piece which just adds a password modal auth before the actual edit modal and also restricts some of the annotations by hard coding some locked down ones. Resolves #341 & resolves #343.

Additionally:
* I randomly also made a common spinner component as part of this because it was bugging me having to implement it so many times.
* Pushed some stuff into common use hooks 
* Tried to simplify some of the ways props were passed down to reduce render counts and enhance readability
* Fixed the way getFiles worked (this is already on `main` but because this is merged into a different branch it shows as a diff)

## Testing
Tested locally against staging and production (staging seems to recently have gotten a data change though making it difficult to test with)

## TODO

- [x] Talk with users about what other programs we need to add and then add them


SeanLeRoy:
Note: We need to add some more programs to the list of supported ones before releasing -- talking with users about that right now

packages/core/components/Modal/EditMetadata/index.tsx
```diff
@@ -18,16 +18,23 @@ import styles from "./EditMetadata.module.css";
 // a more robust solution in the future such as login or a more secure method.
 // or find a way to get this information passed in via GH secrets to the packaged
 // web bundle
+// TODO: Actually lets try to request this from GH or something..?
```
SeanLeRoy, line 21: TODO: Before merging - test putting this into a file or something at least to get it out of this level of hard-coded

packages/core/entity/Annotation/index.ts
```diff
@@ -81,6 +82,10 @@ export default class Annotation {
         return this.annotation.isOpenFileLink || false;
     }
 
+    public get isImmutable(): boolean {
```
aswallace, line 85: wondering if would benefit from an explanatory comment here (e.g., that this is mainly to prevent editing on certain top level annotations)

packages/core/entity/AnnotationFormatter/index.ts
```diff
@@ -22,7 +23,7 @@ export const AnnotationTypeIdMap = {
     [AnnotationType.BOOLEAN]: 3,
     [AnnotationType.DATETIME]: 4,
     [AnnotationType.DROPDOWN]: 5,
-    // [AnnotationType.LOOKUP]: 6, // Not currently supported
+    [AnnotationType.LOOKUP]: 6, // Not currently supported
```
kmitcham, line 26: is this comment still valid?

packages/core/hooks/useFileAccessContextMenu.ts
```diff
@@ -136,7 +136,7 @@ export default (filters?: FileFilter[], onDismiss?: () => void) => {
                         ],
                     },
                 },
-                ...(isQueryingAicsFms
+                ...(isQueryingAicsFms && isOnWeb
```
aswallace, line 139: are we preventing editing on desktop?

packages/core/components/StatusMessage/StatusMessage.module.css
```diff
@@ -4,7 +4,7 @@
     width: 40vw; 
     height: 100%;
     left: 30px;
-    z-index: 1000;
+    z-index: 99999999999;
```
aswallace, line 7: originally didn't do this because then it overlaps with the modals, but realize now it's not really an issue since they can just be closed 
image

packages/core/services/AnnotationService/HttpAnnotationService/immutableAnnotationNames.ts
```diff
@@ -0,0 +1,72 @@
...
+    ...TOP_LEVEL_FILE_ANNOTATION_NAMES,
+    "Cache Eviction Date",
+    "Cas9",
+    "Cas9 Batch",
+    "Cell Batch",
+    "Cell Line",
+    "Cell Line R And D",
+    "Cell Line Source",
+    "Cell Population Id",
+    ,
```
aswallace, line 13: extra comma?

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -102,6 +102,9 @@ export default class HttpFileService extends HttpServiceBase implements FileServ
     public async getFiles(request: GetFilesRequest): Promise<FileDetail[]> {
         const { from, limit, fileSet } = request;
 
+        if (limit - from > 1000) {
+            throw new Error("uhh" + limit + " " + from);
```
aswallace, line 106: still needed?

packages/core/state/interaction/logics.ts
```diff
@@ -516,24 +516,26 @@ const editFilesLogic = createLogic({
...
                 filters,
                 fileService,
                 sort: sortColumn,
             });
             const totalFileCount = await fileSet.fetchTotalCount();
+            console.log("FILTERS", filters, totalFileCount);
             filesSelected = await fileSet.fetchFileRange(0, totalFileCount);
         } else {
             filesSelected = await fileSelection.fetchAllDetails();
+            console.log("non-filters", filesSelected.length);
```
aswallace, line 538: still needed?

aswallace (commented):
I'm only partway through reviewing, just adding comments so far for now

packages/core/hooks/useFileAccessContextMenu.ts
SeanLeRoy: Yea I was originally but I think I'll add it back and use LK or something to grab the passwords - I just didn't want the desktop to be a relic holding stale passwords

packages/core/services/FileService/HttpFileService/index.ts
SeanLeRoy: lol

packages/core/components/Icons/LoadingIcon.tsx
```diff
@@ -0,0 +1,25 @@
...
+import * as React from "react";
+
+import styles from "./LoadingIcon.module.css";
+
+interface Props {
+    className?: string;
+    invertColor?: boolean;
+    size?: SpinnerSize;
+
+    "data-testid"?: string;
```
BrianWhitneyAI, line 12: what is this for?

packages/core/components/Modal/EditMetadata/PasswordForm.tsx
```diff
@@ -0,0 +1,88 @@
...
+            </>
+        );
+    }
+
+    let infoMessage;
+    if (props.isInvalidSelection) {
+        infoMessage = (
+            <p className={styles.errorMessage}>
+                The files you have selected do not share the same program or do not have a program
+                at all.
```
BrianWhitneyAI, line 42: "program at all" feels a little weird. maybe "program metadata tag"

BrianWhitneyAI (approved):

packages/core/components/Icons/LoadingIcon.tsx
SeanLeRoy: This is a prop used by the testing components - generally we don't need to explicitly define it like this but in this case we do

packages/core/components/EditMetadata/ExistingAnnotationPathway.tsx
```diff
@@ -65,16 +81,58 @@ export default function ExistingAnnotationPathway(props: ExistingAnnotationProps
...
         }
-        setSelectedAnnotation(selectedFieldName);
-        setAnnotationType(option?.data);
-        setValueCount(valueMap);
+
+        // String type annotations might be dropdown or lookup
+        // If it's a dropdown, we need to fetch the dropdown options
+        if (option?.data === AnnotationType.STRING) {
+            annotationService
+                .fetchAnnotationDetails(selectedFieldName)
```
aswallace, line 89: Is this likely to ever take long (i.e., any chance we'd need a loading indicator on the UI side?)

packages/core/components/Modal/EditMetadata/PasswordForm.module.css
```diff
@@ -0,0 +1,79 @@
+.error-message {
+    background-color: rgba(194, 48, 48, 0.5);
```
aswallace, line 2: there are also the global variables `--error-status-background-color` and `--error-status-text-color` which are what we use for backgrounds/borders in the status messages

packages/core/components/Modal/EditMetadata/PasswordForm.tsx
```diff
@@ -0,0 +1,88 @@
...
+ */
+export default function PasswordForm(props: Props) {
+    const [password, setPassword] = React.useState("");
+
+    if (!props.validPrograms) {
+        return (
+            <>
+                <p>
+                    A password is required for editing or creating new metadata. Each password is
+                    specific to a program, only files associated with that program can be edited.
```
aswallace, line 29: tiny nit: program; instead of program,
also in line 65

aswallace (approved):
Looks good!! I couldn't fully test locally but tried out the pre-submit behavior and things work as expected

packages/core/components/EditMetadata/ExistingAnnotationPathway.tsx
SeanLeRoy: I didn't see any performance slowness in my testing so I'm going to roll with this for now and if we ever notice it start to slow we can add some loading stuff

packages/core/components/EditMetadata/ExistingAnnotationPathway.tsx
```diff
@@ -25,27 +27,41 @@ interface ExistingAnnotationProps {
...
+    const annotationValueByNameMap = useAnnotationValueByNameMap();
+    const filters = useSelector(interaction.selectors.getFileFiltersForVisibleModal);
+    const annotationService = useSelector(interaction.selectors.getAnnotationService);
+    const annotationOptions: IComboBoxOption[] = Annotation.sort(
+        useSelector(metadata.selectors.getAnnotations)
+    ).map((annotation) => ({
+        key: annotation.name,
+        text: annotation.displayName,
+        disabled: annotation.isImmutable,
+        title: annotation.isImmutable
```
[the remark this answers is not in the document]
SeanLeRoy: I'm able to see it :(

packages/core/entity/AnnotationFormatter/index.ts
SeanLeRoy: Nope!

packages/core/hooks/useFileAccessContextMenu.ts
SeanLeRoy: Now that the stuff is in the bucket this now allows editing on desktop too

packages/core/components/Modal/EditMetadata/index.tsx
SeanLeRoy: I moved to the bucket (ty Anya)

PR #466, opened 2025-03-07
Add simple query argument input

### LINKED ISSUE #436: Add simple query argument for human readable URL input
See context in #432, but it seems like a simple addition to make it easy for users to construct a simple url linking to a dataset in BFF like so https://bff.allencell.org?url=https://my-data-server/my_file.csv. Of course this would need to assume things based on the URL provided or default to certain parameters, but it could be a nice simple addition.

SeanDuHare:
## Context
The query arguments for BFF are fairly complex which is necessary for the amount of customization we allow via URL sharing however it would be nice to be able to supply just a simplified version of it for when you just need to share a BFF link to a dataset which is what this does. Also included in this is the renaming of the entity FileExplorerURL -> SearchParams which felt more accurate. I tagged the two files where there are actual code changes.

Resolves #436 

## Testing
Tested running locally like so:
http://localhost:9001/app?url=https://biofile-finder-datasets.s3.us-west-2.amazonaws.com/Variance%20Paper%20Dataset.csv


packages/core/entity/SearchParams/index.ts
SeanLeRoy, line 1: This file has actual code changes

packages/core/entity/SearchParams/test/searchparams.test.ts
SeanLeRoy, line 1: This file has actual code changes (an added test)

packages/core/entity/SearchParams/index.ts
```diff
@@ -177,12 +177,29 @@ export default class FileExplorerURL {
...
+    }
+
+    private static decodeSimpleParams(params: URLSearchParams): SearchParamsComponents {
+        const unparsedURLs = params.getAll("url");
+        return {
+            ...EMPTY_QUERY_COMPONENTS,
+            sources: unparsedURLs.map((uri) => ({
+                uri,
+                ...getNameAndTypeFromSourceUrl(uri),
+            })),
```
aswallace, line 198: If there are other params present will they be ignored? e.g., http://localhost:9001/app?url=https://biofile-finder-datasets.s3.us-west-2.amazonaws.com/Variance%20Paper%20Dataset.csv&group=Cell+Line

aswallace (approved):
Do we plan to simplify other params as well or just this specific case for now?

packages/core/entity/SearchParams/index.ts
SeanLeRoy: Yea - for more complex stuffs they'll need to use the other system (or maybe we can add on later TBD)

SeanLeRoy:
> Do we plan to simplify other params as well or just this specific case for now?

This is just for simple use case like maybe another app wants to send their dataset to us

BrianWhitneyAI (approved):

PR #467, opened 2025-03-11
Add 'Browser' as an open with option + fix some path bugs

SeanLeRoy:
## Context
AIBS requested we add the Browser as an option for opening files with since they discovered this will happen (sometimes) to .png files that you click "Download" for (that aspect we can't really control)

## Changes
* Adds "Browser" as an option with option
* Cleans up path properties and getter to be simpler - also there was some buggy stuff happening to our internal paths due to being linked to the NGINX server unnecessarily + to external thumbnails since we were checking for the name not the path

## Testing
Tested using our data & some makeshift datasets pointing to random txt, png, and gifs on the web

### DIFF
--- a/packages/core/entity/FileDetail/index.ts
+++ b/packages/core/entity/FileDetail/index.ts
@@ -156,35 +156,10 @@ export default class FileDetail {
         return path as string;
     }
 
-    public get cloudPath(): string {
-        //// REMOVE THIS (BACKWARDS COMPAT)
-        if (this.path.startsWith("/allen")) {
-            return this.convertAicsDrivePathToAicsS3Path(this.path);
-        }
-        ////
-
-        // AICS FMS files' paths are cloud paths
-        return this.path;
-    }
-
     public get localPath(): string | null {
-        //// REMOVE THIS (BACKWARDS COMPAT)
-        if (this.path.startsWith("/allen")) {
-            return this.path;
-        }
-        ////
-
-        if (this.getAnnotation("Cache Eviction Date")) {
-            return this.getLocalPath();
-        }
-        return null;
-    }
-
-    public getLocalPath(): string | null {
-        const localPrefix = NAS_HOST_PREFIXES[this.env];
         const cloudPrefix = `${AICS_FMS_S3_URL_PREFIX}${AICS_FMS_S3_BUCKETS[this.env]}`;
-
-        if (this.path.startsWith(cloudPrefix)) {
+        if (this.getAnnotation("Cache Eviction Date") && this.path.startsWith(cloudPrefix)) {
+            const localPrefix = NAS_HOST_PREFIXES[this.env];
             const relativePath = this.path.replace(`${cloudPrefix}`, "");
             return `${localPrefix}${relativePath}`;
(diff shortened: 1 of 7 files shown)


packages/core/entity/FileDetail/index.ts
```diff
@@ -156,62 +156,20 @@ export default class FileDetail {
...
-    public get downloadPath(): string {
-        //// REMOVE THIS (BACKWARDS COMPAT)
-        if (this.path.startsWith("/allen")) {
-            return `http://aics.corp.alleninstitute.org/labkey/fmsfiles/image${this.path}`;
-        }
-        ////
-
-        const localPath = this.getLocalPath();
-        // For AICS files that are available on the Vast, users can use the cloud path, but the
-        // download will be faster and not incur egress fees if we download via the local network.
```
BrianWhitneyAI, line 204: No more labkey download?

kmitcham (approved):

packages/core/entity/FileDetail/index.ts
SeanLeRoy: Since all files have cloud paths we can just download from there now rather than use the NGINX server redirect to the Vast

SeanLeRoy: True, I am able to download that file on this branch

packages/core/entity/FileDetail/index.ts
```diff
@@ -156,62 +156,20 @@ export default class FileDetail {
...
-        }
-        return null;
-    }
-
-    public getLocalPath(): string | null {
-        const localPrefix = NAS_HOST_PREFIXES[this.env];
         const cloudPrefix = `${AICS_FMS_S3_URL_PREFIX}${AICS_FMS_S3_BUCKETS[this.env]}`;
-
-        if (this.path.startsWith(cloudPrefix)) {
+        if (this.getAnnotation("Cache Eviction Date") && this.path.startsWith(cloudPrefix)) {
```
pgarrison, line 161: I confirmed that all files in FSS with `should_be_in_local` also have a cache eviction date, so this should be safe to deploy 👍

pgarrison (approved):

aswallace (approved):

aswallace:
Minor confusion on my part about the context, is the desired behavior for this to open the png/etc so that it's visible in a tab or for the file to download? When I've been manually testing the file just downloads, so I'm still a bit unclear about the difference between this and using the regular 'DOWNLOAD' button

SeanLeRoy:
> Minor confusion on my part about the context, is the desired behavior for this to open the png/etc so that it's visible in a tab or for the file to download? When I've been manually testing the file just downloads, so I'm still a bit unclear about the difference between this and using the regular 'DOWNLOAD' button

This is unfortunately browser & path dependent behavior - in general if we don't specify a download attribute for a download in a browser it should default to opening it however that isn't always the case (seemingly specifically with S3 files...) so this does a best effort at trying to open when we say open and download when we say download. Note: the inverse of this is also true where sometimes it'll open a new tab rather than actually downloading :(

PR #468, opened 2025-03-11
Remove .ome from file name generated

SeanLeRoy:
I noticed .ome is present in the auto-generated file name for .ome.tiff and .ome.zarr files which is kinda incorrect so this adds an additional regex to remove it when/if present


aswallace (approved):

kmitcham (approved):
It would be great if there was a way to put unit tests on that, but I don't have an easy way to do it offhand

Issue #349, opened 2024-11-23
Add option to make modal not movable


aswallace:
Users should not see the "move" cursor for the edit modal. If we still want to allow other modals to be movable, should add optional param that changes the cursor & modal behavior
<img width="733" alt="image" src="https://github.com/user-attachments/assets/6ad23a10-965c-4be5-9217-a6223c0a6dab">

_Originally posted by @aswallace in https://github.com/AllenInstitute/biofile-finder/issues/338#issuecomment-2494703505_

PR #475, opened 2025-03-13
Feature/optional params on components

### LINKED ISSUE #348: Redundant tooltips on button component
Tooltips are not necessary on buttons when they are identical to the button label. Either make `title` an optional attribute, 
or pass empty string in all buttons that use the same tooltip & label

image

_Originally posted by @aswallace in https://github.com/AllenInstitute/biofile-finder/issues/338#issuecomment-2494703505_

aswallace:
## Context
Minor UX housekeeping: These component properties should be made optional to conform to UX recommendations:
1. Tooltips are only useful if they add info--should not be present if they would exactly match the button text
2. Most of our modals should not be movable/should not show the drag cursor

closes #348 and #349 

## Changes
- Made `title` (the tooltip) an optional property for `Base`, `Primary` and `Secondary` buttons. Still required on `Tertiary` since those are icon-only 
- Made modals non-draggable by default by adding an optional `isDraggable` property

## Testing
- Updated unit tests since the button `dataset-id` was previously constructed from `title`

## To do
- [x] Check with UX on which (if any) modals should still be draggable
    - edit: confirmed none need to be draggable atm, but can leave param in case we want it in the future
- I made some of the tooltips more descriptive instead of removing them, but am open to suggestions

### DIFF
--- a/packages/core/components/Modal/CodeSnippet/index.tsx
+++ b/packages/core/components/Modal/CodeSnippet/index.tsx
@@ -108,5 +108,5 @@ export default function CodeSnippet({ onDismiss }: ModalProps) {
         </>
     );
 
-    return <BaseModal body={body} onDismiss={onDismiss} title="Code snippet" isDraggable />;
+    return <BaseModal body={body} onDismiss={onDismiss} title="Code snippet" />;
 }
--- a/packages/core/components/Modal/CodeSnippet/index.tsx
+++ b/packages/core/components/Modal/CodeSnippet/index.tsx
@@ -108,5 +108,5 @@ export default function CodeSnippet({ onDismiss }: ModalProps) {
         </>
     );
 
-    return <BaseModal body={body} onDismiss={onDismiss} title="Code snippet" />;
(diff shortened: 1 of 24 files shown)


packages/core/components/Modal/CodeSnippet/index.tsx
```diff
@@ -108,5 +108,5 @@ export default function CodeSnippet({ onDismiss }: ModalProps) {
         </>
     );
 
-    return <BaseModal body={body} onDismiss={onDismiss} title="Code snippet" />;
+    return <BaseModal body={body} onDismiss={onDismiss} title="Code snippet" isDraggable />;
```
SeanLeRoy, line 111: Kinda unclear why we ever want these to be draggable - should we just remove support for it?

SeanLeRoy (approved):

packages/core/components/Modal/CodeSnippet/index.tsx
aswallace: Removed in [2adc63c](https://github.com/AllenInstitute/biofile-finder/pull/475/commits/2adc63c65294879ecc03d07837f242a06ff3ea0e)

Issue #419, opened 2025-02-07
File Path (Local VAST) column is empty


pgarrison:
Steps to reproduce:
* Find some FMS files that are on the VAST
* Right click the file list header
* Click "Modify columns"
* Find "File Path (Local VAST)" in the list of columns and click it
* Click out of the pop-up to close it
* Scroll the file list right to show the "File Path (Local VAST)" column
* Actual behavior: All files show "--" for the local path column (see screenshot)
* Expected behavior: Local paths are displayed for FMS files on VAST

<img width="1384" alt="Image" src="https://github.com/user-attachments/assets/7d7a725c-16ad-4a85-9b3d-524e3e362a77" />

PR #476, opened 2025-03-13
feature/local-path-fes-update

### LINKED ISSUE #477: Empty Local Path column for csv export.
### Description 

When exporting a CSV with the "File Path (Local Vast)" Option selected the column is empty. This is because the csv is generated from a direct call to FES which does not know about "File Path (Local Vast)".

BrianWhitneyAI:
## Description 

This PR is a companion PR to https://github.com/AllenCellSoftware/file-explorer-service/pull/142. We have decided to create local file path in FES rather than in BFF. As a result this PR fixes #477 and #419 

### Testing 
I test that this feature is working correctly by locally hosting FES and exporting a CSV with files that on/off the Vast. I looked at the file details panel as well. Additionally "File Path (Local Vast)" appears at the bottom of query annotation selections (Filter, Sort, Group By) and is disabled.


BrianWhitneyAI:
I am having some trouble with the test for FileAnnotationList and the host mount adjustment.

packages/core/hooks/useOpenWithMenuItems/index.tsx
```diff
@@ -8,6 +8,7 @@ import FileFilter from "../../entity/FileFilter";
 import { interaction, metadata } from "../../state";
 
 import styles from "./useOpenWithMenuItems.module.css";
+import AnnotationName from "../../entity/Annotation/AnnotationName";
```
aswallace, line 11: tiny nit: up with other entity imports

aswallace (approved):

kmitcham (approved):

packages/core/components/FileDetails/FileAnnotationList.tsx
```diff
@@ -34,11 +34,15 @@ export default function FileAnnotationList(props: FileAnnotationListProps) {
...
         async function formatPathForHost() {
             if (!fileDetails || !active) return;
-
-            const localPath = fileDetails.localPath;
-            const path = localPath ? await executionEnvService.formatPathForHost(localPath) : null;
-
-            setLocalPath(path);
+            const localPath = fileDetails.getFirstAnnotationValue(AnnotationName.LOCAL_FILE_PATH);
+            if (!localPath) {
+                setLocalPath(null);
```
SeanLeRoy, line 39: Why set this to `null` here? It looks like if the boolean check on 43 fails we default to `null` anyway

SeanLeRoy (approved):

packages/core/hooks/useOpenWithMenuItems/index.tsx
BrianWhitneyAI: resolved in [55ab0e5](https://github.com/AllenInstitute/biofile-finder/pull/476/commits/55ab0e5082133b9c435eaef5fb5a52e2f1af5720)

packages/core/components/FileDetails/FileAnnotationList.tsx
BrianWhitneyAI: can't do exactly that because of the typing. Did something similar though in [55ab0e5](https://github.com/AllenInstitute/biofile-finder/pull/476/commits/55ab0e5082133b9c435eaef5fb5a52e2f1af5720)

SeanLeRoy (approved):

Issue #496, opened 2025-04-08
Unable to download local path non-zarrs on desktop


aswallace:
## Description

Downloads that have a local path and are NOT zarr files fail to download on desktop. 
[Reported by Jacqueline](https://allencellscience.slack.com/archives/C021CEY96LD/p1744065477782299?thread_ts=1744060051.380649&cid=C021CEY96LD): 
> It won't allow me to download when I hit the button. I received an error "File download failed for file AD00006088_20240513_X02_001_s14_z17to61_crop_inset2x_shifted_scalebar5uM_RGB_z14.tif. Details:
Failed to construct 'URL': Invalid URL" 

Succeeds on web. Likely the same as #487 

## Expected Behavior

Should download successfully

## Reproduction

Attempt to download any file that has a local path and is NOT a zarr file from desktop.

## Environment
_Which version of BioFile Finder has the bug and with what environment?_
- [ ] Web
   - Browser: [e.g., Chrome, Safari]
- [x] Desktop
   - OS: [e.g. Windows, Linux, Mac with Apple chip, Mac with Intel chip]
   - Version: [e.g. 8.1.1]

Issue #487, opened 2025-03-21
Failed to download lif file


BrianWhitneyAI:
### Description 
this file https://s3.us-west-2.amazonaws.com/production.files.allencell.org/753/2c5/c93/de0/3fb/83a/040/474/6f0/276/cc/AICS-102_CofA_FinalBank.lif fails to download in BFF but is successful via URl. 

### Traceback
```
Unable to access AICS network Error: Request failed with status code 500
```
and message 
```
File download failed for file AICS-102_CofA_FinalBank.lif. Details:Failed to construct 'URL': Invalid URL
```

SeanLeRoy:
Works on web https://bff.allencell.org/app?c=file_name%3A0.4%2CKind%3A0.2%2CType%3A0.25%2CScene+Index%3A0.25&filter=%7B%22name%22%3A%22Cell+Line%22%2C%22value%22%3A%22AICS-102%22%2C%22type%22%3A%22default%22%7D&source=%7B%22name%22%3A%22AICS+FMS%22%7D&sort=%7B%22annotationName%22%3A%22uploaded%22%2C%22order%22%3A%22DESC%22%7D

PR #497, opened 2025-04-08
enable localpath downloads for non-zarr on desktop

aswallace:
## Context
Desktop was only checking `fileInfo.data`, but we now also use `fileInfo.path` to download files that have a local path. Because of this, the conditional was reading the path as undefined and skipping to the `else` condition. Then it would attempt to use the http file download method and fail to parse the file path as a url

## Changes
Adds `path` and checks for a `destination` (which indicates whether we're downloading a buffer and should use the http file method).

## Testing
Manually testing on desktop:
- Downloaded the files where the error was reported in #487 and #496 
- Tested downloading other files that have a local path and are not zarrs
- Made sure can still download files with only a cloud path


BrianWhitneyAI (approved):
Thank you!

PR #498, opened 2025-04-09
(Large) Feature/Add "no value" group behavior

aswallace:
## Context
When grouping on a metadata field, we currently filter out any files that do not have a value for that field. This can make folders/groups disappear (especially when nesting multiple groups) which can be unexpected behavior, and also limits which portions of their data users are able to organize and view at once.

## Changes
- Adds a checkbox to the "Group by" menu that allows users to toggle whether to display files/groupings that have no value for selected fields 
    - This is NOT default behavior. We will still default to filtering out null valued files. Old BFF links will not break
- Created a new util function for finding "child node" values in the directory tree that accounts for whether the "show null value" property is active
    - Updated "expand all" logic to use util as well

#### Manual testing: Tested with and without the no value option on both FMS and uploaded CSVs
  - Multiple nested groups where some have null values and others don't
  - Expand/collapse all
  - Saving metadata as csv/json/parquet
  - Downloading individual files

#### Unit testing:
- Added unit tests for the new child node util function
I haven't had much success writing useful tests for most of the other changes, but I'll keep looking into it

## Screenshots
Screenshot 2025-04-08 at 7 00 53 PM
Screenshot 2025-04-08 at 7 02 35 PM


packages/core/components/Checkbox/index.tsx
```diff
@@ -25,6 +26,8 @@ export default function BaseCheckbox(props: Props) {
     }
     return (
         <Checkbox
+            // defaults to undefined (handles its own state)
```
SeanLeRoy, line 29: What here defaults to undefined? Isn't `isChecked` a boolean?

packages/core/components/DirectoryTree/DirectoryTreeNode.tsx
```diff
@@ -49,6 +50,19 @@ export default function DirectoryTreeNode(props: DirectoryTreeNodeProps) {
         return fileSelection.isFocused(fileSet);
     }, [fileSelection, fileSet, collapsed]);
 
+    // Check file count to avoid displaying empty folders
```
SeanLeRoy, line 53: Hmm why did you decide to go with the route of having each component check asynchronously if it has values as opposed to as the time of gathering the folders to show? This seems like it could cause a lot of network calls.

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -0,0 +1,113 @@
...
+    const depth = pathToNode.length;
+    const annotationNameAtDepth = hierarchy[depth];
+    const userSelectedFiltersForCurrentAnnotation = combinedFileFilters
+        .filter(
+            (filter) => filter.name === annotationNameAtDepth && filter.type !== FilterType.EXCLUDE
+        )
+        .map((filter) => filter.value);
+
+    if (isRoot) {
+        const rootHierarchy = shouldShowNullGroups ? [annotationNameAtDepth] : hierarchy;
```
SeanLeRoy, line 58: Is `hierarchy` an empty list here? Could you add a comment explaining a bit of the difference between when we want `null` nodes and don't here?

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    const depth = pathToNode.length;
+    const annotationNameAtDepth = hierarchy[depth];
+    const userSelectedFiltersForCurrentAnnotation = combinedFileFilters
+        .filter(
+            (filter) => filter.name === annotationNameAtDepth && filter.type !== FilterType.EXCLUDE
+        )
+        .map((filter) => filter.value);
+
+    if (isRoot) {
```
SeanLeRoy, line 57: I think this could be written like so to simplify the ternary operator check:
```
if (shouldShowNullGroups) {
        // Fetch all values under node, ignoring past hierarchy
        values = await annotationService.fetchRootHierarchyValues(
            [annotationNameAtDepth],
            combinedFileFilters
        );
} else if (isRoot) {
         values = await annotationService.fetchRootHierarchyValues(
            hierarchy,
            combinedFileFilters
        );
} else {
        values = await annotationService.fetchHierarchyValuesUnderPath(
            hierarchy,
            pathToNode,
            combinedFileFilters
        );
}
```

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -0,0 +1,113 @@
...
+        );
+    } else {
+        values = await annotationService.fetchHierarchyValuesUnderPath(
+            hierarchy,
+            pathToNode,
+            combinedFileFilters
+        );
+    }
+
+    // Each specialized filter may be in one or both of `selectedFilters` and `fileSet.[specialized filter type]`.
```
SeanLeRoy, line 77: Are the `fileSet.<specialized filter type>` filters always a subset of the filters  in the `selectedFilters`? I didn't check, but I thought that is where they get pulled from which would simplify the way this is grabbing them.

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -0,0 +1,113 @@
...
+        .concat(fileSet?.includeFilters || []);
+    const excludeFilters = combinedFileFilters
+        .filter((f) => f.type === FilterType.EXCLUDE && !fileSet?.excludeFilters?.includes(f))
+        .concat(fileSet?.excludeFilters || []);
+    const fuzzyFilters = combinedFileFilters
+        .filter((f) => f.type === FilterType.FUZZY && !fileSet?.fuzzyFilters?.includes(f))
+        .concat(fileSet?.fuzzyFilters || []);
+    const filteredValues = values
+        .filter((value) => {
+            if (includeFilters?.some((filter) => filter.name === annotationNameAtDepth))
```
SeanLeRoy, line 90: This boolean check isn't dependent on the iteratee (`value`) which makes me think it could be written a bit more performant. I think what this is trying to do is include all values if the `annotationNameAtDepth` is an "include" filter and exclude all values if the `annotationNameAtDepth` is an "exclude" filter so maybe:
```
let filteredValues: <whatever the type is>[];
if (includeFilters?.some((filter) => filter.name === annotationNameAtDepth)) {
     // User wants this annotation
     filteredValues = values
} else if (excludeFilters?.some((filter) => filter.name === annotationNameAtDepth))) {
    // User does not want this annotation
    filteredValues = [];
} else if (!isEmpty(userSelectedFiltersForCurrentAnnotation)) {
    if (fuzzyFilters?.some((fuzzy) => fuzzy.name === annotationNameAtDepth)) {
          filteredValues = values.filter(value => value.includes(userSelectedFiltersForCurrentAnnotation[0]));
    } else {
          filteredValues = values.filter(value => userSelectedFiltersForCurrentAnnotation.includes(value));
    }
} else {
     // No filters exclude this annotation
     <I don't know why we would get to this point? Is this no filters...?>
     filteredValues = values;
}
```

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -0,0 +1,113 @@
...
+        );
+    }
+
+    // Each specialized filter may be in one or both of `selectedFilters` and `fileSet.[specialized filter type]`.
+    // Avoid double-counting
+    const includeFilters = combinedFileFilters
+        .filter((f) => f.type === FilterType.ANY && !fileSet?.includeFilters?.includes(f))
+        .concat(fileSet?.includeFilters || []);
+    const excludeFilters = combinedFileFilters
+        .filter((f) => f.type === FilterType.EXCLUDE && !fileSet?.excludeFilters?.includes(f))
```
SeanLeRoy, line 83: I think there are never exclude filters in your `combinedFileFilters` is that intended?

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -0,0 +1,113 @@
...
+        .concat(fileSet?.excludeFilters || []);
+    const fuzzyFilters = combinedFileFilters
+        .filter((f) => f.type === FilterType.FUZZY && !fileSet?.fuzzyFilters?.includes(f))
+        .concat(fileSet?.fuzzyFilters || []);
+    const filteredValues = values
+        .filter((value) => {
+            if (includeFilters?.some((filter) => filter.name === annotationNameAtDepth))
+                return true;
+            if (excludeFilters?.some((filter) => filter.name === annotationNameAtDepth))
```
SeanLeRoy, line 92: Since we don't rely on the annotation values themselves to determine whether the values matter here can we avoid querying for them at all?

packages/core/components/DirectoryTree/useDirectoryHierarchy.tsx
```diff
@@ -209,6 +187,9 @@ const useDirectoryHierarchy = (
                             take(pathToChildNode, depth + 1)
                         ).map((pair) => {
                             const [name, value] = pair as [string, string];
+                            if (value === NO_VALUE_NODE) {
+                                return new ExcludeFilter(name);
```
SeanLeRoy, line 191: Could `FileFilter` auto construct an `ExcludeFilter` when the `value` is `NO_VALUE_NODE `? I think it makes things cleaner when this kind of logic lives closer to the domain it is helping.

packages/core/components/DirectoryTree/useDirectoryHierarchy.tsx
```diff
@@ -233,12 +214,17 @@ const useDirectoryHierarchy = (
                             sort: sortColumn,
                         });
 
+                        const displayValue =
+                            value === NO_VALUE_NODE
+                                ? `No value ("${hierarchy[depth]}")`
+                                : annotationAtDepth?.getDisplayValue(value) || value;
```
SeanLeRoy, line 220: Same sort of thing as my earlier comment where I wonder if this would make more sense existing within the `getDisplayValue` function rather than in here.

packages/core/components/ListPicker/index.tsx
```diff
@@ -155,6 +161,17 @@ export default function ListPicker(props: ListPickerProps) {
                             Clear all
                         </ActionButton>
                     </Tooltip>
+                    {shouldShowNullGroups !== undefined && ( // avoid colliding with falsy value
```
SeanLeRoy, line 164: Why would `shouldShowNullGroups` be undefined?

SeanLeRoy (commented):
Might want to talk offline about some of what `findChildNodes` is doing but overall I like what you did with this! Looks like you were able to keep most of the complexity contained and re-used 🚀

packages/core/components/Checkbox/index.tsx
aswallace: This was referring to the internal FluentUI component (not ours), which doesn't require the `checked` property so defaults to undefined when that isn't provided. I think my comment is more confusing than clarifying though so I'll take it out!

packages/core/components/DirectoryTree/findChildNodes.ts
aswallace: Will add! Hierarchy is not an empty list, it's the whole group structure. When we want `null` nodes, we should ignore the rest of the hierarchy and only look at the values for the current annotation. Using the hierarchy implicitly adds `include` filters for all the other fields, which forces us to drop any values at the current level where something down the line is missing a value

packages/core/components/ListPicker/index.tsx
aswallace: We use the ListPicker component in multiple places, and for some of them we don't want the checkbox to show up (e.g., Sort). I'm double-using the `shouldShowNullGroups` property here by making it optional -- When we want the checkbox to show up, it's set to either true or false (which also coincides with how the checkbox should be set). When we don't want it to show up, I leave off the prop and it stays undefined

packages/core/components/DirectoryTree/findChildNodes.ts
aswallace: Discussed offline, combined in [d47cb48](https://github.com/AllenInstitute/biofile-finder/pull/498/commits/d47cb4856d4b94b0323dcc50c1d6301f67af90b7)

packages/core/components/DirectoryTree/useDirectoryHierarchy.tsx
aswallace: Changed in [547aed0](https://github.com/AllenInstitute/biofile-finder/pull/498/commits/547aed022256a7672ae3c741610098864a8bd1ec)

packages/core/components/DirectoryTree/DirectoryTreeNode.tsx
aswallace: Moved to findChildNodes per offline discussion in [c39b4dc](https://github.com/AllenInstitute/biofile-finder/pull/498/commits/c39b4dc9fc1abc768e6687e4d5c24d6f5a1ed288)

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -0,0 +1,102 @@
...
+        ancestorNodes,
+        currentNode,
+        fileSet,
+        hierarchy,
+        annotationService,
+        fileService,
+        shouldShowNullGroups,
+    } = defaults({}, params, DEFAULTS);
+
+    const isRoot = currentNode === ROOT_NODE;
```
BrianWhitneyAI, line 36: This looked funny to me, maybe just separate `(currentNode === ROOT_NODE)`

BrianWhitneyAI (approved):

packages/core/components/DirectoryTree/findChildNodes.ts
aswallace: I tried adding the parentheses in [a43d4e1](https://github.com/AllenInstitute/biofile-finder/pull/498/commits/a43d4e1a4905481ad5cec33b4798d0d714516125) but the linter removed them when I pushed

pgarrison (approved):

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -0,0 +1,99 @@
...
+        return shouldShowNullGroups ? [NO_VALUE_NODE] : [];
+    }
+
+    const userSelectedFiltersForCurrentAnnotation = fileSet.filters
+        .filter((filter) => filter.name === annotationNameAtDepth)
+        .map((filter) => filter.value);
+
+    if (shouldShowNullGroups) {
+        // Fetch all values under current node, ignoring past hierarchy
+        // Including the full hierarchy would filter out files that miss any part of the hierarchy
```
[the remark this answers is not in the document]
aswallace: My comment is definitely not super clear, open to some re-wording here. 

Let's say that we've applied the grouping hierarchy A -> B -> C. We're currently at level B under A=red and want to find the values for B that have any files.
In a normal setting, we pass the full hierarchy [A, B, C] to the endpoint, which applies those as "include" filters (they're all required). We also have an A=red filter applied because that's the parent folder.
With null groups turned on, we only need the A=red filter plus hierarchy=[B]. We don't want C in the hierarchy because we want to also capture files that don't have a C, and we don't need to include A because we already have A=red.

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -0,0 +1,100 @@
...
+    // We trim that from the path as it is not meaningful in this context
+    const pathToNode = pull([...ancestorNodes, currentNode], ROOT_NODE);
+
+    let values: string[] = [];
+
+    const depth = pathToNode.length;
+    const annotationNameAtDepth = hierarchy[depth];
+    if (fileSet.excludeFilters?.some((filter) => filter.name === annotationNameAtDepth)) {
+        // User does not want files with this annotation; don't return any non-null values.
+        return shouldShowNullGroups ? [NO_VALUE_NODE] : [];
```
SeanLeRoy, line 47: We need to do a check for the count here of the `NO_VALUE_NODE` right to make sure it has values?

SeanLeRoy (approved):

packages/core/components/DirectoryTree/findChildNodes.ts
aswallace: ah true, I'll move the count up to the top

aswallace: fixed in [d01d0fb](https://github.com/AllenInstitute/biofile-finder/pull/498/commits/d01d0fb644e756e7f00aa092a4e9b2062af85151)

packages/core/components/DirectoryTree/findChildNodes.ts
```diff
@@ -85,16 +95,13 @@ export async function findChildNodes(params: FindChildNodesParams): Promise<stri
     }
 
     const filteredValuesSorted = filteredValues.sort(naturalComparator);
-    if (!shouldShowNullGroups || userSelectedFiltersForCurrentAnnotation.length) {
+    // Don't add NO_VALUE_NODE if there are user-applied filters for the annotation
+    if (
+        !shouldShowNullGroups ||
```
SeanLeRoy, line 100: This is a nit more than anything but I think this is more readable in reverse ex. `shouldShowNullGroups && noValueFileCount > 0 && !userSelectedFiltersForCurrentAnnotation.length` but nbd either way. lgtm

packages/core/components/DirectoryTree/findChildNodes.ts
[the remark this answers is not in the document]
pgarrison: I see, that makes sense! The phrase "past hierarchy" tripped me up because I imagined the code evaluating level A first, then B, so I thought the comment meant A was removed.

packages/core/components/DirectoryTree/useDirectoryHierarchy.tsx
```diff
@@ -214,16 +188,8 @@ const useDirectoryHierarchy = (
 
                         // Filters are a combination of any user-selected filters and the filters
                         // at a particular path in the hierarchy.
-                        //
-                        // Remove any user-applied filters for any annotation within the current path.
-                        // E.g., if under the path "AICS-12" -> "ZSD-1", and a user has applied the filters FileFilter("Channel Type", "Raw 488nm")
-                        // and FileFilter("Cell Line", "AICS-33"), we do not want to include the latter in the query for this FileList.
-                        const hierarchyAnnotationNames = new Set(hierarchy);
-                        const userAppliedFilters = selectedFileFilters.filter(
-                            (f) => !hierarchyAnnotationNames.has(f.name)
```
aswallace, line 223: This caused #507

PR #508, opened 2025-04-25
Support subpaths into zarr/n5 containers

krokicki:
## Context
This addresses issue #503 to support OME-Zarrs with subpaths. After the usual check for extension (everything after the last dot) we now check to see if the path includes ".n5" or ".zarr" anywhere (similar to how #493 implemented the same feature in another part of the code).  

## Changes
Add additional checks to see if the path `includes` ".zarr" or ".n5", which can occur in the middle of a path.

I removed the "n5" check from the `switch` statement since it's no longer necessary, but I left the extra "zarr" in the `switch` for now, because it's not clear to me if all zarr viewers support subpaths within a container. Neuroglancer and OME-NGFF validator definitely do. Vol-E and Agave don't seem to, in which case maybe they should be removed from return statement after the `includes` check, but I'm not the one to make that call.

## Testing
Loaded [my test data CSV](https://docs.google.com/spreadsheets/d/e/2PACX-1vTTpnSiPRxOuJgq3dQ7aLYGtMPd2KYTKK6lYqz0geHmnygqMLYDC1XkRTHjPXJ9u_oC3t0Nu0FBIS_r/pub?gid=200533885&single=true&output=csv) and clicked through several examples to ensure that Neuroglancer appears under "Support file type" in "Open File".

PR #509, opened 2025-04-25
Fixed data format prefix when opening n5 image in neuroglancer

krokicki:
## Context
I noticed that when trying to open N5's in Neuroglancer, it tries to load them as Zarrs and fails. This is because the data URL is hardcoded with a `zarr://` prefix.

## Changes
Changed code to return `zarr://` or `n5://` depending on the extension in the path.

## Testing
Loaded [my test data CSV](https://docs.google.com/spreadsheets/d/e/2PACX-1vTTpnSiPRxOuJgq3dQ7aLYGtMPd2KYTKK6lYqz0geHmnygqMLYDC1XkRTHjPXJ9u_oC3t0Nu0FBIS_r/pub?gid=200533885&single=true&output=csv) (requires #508 fix applied) and clicked through several examples to ensure that N5's open correctly in Neuroglancer.

### DIFF
--- a/packages/core/hooks/useOpenWithMenuItems/index.tsx
+++ b/packages/core/hooks/useOpenWithMenuItems/index.tsx
@@ -97,7 +97,7 @@ const APPS = (fileDetails?: FileDetail): Apps => ({
         key: AppKeys.NEUROGLANCER,
         text: "Neuroglancer",
         title: `Open files with Neuroglancer`,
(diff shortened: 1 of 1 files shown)


SeanLeRoy (approved):
Oh this is awesome! We have some ways to go for getting a good default for Neuroglancer links. If you have any examples of how you/your org auto-generate links for neuroglancer (if you do) that would be great to add as an issue that we could build off of :)

aswallace (approved):

BrianWhitneyAI (approved):

krokicki:
@SeanLeRoy For sure! Creating working Neuroglancer links for e.g. multichannel images requires reading the OME-Zarr metadata. I'm working on a Python library to do that, which I can share when it's ready. It should be straight-forward translate to Javascript.

PR #511, opened 2025-04-26
Fix over-inclusive filter behavior

### LINKED ISSUE #507: Groups incorrectly include all filtered values
## Description
When a user groups on a particular field (e.g., barcode) and then also applies filters, groups will contain ALL the files instead of just the ones that have the value of that folder.
We've actually seen and fixed a similar but potentially unrelated issue before with DuckDB/.csv files (#442), but it seems to have popped up again after the new "no-value" grouping behavior updates in #498.

## Expected Behavior

Each group should only include files that have the field value of the group they're in

## Reproduction

From Antoine: 
> Here I select all files from 2 barcodes, and I groupby by barcodes. It does create the 2 barcodes folder, but it put all the selected files in all the folders.  Here as you can see the folder for the barcode 6618 also contains files from the barcode 2227
Image

## Environment
- [x] Web
- [x] Desktop
   - Version: [e.g. 8.1.1]

aswallace:
## Context
In #498 we [took out logic](https://github.com/AllenInstitute/biofile-finder/pull/498/files#r2061171756) that removed all hierarchy filters from child nodes. This made it so that ALL filters were included in child node file sets; e.g., if we were grouping on Color and had color=red and color=blue filters applied, BOTH filters got passed to each "color" child node. 
resolves #507 

## Changes
We don't want to put back the original code as it was, since we do need to leave some hierarchy-related filters in (e.g., keep filters for fields further along in the hierarchy that we haven't reached yet, and keep include/exclude filters for any field in the hierarchy). 

Also caught a similar bug for ranges & include filters:
- [Example on staging for range bug](https://staging.bff.allencell.org/app?c=file_name%3A0.4%2CKind%3A0.2%2CType%3A0.25%2Cfile_size%3A0.15&group=Plate+Barcode&group=Position+Index&filter=%7B%22name%22%3A%22Plate+Barcode%22%2C%22value%22%3A%223500005550%22%2C%22type%22%3A%22default%22%7D&filter=%7B%22name%22%3A%22Plate+Barcode%22%2C%22value%22%3A%223500005551%22%2C%22type%22%3A%22default%22%7D&filter=%7B%22name%22%3A%22uploaded%22%2C%22value%22%3A%22RANGE%282024-04-25T07%3A00%3A00.000Z%2C2025-04-26T06%3A59%3A59.964Z%29%22%2C%22type%22%3A%22default%22%7D&filter=%7B%22name%22%3A%22Position+Index%22%2C%22value%22%3A%22RANGE%281%2C10%29%22%2C%22type%22%3A%22default%22%7D&openFolder=%5B%223500005551%22%5D&openFolder=%5B%223500005550%22%5D&source=%7B%22name%22%3A%22AICS+FMS%22%7D&sort=%7B%22annotationName%22%3A%22uploaded%22%2C%22order%22%3A%22DESC%22%7D)

## Testing
- Manually tested on the case mentioned in the issue, plus cases with more complex nesting & filters applied
- Unit tests for `findChildNodes` changes
- Have not yet added tests for `useDirectoryHierarchy`, may need to do more refactoring in the future to make it more easily testable

### DIFF
--- a/packages/core/components/DirectoryTree/findChildNodes.ts
+++ b/packages/core/components/DirectoryTree/findChildNodes.ts
@@ -2,6 +2,7 @@ import { defaults, isEmpty, pull } from "lodash";
 
 import { NO_VALUE_NODE, ROOT_NODE } from "./directory-hierarchy-state";
 import FileSet from "../../entity/FileSet";
+import { FilterType } from "../../entity/FileFilter";
 import ExcludeFilter from "../../entity/FileFilter/ExcludeFilter";
 import { AnnotationService, FileService } from "../../services";
 import { naturalComparator } from "../../util/strings";
@@ -58,7 +59,12 @@ export async function findChildNodes(params: FindChildNodesParams): Promise<stri
     }
 
     const userSelectedFiltersForCurrentAnnotation = fileSet.filters
-        .filter((filter) => filter.name === annotationNameAtDepth)
+        .filter(
+            (filter) =>
+                filter.name === annotationNameAtDepth &&
+                !filter.value.includes("RANGE") && // 'RANGE' filters are handled by the value fetching endpoint
+                filter.type !== FilterType.ANY // 'Include' filters have a blank value and shouldn't be counted here
+        )
         .map((filter) => filter.value);
 
     if (shouldShowNullGroups) {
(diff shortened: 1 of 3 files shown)


BrianWhitneyAI (approved):

kmitcham (approved):
I'm a little surprised to not see updates to the tests for this behavior.

aswallace:
> I'm a little surprised to not see updates to the tests for this behavior.

I added in the unit tests (which I'd written but apparently not pushed) for the changes to `findChildNodes`. Unfortunately, `useDirectoryHierarchy` isn't written in a very testable way, but I can look into adding something in

PR #514, opened 2025-04-29
Bugfix/Unable to close Electron child windows

### LINKED ISSUE #513: Unable to close LabkeyPlateUI after interactions
## Description

When Electron opens a sub-window, it doesn't handle `unload` events the same way a Chrome browser does. The pop up that usually warns users about exiting a tab in Chrome doesn't show up and thus prevents users from being able to close the window at all

Here's what happens in Chrome: 
Image
In Electron, it just doesn't close

## Expected Behavior

We need to either handle those kinds of pop-ups separately, or ignore them and force the window to close

## Reproduction

From [Chris F](https://allencellscience.slack.com/archives/C2ANQFKL3/p1745858225868619): 
>  I opened a plate UI for an existing file in AICS FMS via BFF. I tried to view different details in the wells of the plate and then when I tried to click to close the plate, the plate UI window would not close when I clicked the red x or when I entered command+w.

## Environment
- [ ] Web
- [x] Desktop
   - OS: Mac, but likely others
   - Version: 8.4

aswallace:
## Context
resolves #513 
When desktop users opened files in LabKey from BFF, if they interacted with LabKey at all they were unable to close the Electron window and had to quit/restart the entire app. 
This is because LabKey has a `beforeunload` event that warns users about unsaved changes when they try to exit, which is supposed to show up like this: 
image

However, Electron doesn't handle those events the same way, so was blocking the alert and any way of closing the window.

## Changes
Adds a custom handler for opening new windows in Electron ([docs here,](https://www.electronjs.org/docs/latest/api/window-open/) gets used any time we use `window.open` or `target=_blank` in desktop). This allows us to watch for `beforeunload` via `will-prevent-unload` events.
Now users will see this alert:
image

## Testing
Only manual testing:
- Tested by opening files in LabKey from BFF, interacting with interface and then closing (tested both leave and cancel)
- Made sure other apps (Agave, Neuroglancer, Vol-E) still work in desktop
- Checked the alert also works for other sites like GitHub (edit file and exit without saving)

### DIFF
--- a/packages/desktop/src/main/index.ts
+++ b/packages/desktop/src/main/index.ts
@@ -1,5 +1,5 @@
 import * as path from "path";
-import { app, BrowserWindow, Menu } from "electron";
+import { app, BrowserWindow, dialog, Menu } from "electron";
 
 import getMenuTemplate from "./menu";
 import ExecutionEnvServicElectron from "../services/ExecutionEnvServiceElectron";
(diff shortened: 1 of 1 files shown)


packages/desktop/src/main/index.ts
```diff
@@ -82,6 +82,29 @@ const createMainWindow = () => {
         mainWindow = undefined;
     });
 
+    // Customize window.open or target="_blank" calls that create new Electron windows
```
BrianWhitneyAI, line 85: maybe add a labkey specifc comment so that if something changes its easier for devs to pick up?

BrianWhitneyAI (approved):

packages/desktop/src/main/index.ts
aswallace: added something in [ed545f2](https://github.com/AllenInstitute/biofile-finder/pull/514/commits/ed545f22479263b3218dc9c8142af30d20da090e)

kmitcham (approved):
One change, clearly explained.  Is this the perfect PR?

PR #516, opened 2025-05-02
Open selections of files with metadata in Vol-E

frasercl:
## Context

Since allen-cell-animated/vole-app#362, Vol-E has supported opening a group of multiple images as a single multi-scene collection. Scientist users want the ability to select many files in BFF and open them as a multi-scene collection in Vol-E, and to have metadata for those files available in Vol-E's metadata panel.

In the past, we've used URL query parameters (`?url=...`) to express which image to open. But arbitrarily long lists of image URLs, or arbitrarily-sized metadata records, could easily create unusably large URLs. So we've opted to add a new approach which pulls image URLs and metadata from the browser's local storage. allen-cell-animated/vole-app#379 implemented the Vol-E side of this transaction, and that PR's description gets into the whole procedure in more detail. But briefly, here's what BFF needs to do when the user clicks to open a selection of files in Vol-E:

1. Load a particular Vol-E route (`/write_storage`) as an `iframe`
2. Once it's loaded, transfer URLs and metadata using `postMessage`
3. Wait to get a success message back, then open Vol-E

## Changes

Adds a new click handler for when Vol-E is selected in the "open with" context menu. Per UX's recommendation, no UI changes are made to allow users to pick between the old behavior (just open the focused file) and the new one &mdash; we assume that if the user has made a selection, they always expect to open the whole selection.

The click handler relies on a brand new custom hook: `useMessageExternalSite`, which, along with the `MessageExternalSiteProvider` component, lazily creates an `iframe` when the send function it provides is called.

## Testing

Note that currently, the changes in allen-cell-animated/vole-app#379 have _not_ been deployed to production (they will be very soon), so the changes here will not work on their own. Until they are, you can test by:

1. `npm start` your own local copy of vole-app
2. Change the `VOLE_BASE_URL` constant in useOpenWithMenuItems/index.tsx to `"http://localhost:9020"`

Then, `npm start` BFF, make a selection of multiple OME-Zarr images, and right-click and open with Vol-E.


BrianWhitneyAI (approved):

BrianWhitneyAI (approved):

packages/core/hooks/useMessageExternalSite.tsx
```diff
@@ -0,0 +1,110 @@
+import React from "react";
+
+type MessageExternalSiteContextType = {
+    send: (url: string, message: NonNullable<unknown>) => void;
+    setOnReceive: (onReceive: ((message: unknown) => void) | undefined) => void;
+};
+
+const noop = () => undefined;
```
aswallace, line 8: consistency nit: I believe elsewhere we're using the lodash `noop`, but am also fine with leaving this as is

aswallace (approved):
Tested locally, looks good! 
The merge conflicts may be because of recent changes to the 'open with' menu, specifically for detecting zarr files (#508) and the way electron opens new windows (#514). I don't think they'll affect the functionality of this PR, but worth keeping an eye on when you merge.

packages/core/hooks/useMessageExternalSite.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * invisible `iframe` when we want to post a message to an external site.
+ *
+ * This exists separately from the hook so that the `iframe` can be placed in a stable location in
+ * the document, allowing the hook to be used in components that unmount and re-mount often without
+ * reloading the `iframe` over and over.
+ */
+const MessageExternalSiteProvider: React.FC = ({ children }) => {
+    const [iFrameUrl, setIFrameUrl] = React.useState<string | null>(null);
+    const onReceiveRef = React.useRef<((message: unknown) => void) | undefined>(undefined);
+    const pendingMessageRef = React.useRef<NonNullable<unknown> | null>(null);
```
SeanLeRoy, line 25: I don't think I've seen `NonNullable` used like this, whats the benefit of providing it here?

packages/core/hooks/useOpenWithMenuItems/index.tsx
```diff
@@ -207,6 +217,48 @@ export default (fileDetails?: FileDetail, filters?: FileFilter[]): IContextualMe
         metadata.selectors.getAnnotationNameToAnnotationMap
     );
     const loadBalancerBaseUrl = useSelector(interaction.selectors.getLoadBalancerBaseUrl);
+    const [sendMessageToVole, setOnReceiveFromVole] = useMessageExternalSite(
+        `${VOLE_BASE_URL}/write_storage`
+    );
+
+    const openInVole = React.useCallback(async () => {
```
SeanLeRoy, line 224: How many files have you seen this succeed with? Do we have an idea of when this starts to slow down or break (if at all?)?

SeanLeRoy: I'm wondering if this is overkill for if the user just wants to pass a single file to Vol-E. But maybe if there isn't a slow-down it doesn't matter?

SeanDuHare (approved):

frasercl:
https://github.com/allen-cell-animated/vole-app/pull/379 is deployed, but changes aren't working as expected. This looks to be due to *storage partitioning*, a behavior where browsers give sites one local storage for standalone (non-`iframe`) contexts and a different one for every site they're embedded in. So the Vol-E `iframe` doesn't write to "*vole.allencell.org*" storage, but specifically to "*vole.allencell.org embedded in bff.allencell.org*" storage, which becomes inaccessible when Vol-E is opened standalone. This was apparently introduced somewhat recently, at least relative to the guides that recommended this approach to cross-site data sharing &mdash; see e.g. [chrome's explainer here](https://privacysandbox.google.com/cookies/storage-partitioning). Local testing worked when both sites are localhost, but the same behavior appears when one or the other is referred to as 127.0.0.1 instead.

There is a solution in the [Storage Access API](https://developer.mozilla.org/en-US/docs/Web/API/Storage_Access_API), which allows embedded sites to request access to their standalone local storage. Integrating these changes might take some work though, including likely meaningful changes to this PR. I'm going to take this back into draft for now to investigate this, and may re-request reviewers depending on how significantly I have to change this side.

PR #519, opened 2025-05-07
Serving sub dirs

will-moore:
## Context
This is based on Sean's branch at https://github.com/AllenInstitute/biofile-finder/tree/feature/serving-sub-dirs but sets the basename to be an empty string (so that this is hopefully not a breaking change)?

With this branch, I am able to build the bff app and serve it from within OMERO web, dynamically setting the `basename` to the URL where this is being served (depending on the OMERO.web deployment settings) by editing the `app...js` code like this in the Django views.py code:

```python
content = content.replace('{basename:""}', f'{{basename:"{basename}"}}')
```

I tried creating a `const BASENAME = ""` and `basename: BASENAME,` in the index.tsx in the hope it would give a more precise target for the `replace()` above, but the `const BASENAME` is compiled away in the bff build so we get the same `app...js` anyway.
It seems to be working fine for now, but you may have an idea how to make this a bit cleaner/safer etc? 

## Testing
I'm hoping that these changes have no effect on the BFF app itself.

I'll leave this as a draft PR until we've done more testing on the omero_biofilefinder app. I'll link the PR here soon...

NB: I don't seem to need the other change `path: "*",` for `<Home />` that Sean suggested. Just noting it here in case it's needed for something else?


SeanLeRoy:
Interesting!! I'd prefer something that didn't require you having to hack into the code at runtime for your own benefit, but if this works for you & doesn't break it for us then seems fine to me :) 

Did this allow users to refresh on sub-pages (ex. if you navigate to the `/datasets` page and click refresh does that work)? This is what the "*" was supposed to help with by creating a fallback.

SeanLeRoy:
@will-moore Is this working for you? Would you like to roll with this for now or do you want us to investigate a different solution?

will-moore:
Sorry - Yes, this is working for me for now.
When I serve BFF from omero-web, it doesn't yet support refresh of other URLs such as `/datasets` because that route isn't handled by the omero_biofilefinder Django server, so you don't even get the app served. That will require more work at some point, including testing of the "*" effect.
But if the current change here is good from your end then it would be great to get it in. Thanks

SeanLeRoy (approved):

PR #522, opened 2025-05-12
Bugfix/Query UI stuck in loading state

### LINKED ISSUE #517: Show errors for parameter parsing issues
## Description

When the URL is misformed, like [this URL](https://bff.allencell.org/app?c=File+Name%3A0.5%2CBatch%3ANaN%2CRound%3ANaN%2CDate%3ANaN&source=%7B%22name%22%3A%22flyefish_all-data%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22http%3A%2F%2Flocalhost%3A8000%2Fflyefish_all%2Fdata.csv%22%7D&sourceMetadata=%7B%22name%22%3A%22flyefish_all-cols%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22http%3A%2F%2Flocalhost%3A8000%2Fflyefish_all%2Fcolumns.csv%22) missing the final closing brace, BFF continues to display the loading animation even once an error appears in the console.

## Expected Behavior

There should be an error displayed to the user.

## Reproduction

Click the link above.

## Environment
- Firefox

aswallace:
## Context
closes #517 
When attempting to load a URL with malformed JSON, the `Query` and `FileList` components were getting stuck in a loading state and not showing the error to the user. 

Relatedly, if a user only had a single query and then deleted it, the query panel would also show an eternal loading state because the search params weren't getting removed from the URL (so the `Query` component assumed something was supposed to be loading).

## Changes
JSON loading bug:
- Adds a try/catch around the `decode` step so we can dispatch an error status to the UI
- Checks for reload errors in the Query component so that we stop showing the spinner 

Deleting only query bug:
- Removes query params from URL if there are no more queries left (otherwise, loads the first remaining query in the list)

Also fixes styling for the example table that appears when there are no queries loaded

## Testing
- Manually tested [the url provided in the tagged issue](https://bff.allencell.org/app?c=File+Name%3A0.5%2CBatch%3ANaN%2CRound%3ANaN%2CDate%3ANaN&source=%7B%22name%22%3A%22flyefish_all-data%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22http%3A%2F%2Flocalhost%3A8000%2Fflyefish_all%2Fdata.csv%22%7D&sourceMetadata=%7B%22name%22%3A%22flyefish_all-cols%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22http%3A%2F%2Flocalhost%3A8000%2Fflyefish_all%2Fcolumns.csv%22), plus a few other intentionally malformed URLs
- Added a loading state unit test

### JSON loading bug
Before (stuck in loading) vs after (shows error & stops loading)
Screenshot 2025-05-12 at 12 13 51 PM
Screenshot 2025-05-12 at 12 14 09 PM

### Deleting only query bug
Before (query shows loading state forever) vs after (no search params, default empty display) --EDITED for [cd550c9](https://github.com/AllenInstitute/biofile-finder/pull/522/commits/cd550c9c5af2d3f1c2d96c8f4c5d7a66b80f96d9)
Screenshot 2025-05-12 at 12 11 35 PM
image


kmitcham (approved):

packages/core/state/interaction/logics.ts
```diff
@@ -69,12 +69,22 @@ const initializeApp = createLogic({
...
-            );
+            try {
+                dispatch(
+                    selection.actions.addQuery({
+                        name: DEFAULT_QUERY_NAME,
+                        parts: SearchParams.decode(window.location.search),
+                    })
+                );
+            } catch (err) {
+                // Parsing error from SearchParams.decode
```
BrianWhitneyAI, line 80: This catches all errors correct? does this comment align with that?

BrianWhitneyAI (approved):

SeanLeRoy:
In your "after" it looks like it still has a query selected its just disabled - could we clear that completely?

SeanLeRoy (approved):
Yes!! So glad this is happening

packages/core/state/interaction/logics.ts
aswallace: Technically no, it only catches the decode error since `addQuery` has its own error handling. I could write this equivalently (which may have been more clear) if I separated `const parts = SearchParams.decode()` as its own step and just wrapped that, but then I'd need to separately prevent `addQuery` from running...

aswallace:
> In your "after" it looks like it still has a query selected its just disabled - could we clear that completely?

Realized in looking into this comment that even when resetting the URL/search params, the store was actually persisting the query properties (which is why the disabled query still had the props). So in [cd550c9](https://github.com/AllenInstitute/biofile-finder/pull/522/commits/cd550c9c5af2d3f1c2d96c8f4c5d7a66b80f96d9) I've changed it from modifying `window.location` to instead adding logic that resets all query-related state properties to their default initial values, which fortunately has a side effect of also clearing the url search params.

PR #529, opened 2025-05-16
Bugfix/Date range infinite re-renders

### LINKED ISSUE #528: Using date range crashes page
## Description

Sometimes when I go to filter on a date range, the entire page will hang and then crash? Bad!

## Expected Behavior

Should be able to filter on dates without crashing

## Reproduction

- Open a new BFF app window
- There should already be an 'Uploaded' filter applied
- Attempt to edit the 'Uploaded' filter
- App will crash

Alternatively:
- In any query, try to apply a date range filter with both start and end dates (e.g., Uploaded)

## Environment
- [x] Web
- [x] Desktop
   - Version: [e.g. 8.1.1]

aswallace:
## Context
resolves #528 

Using the date range selection component was causing the app to hang and then crash. It seems we weren't checking whether values had actually changed before triggering the search, so the page was entering a re-render loop that sent infinite network calls. (Reproduce by opening a new BFF window and then trying to edit the `Uploaded` dates)

Also, the UTC conversion code was causing the starting date/time to change every time the DateRangePicker component was rendered (kept bumping it forward by 7 hours), which may have also contributed to the issue. I don't remember this happening before, but I'm unsure if something changed in the way we handle dates since this code was originally added.

## Changes
- Adds a value comparison check before triggering `onSearch`
- Gets rid of the UTC to ISO conversion function. It seems to now be causing issues rather than solving them, but again, unsure if something has changed?

## Testing
- Made sure selecting date ranges (or triggering the component) no longer causes app crashes
- Tested applying date ranges and made sure the returned values were logically consistent (e.g., files uploaded at 12am on the 13th show up when the start date is the 13th)
- An old unit test failed and needed to be updated with this change because 00:00:00Z is technically the previous day in our time zone. I'm sure there's a more elegant way to handle this, so open to suggestions 

### Potential to do?
- This may mean we don't account for different time zones now? Our UTC vs ISO date format situation is confusing me
- Will look into adding more unit tests as needed

### DIFF
--- a/packages/core/components/DateRangePicker/index.tsx
+++ b/packages/core/components/DateRangePicker/index.tsx
@@ -16,21 +16,11 @@ interface DateRangePickerProps {
     currentRange: FileFilter | undefined;
 }
 
-// Because the datestring comes in as an ISO formatted date like 2021-01-02
-// creating a new date from that would result in a date displayed as the
-// day before due to the UTC offset, to account for this we can add in the offset
-// ahead of time.
-export function extractDateFromDateString(dateString?: string): Date | undefined {
-    if (!dateString) {
-        return undefined;
-    }
-    const date = new Date(dateString);
-    date.setMinutes(date.getTimezoneOffset());
-    return date;
-}
-
 const DATE_ABSOLUTE_MIN = new Date();
 DATE_ABSOLUTE_MIN.setFullYear(2000);
+// End of today
+const DATE_ABSOLUTE_MAX = new Date();
+DATE_ABSOLUTE_MAX.setHours(23, 59, 59);
 
 /**
  * This component renders a simple form for selecting a minimum and maximum date range
@@ -50,10 +40,16 @@ export default function DateRangePicker(props: DateRangePickerProps) {
             // To handle that, we subtract a day from the upper bound used by the filter, then present the result
             oldEndDate.setDate(oldEndDate.getDate() - 1);
         }
-        const endOfToday = new Date();
-        endOfToday.setHours(23, 59, 59);
         const newStartDate = startDate || oldStartDate || DATE_ABSOLUTE_MIN;
-        const newEndDate = endDate || oldEndDate || endOfToday;
+        const newEndDate = endDate || oldEndDate || DATE_ABSOLUTE_MAX;
+        // Avoid re-triggering search if the values haven't changed
+        if (
+            newStartDate?.valueOf() === oldStartDate?.valueOf() &&
+            newEndDate?.valueOf() === oldEndDate?.valueOf()
+        ) {
+            return;
+        }
+
         if (newStartDate && newEndDate) {
             // Add 1 day to endDate to account for RANGE() filter upper bound exclusivity
             const newEndDatePlusOne = new Date(newEndDate);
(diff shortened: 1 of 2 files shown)


BrianWhitneyAI (approved):

pgarrison (commented):
I think I'd like some more information. Please feel free to ignore any of these if they are unhelpful.

* I believe MMS stores Dates as YYYY-MM-DD and DateTimes in ISO-8601 with time and timezone. Does BFF receive both formats from FES?
  * If so, it may be that some of the complexity is coming from trying to use a single DateRangePicker component for both dates and datetimes.
  * What format is sent by FES for the date you are using for your "12am on the 13th"?
* What user actions would cause a datetime in UTC+0 or UTC+8 to be sent to the DateRangePicker?
* I have a hunch that the test you modified will fail it's run on a computer in UTC+9, is that right?
  * In what circumstances will BFF users in different timezones get different results?
  * We may be interested in explicitly [setting the timezone](https://github.com/marketplace/actions/set-timezone) of CI

It seems like the case you tested manually of a file uploaded at 12am on the 13th showing up when the start date is the 13th would be a good unit test to add.

kmitcham (approved):
Timezones are always a pain.

aswallace:
> I believe MMS stores Dates as YYYY-MM-DD and DateTimes in ISO-8601 with time and timezone. Does BFF receive both formats from FES?

Both appear to use the same format, I think?
[Querying FES for "Processing date"](https://production.int.allencell.org/file-explorer-service/1.0/annotations/hierarchy/root?order=Processing%20Date) (which [MMS confirms](http://aics/metadata-management-service/1.0/annotation?annotation=processing%20date) is a `Date` type) returns values like `"2018-09-28T00:00:00.000+00:00","2017-12-09T00:00:00.000+00:00"`. 
[Querying on "Collection date"](https://production.int.allencell.org/file-explorer-service/1.0/annotations/hierarchy/root?order=Collection%20Date) (which [MMS confirms](http://aics/metadata-management-service/1.0/annotation?annotation=collection%20date) is a `DateTime`) returns values like `"2020-12-15T12:40:48.000+00:00", "2023-08-03T20:00:00.000+00:00",`

aswallace:
@pgarrison additional info: Looks like the FluentUI DatePicker component automatically selects a time-- e.g, if I use the UI component to select a `Processing Date` start date of `February 01 2020` and end date of `March 01 2024`, then the filter that actually gets applied is `Processing Date=RANGE(2020-02-01T08:00:00.000Z,2024-03-02T08:00:00.000Z)`

Manually changing my computer's time zone to Anchorage (UTC-9) does indeed cause the tests to fail, though the same filter gets applied when I select the same dates in the UI (`Processing Date=RANGE(2020-02-01T08:00:00.000Z,2024-03-02T08:00:00.000Z)`)

pgarrison:
> though the same filter gets applied when I select the same dates in the UI (Processing Date=RANGE(2020-02-01T08:00:00.000Z,2024-03-02T08:00:00.000Z))

@aswallace Curious. Perhaps BFF identifies the timezone when it starts up? Or when it's installed? I expect we aren't hardcoding BFF to PST?

aswallace:
> Perhaps BFF identifies the timezone when it starts up? Or when it's installed? I expect we aren't hardcoding BFF to PST?

@pgarrison Turns out that I just wasn't setting timezone properly for browser since it takes location from IP rather than computer settings. Modifying tz at a browser level does correctly change the TZ shift in the filter.

However, we also actually **do** set it to "America/Los Angeles" for our formatter... which was added 5 years ago, so I suspect it's a relic of when the tool was only for AICS data? Also looks like we've thought about this before based on [this comment](https://github.com/AllenInstitute/biofile-finder/blob/40e52371f01e87e7ddbe41590ea5eae6e238118b/packages/core/entity/AnnotationFormatter/date-time-formatter.ts#L1-L8)

SeanDuHare (approved):

Issue #531, opened 2025-05-22
File name fails to generate for partial columns


aswallace:
## Description

When a dataset has a "File Name" column but some or all of the rows don't actually have a value for name (e.g., the column is empty), performing any action that requires `FileDetail.name` causes the app to error out. This includes selecting a file since the FileDetails panel requires name, and also switching to the thumbnail view since we label the thumbnails with the name.

Originally identified [by Antoine](https://allencellscience.slack.com/archives/C073UPM2EKD/p1747783535972709)

## Expected Behavior

We have options...
1. Check early on if "File Name" exists but has null values, and force the user to either fix or remove it
2. Check if "File Name" exists but has null values at the same time that we currently auto-generate file names when the column DOESN'T exist, and then selectively auto-generate names just for the rows that don't have them
3. Don't bother checking, and generate a file name on demand based on File Path (which is the one required column)
4. Something else?

## Reproduction

Upload a csv like the following: 

  File Path | File Name | File ID
------|------|-----
some/path/abc.png | abc | 1
some/path/no-name-provided.png | | 2
some/path/def.png | def | 3

Selecting the second file (File ID 2) will cause the app to throw an error

<img width="869" alt="Image" src="https://github.com/user-attachments/assets/24001669-c166-4485-b2f5-003117f27fef" />

## Environment
_Which version of BioFile Finder has the bug and with what environment?_
- [ ] Web
   - Browser: [e.g., Chrome, Safari]
- [ ] Desktop
   - OS: [e.g. Windows, Linux, Mac with Apple chip, Mac with Intel chip]
   - Version: [e.g. 8.1.1]

PR #532, opened 2025-05-22
Autogenerate missing file names

aswallace:
## Context
One potential option for resolving #531.
The EMT csv that Antoine was uploading had an empty File Name column, which meant that our database creation step skipped the conditional that would have caused it to autogenerate file names from the file paths.
Full context and potential options in issue description. This is option 3 from that list (quickest fix and lowest effort but feels slightly hacky)
Just opening as a draft right now because we likely want to explore other options

## Changes
If there isn't a file name, before throwing an error, first try generating one on the spot based on the file path

## Testing
If we think we might go with this, I'll spend more time on unit testing

Tested manually with
- the dataset Antoine provided where there's a "File Name" column but no values in it
- another simpler one I made where some but not all of the files have values for "File Name"

### DIFF
--- a/packages/core/entity/FileDetail/index.ts
+++ b/packages/core/entity/FileDetail/index.ts
@@ -121,7 +121,13 @@ export default class FileDetail {
     public get name(): string {
         const name = this.fileDetail.file_name || this.getFirstAnnotationValue("File Name");
         if (name === undefined) {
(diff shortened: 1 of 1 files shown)


aswallace:
Closing in favor of #533

PR #520, opened 2025-05-09
Feature/collapse large annotations

aswallace:
## Context
Addresses a few UI/UX issues, but mainly being able to collapse/expand super long metadata values
closes #168 , _hopefully_ closes #506 

Available for testing on [staging](https://staging.bff.allencell.org/app?c=file_name%3A0.4%2CKind%3A0.2%2CType%3A0.25%2Cfile_size%3A0.15&group=Acquisition+Time+by+Timepoint&group=Notes&filter=%7B%22name%22%3A%22uploaded%22%2C%22value%22%3A%22RANGE%282024-05-09T07%3A00%3A00.000Z%2C2025-05-10T06%3A59%3A59.558Z%29%22%2C%22type%22%3A%22default%22%7D&source=%7B%22name%22%3A%22AICS+FMS%22%7D&sort=%7B%22annotationName%22%3A%22uploaded%22%2C%22order%22%3A%22DESC%22%7D&showNulls=true)

#### Main change:
Added expand/collapse logic for metadata values that wrap to more than 4 lines. 
After discussing with @lynwilhelm, departed slightly from the design in the issue description for logistic/style consistency reasons
image
image

#### Other changes:
- Truncates super long folder names (from the center since end bit is often relevant/distinctive) when grouping and adjusted their alignment. I didn't end up making these expandable because I didn't want to add another clickable element on the folder, but I'm hoping it won't be super common to group on fields that are that long anyway
image
image

- HOPEFULLY fixes the file row height issue
    - Made the row height slightly taller and set a line-height 
- Hides the pseudo element we use to capture focus  (white input in screenshot, shows up when you collapse the query panel-- was missing units on the css):
image

## Testing
- Unit tests for expand/collapse metadata values

Manually tested:
- expand/collapse for various long metadata values, made sure is responsive to changes in the panel size
- line height: tried zooming in/out on different browsers, @thao-do verified on staging deployment

PR #535, opened 2025-05-23
Bugfix/slicing non string folder names

### LINKED ISSUE #534: Duration grouping crashes app
## Description

identified [by Antoine](https://allencellscience.slack.com/archives/C073UPM2EKD/p1747936437818559)
Grouping by Time Since CHIR (or other durations that include a zero group) causes the app to crash with `c.slice is not a function`. 
Result of undiscovered bug from #520 because the duration formatter formats "0" as a number instead of as a string (despite the prop being typed as a string)

## Expected Behavior

Should be able to group on any annotation type

## Reproduction

Group on "Time Since CHIR" on web or recent desktop versions

## Environment
- [x] Web
- [x] Desktop
   - Version: 8.4.3

aswallace:
## Context
resolves #534 
Grouping on duration-type annotations that have "0" as one of the group values was causing the app to error out. Our duration formatter was returning the display value of "0" as a number instead of as a string, which meant that `DirectoryTreeNodeHeader` would throw an error when attempting to `slice` it.

## Changes
- Updated the duration formatter
- Added a "just in case" `.toString()` function in the DirectoryTreeNodeHeader to avoid running into something similar with other annotation types

## Testing
- Added unit tests that were missing from #520 and updated formatter unit tests


kmitcham (approved):

SeanLeRoy (approved):
LGTM!

Issue #197, opened 2024-08-06
Improve styling of queries in collapsed state


lynwilhelm:
It's been noticed by Graham and design that the queries have a few styling issues:

1. Need more contrast (consider box shadow to help each look more distinct)
2. All of the text creates quite a bit of noise. Not sure if ALL of the info is necessary to show in collapsed state (see example below)
3. The chevron is missing (in collapsed state it should be pointing down, in open state should be pointing up)
4. Titles may need to stand out more (Graham suggested just a simple underline)
5. Do all of the filter values need to show? Would count of filers applied suffice?

Sean suggested asking users what they would need when creating multiple queries. Is all of the filter and sorting info helpful or is there a reduced version that makes sense? Maybe not?

![Screenshot 2024-08-06 at 4 23 07 PM](https://github.com/user-attachments/assets/044f1215-4107-4caf-b642-5f01c9495013)

aswallace:
Issues 1 and 3 are addressed by #294

lynwilhelm:
To close this ticket out, there are 2 items to change:
1. underline styling to title
2. We will change what is shown in filters to match the design shown here. This helps save space/reduce noise in the collapsed state.

![Image](https://github.com/user-attachments/assets/edaaf565-fcda-4950-a284-18132a34cc9f)

This design is updated in the Figma master styles.

lynwilhelm:
@aswallace @SeanLeRoy Can we discuss this ticket and assign priority? Or do we create a new implementation ticket? I am not suggesting it should get implemented just because I chose to move it across the finish line. I just vetted with Chris F and Antoine in a couple hours today because it was low hanging fruit that's been assigned to me forever.

PR #538, opened 2025-06-04
(Small) Feature/improve styling of collapsed queries

aswallace:
## Context
Updating appearance of collapsed queries in the sidebar per #197 

## Changes
Followed the designs in Figma and added some processing on the filters so that different information appears depending on the filter type (e.g., regular filters should show a count of the values, any/none filters should show "Any value" or "No value" respectively, and range filters shouldn't show anything).

## Testing
Added a couple unit tests

## Screenshots 
Before vs after with multiple applied filters
Screenshot 2025-06-04 at 12 32 35 PM
Screenshot 2025-06-04 at 12 32 40 PM

Hover state
Screenshot 2025-06-04 at 12 32 49 PM

#### Before vs after for a filter that has an extremely long value
Screenshot 2025-06-04 at 1 22 55 PM
Screenshot 2025-06-04 at 1 23 08 PM

### DIFF
--- a/packages/core/components/QuerySidebar/Query.tsx
+++ b/packages/core/components/QuerySidebar/Query.tsx
@@ -11,6 +11,7 @@ import QueryGroup from "../QueryPart/QueryGroup";
 import QuerySort from "../QueryPart/QuerySort";
 import Tooltip from "../Tooltip";
 import { AICS_FMS_DATA_SOURCE_NAME } from "../../constants";
+import { FilterType } from "../../entity/FileFilter";
 import { interaction, metadata, selection } from "../../state";
 import { Query as QueryType } from "../../state/selection/actions";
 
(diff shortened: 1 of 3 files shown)


packages/core/components/QuerySidebar/Query.tsx
```diff
@@ -128,9 +130,33 @@ export default function Query(props: QueryProps) {
                 )}
                 {!!queryComponents.filters.length && (
                     <p className={styles.displayRow}>
-                        <strong>Filter:</strong>{" "}
-                        {queryComponents.filters
-                            .map((filter) => `${filter.name}: ${filter.value}`)
+                        <strong>Filter{queryComponents.filters.length > 1 ? "s" : ""}:</strong>{" "}
+                        {/* Show a condensed version of the filters list with counts instead of values */}
+                        {Object.entries(
+                            queryComponents.filters.reduce((accum, filter) => {
```
SeanLeRoy, line 136: This is pretty busy to put into the component tree here - could we pull this out as a function that this calls instead?

SeanLeRoy (approved):

PR #550, opened 2025-06-30
Feature/warn unsaved edits

### LINKED ISSUE #392: Metadata editing: Warn users when exiting w unsaved changes
When editing metadata in BFF, if the datasource is anything other than AICS (e.g., an uploaded csv or other url), the changes will not persist when the page is closed. 
We should track when users have made changes to datasources in DuckDB. Then when users attempt to leave the page, we should provide an alert/warning so they know their changes will be lost unless they re-download the metadata as a csv.
Related to #391

aswallace:
## Context
When users perform edits on non-FMS data sources, the changes aren't permanent (e.g., they only exist in the temporary DuckDB version of the data). If the page is reloaded or exited, they will lose any edits they've made.
Therefore, we want to warn users before they leave/reload. 

resolves #392 

## Changes
- Re-enables editing for non-FMS data sources (was hidden)
- Adds a before unload event for both web and desktop that pops up the browser's alert message.
   - Modern web browsers don't let us customize the alert message, so per UX/design discussion also added in an info message ONLY after the first time a user makes an edit to a non-FMS file (see screenshot) 
   - Desktop needs an additional event handler (per #514, Electron doesn't show alerts from these events to the user). This DOES allow us to customize the message

## Testing
Manually tested by making edits to uploaded datasets and then trying to reload and/or close the window

## Screenshots
Info message after initial edit (non-FMS source)
image

Chrome v121 reload/exit alert (not customizable)
image

Firefox v139 reload/exit alert (not customizable)
image

Desktop reload/exit alert (customized message)
image

### DIFF
--- a/packages/core/App.tsx
+++ b/packages/core/App.tsx
@@ -47,6 +47,7 @@ export default function App(props: AppProps) {
 
     const dispatch = useDispatch();
     const hasQuerySelected = useSelector(selection.selectors.hasQuerySelected);
+    const hasUnsavedChanges = useSelector(interaction.selectors.getHasUnsavedChanges);
     const requiresDataSourceReload = useSelector(selection.selectors.getRequiresDataSourceReload);
     const shouldDisplaySmallFont = useSelector(selection.selectors.getShouldDisplaySmallFont);
     const platformDependentServices = useSelector(
@@ -55,6 +56,23 @@ export default function App(props: AppProps) {
     const [measuredNodeRef, _measuredHeight, measuredWidth] = useLayoutMeasurements<
         HTMLDivElement
     >();
+
+    React.useEffect(() => {
+        const beforeUnloadHandler = (event: BeforeUnloadEvent) => {
+            // Modern browser alert: Does not allow custom messages
+            event.preventDefault();
+
+            // Legacy support (e.g. Chrome/Edge < 119): Allows custom messages
+            event.returnValue =
+                "Edits to external data sources will not be saved. Make sure to download the result in order to keep your changes";
+        };
+
+        if (hasUnsavedChanges) window.addEventListener("beforeunload", beforeUnloadHandler);
+        // remove the event listener
+        return () => {
+            window.removeEventListener("beforeunload", beforeUnloadHandler);
+        };
+    }, [hasUnsavedChanges]);
 
     // Check for updates to the application on startup
     React.useEffect(() => {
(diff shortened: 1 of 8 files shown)


packages/core/App.tsx
```diff
@@ -56,6 +57,24 @@ export default function App(props: AppProps) {
...
 
+    React.useEffect(() => {
+        const beforeUnloadHandler = (event: BeforeUnloadEvent) => {
+            // Modern browser alert: Does not allow custom messages
+            event.preventDefault();
+
+            // Legacy support (e.g. Chrome/Edge < 119): Allows custom messages
+            // Deprecated, does not affect modern browsers
+            event.returnValue =
+                "Edits to external data sources will not be saved. Make sure to download the result in order to keep your changes";
```
aswallace, line 68: Not attached to this wording, essentially just a placeholder. Open to suggestions

packages/core/state/interaction/logics.ts
```diff
@@ -634,6 +638,15 @@ const editFilesLogic = createLogic({
             }
             dispatch(refresh); // Sync state to pull updated files
             dispatch(processSuccess(editRequestId, "Successfully edited files."));
+            if (!hasUnsavedChanges && !isQueryingAicsFms) {
+                dispatch(setHasUnsavedChanges());
+                dispatch(
+                    processInfo(
+                        "edit-info-message",
+                        "Edits made to external data sources are not permanent. Make sure to save/download data to keep your edited versions."
```
aswallace, line 646: Also open to suggestions on this phrasing

BrianWhitneyAI (approved):

packages/desktop/src/main/index.ts
```diff
@@ -107,6 +107,24 @@ const createMainWindow = () => {
...
+     * Catch 'beforeunload' events since Electron handles these differently from browser
+     * Similar to above, but for main window instead of child window
+     */
+    mainWindow.webContents.on("will-prevent-unload", (ev: Event) => {
+        const options = {
+            type: "question",
+            buttons: ["Leave", "Cancel"],
+            message: "Leave site?",
+            detail:
+                "Edits to external data sources may not be saved. Download query results to save edits.",
```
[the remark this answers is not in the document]
aswallace: Yep, I'll make sure they all say the same thing (or point to the same variable) before I merge. The triplication is just because of the annoying difference in old vs new browser policies plus the way electron blocks events

packages/core/App.tsx
kmitcham: Could we add a 'save' button to the alert?  I suppose it might be premature.  What happens if the user makes a series of changes?  Will the alert pop up every time?

aswallace: Combined Sean's rephrasing with Lyndsay's suggestions in [5e1a136](https://github.com/AllenInstitute/biofile-finder/pull/550/commits/5e1a136d1c5039f1fac9a1ecc7a302ab0c214330)

Issue #428, opened 2025-02-11
Add drag and drop and improve "Choose a data source" design and behavior


lynwilhelm:
**Background:** 
This work was initiated by Lyndsay to improve the design of the CSV (table) example, content, and add easier way to select a data source (drag and drop)

**Acceptance criteria:**
- [x] New drag+drop functionality (can accept single ~or multiple files at once~)
- [x] Load button location and behavior (only show when there is something to load)
- [x] Apply (max width) to file selection area section
- [x] New example CSV styling
- [x] Reordering and revisions of content per design
- [x] Add ~another data source~ metadata descriptors file behavior (only show when there is something to load)
- [x] #536 

[LINK TO DESIGN](https://www.figma.com/design/oNQvOAK11YINcbPvNmkgwO/Public-File-Explorer?node-id=3535-20654&t=MvVSYHlVEAqdp39i-0)

PR #552, opened 2025-07-09
Update typescript and eslint

aswallace:
## Context
One of the minor bugs I'm encountering in #428 is an external typescript error for `Intl.format` that was fixed in 4.7, and we currently use 4.6 (technically 4.x)

## Changes
Updating to 4.7 created a conflict with `@typescript-eslint/typescript-estree`: 
```
WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.7.0
```
so also needed to update `eslint`. In updating `eslint`, the recommended version (8.57, which is the last 8.x) in combo with the necessary `@typescript-eslint` updates only supported down to typescript 4.8... so ended up bumping typescript from 4.x (4.6) to 4.8.

EDIT: Pushed typescript update to 4.9 since it works out of the box and is the last 4.x before 5.0. Beyond that, would need to update eslint to 9.x which is breaking for us (but also shouldn't be too hard to fix)

We're technically still behind the latest versions of these, but I was trying to minimize the version changes overall. 

## Testing
Verified that after installing new dependency versions, tests and lint still pass. Also made sure can still run/render web and desktop


aswallace:
>  did you see what else breaks for the next versions up? 

8.57.x is the last 8.x version of eslint, and when I tried 9 we ended up with a lot of linting errors. They probably just changed how/where you set which rules you want to ignore, but would require a little bit more effort which should maybe be a separate ticket.
With typescript I could technically go up to 5.1.x without issues--going past that requires updating eslint

pgarrison (approved):

PR #560, opened 2025-07-25
Bugfix/Make queries with LIMIT clauses return consistent results

### LINKED ISSUE #553: File metadata getting overwritten/replaced in group
## Description

Identified by Rennie [here](https://allencellscience.slack.com/archives/C021CEY96LD/p1752515582664039)
For certain queries, clicking between files occasionally causes the metadata to "overwrite" in the file list. 

This appears to be related to sort behavior, since I'm only able to reproduce it when I have the sort applied, and particularly if I'm sorting on a boolean column

## Expected Behavior
Users should be able to click between files and have data persist

## Reproduction

Go to [this link](https://bff.allencell.org/app?c=Single+Colony+Or+Lumenoid+At+Time+of+Migration%3A0.25%2CScene+Index%3A0.25%2CPosition+Index%3A0.25%2CFile+Name%3A0.25&group=Plate+Barcode&openFolder=%5B%223500007062%22%5D&source=%7B%22name%22%3A%22pub+%287%2F7%2F2025+2%3A12%3A39+PM%29%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22https%3A%2F%2Fdocs.google.com%2Fspreadsheets%2Fd%2Fe%2F2PACX-1vQ9syUr5OHtA0jYGrKezEzhb8Al0NOnzotQYNQiqM4BDzgL7xb7XRefkPpI7YzB2hBK9QM5i7FHjNyf%2Fpub%3Foutput%3Dcsv%22%7D&sort=%7B%22annotationName%22%3A%22Absence+Of+Migrating+Cells+Coming+From+Colony+Out+Of+FOV+At+Time+Of+Migration%22%2C%22order%22%3A%22DESC%22%7D), and click the file with scene index 8. Then scroll down to the bottom of that group and click on the file with scene index 49 (and then the files below it). The first few files will be replaced with the metadata for scene index 8, and the ones after that will be replaced with the metadata for scene index 35

## Environment
- [x] Web
   - Was able to reproduce on all tested browsers (Chrome, FireFox, Safari)
- [x] Desktop
   - Was able to reproduce on desktop by using the same csv as the query in the link above and then following the same steps

aswallace:
## Context
closes #553 
For certain queries, file metadata was getting "overwritten" in the file list when users selected files at the bottom of the list. See [this link](https://bff.allencell.org/app?c=Single+Colony+Or+Lumenoid+At+Time+of+Migration%3A0.25%2CScene+Index%3A0.25%2CPosition+Index%3A0.25%2CFile+Name%3A0.25&group=Plate+Barcode&openFolder=%5B%223500007062%22%5D&source=%7B%22name%22%3A%22pub+%287%2F7%2F2025+2%3A12%3A39+PM%29%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22https%3A%2F%2Fdocs.google.com%2Fspreadsheets%2Fd%2Fe%2F2PACX-1vQ9syUr5OHtA0jYGrKezEzhb8Al0NOnzotQYNQiqM4BDzgL7xb7XRefkPpI7YzB2hBK9QM5i7FHjNyf%2Fpub%3Foutput%3Dcsv%22%7D&sort=%7B%22annotationName%22%3A%22Absence+Of+Migrating+Cells+Coming+From+Colony+Out+Of+FOV+At+Time+Of+Migration%22%2C%22order%22%3A%22DESC%22%7D) for reproduction; scroll to the bottom of the open group and click different files in the list -- file metadata will appear to be replaced. 

The results of using a `LIMIT` clause are [not necessarily deterministic](https://duckdb.org/docs/stable/sql/query_syntax/limit) without an `ORDER BY` clause, which meant that the query we used to return the lazy-loaded subset of the file list was in a different order than the query that's run when we select a single file. Even though both queries were technically sorting on the same field, the sort order WITHIN that sort was different (e.g., if sorting on "Color" and all the files have "blue," the files won't necessarily be in the same order with and without a limit applied unless there's a secondary sort on a unique field)

## Changes
Enforces a secondary/default sort on our internal unique ID field. If we're performing a query with a `LIMIT` and there isn't a sort applied, adds one. 
Also changed the order of `LIMIT` and `OFFSET` per DuckDB [syntax recommendations](https://duckdb.org/docs/stable/sql/query_syntax/limit#syntax)

## Testing
Validated manually using the example provided in the issue. We don't currently have a clear way of automating a test for whether the results are returned in the same order or not.

I did not encounter a noticeable slowdown in query efficiency, but I also did not specifically test for it. Can see about manually measuring performance if there are concerns, but we don't currently have performance testing infrastructure set up for this codebase

### DIFF
--- a/packages/core/entity/SQLBuilder/index.ts
+++ b/packages/core/entity/SQLBuilder/index.ts
@@ -95,10 +95,11 @@ export default class SQLBuilder {
             throw new Error("Unable to build SQL without a FROM statement");
         }
         // LIMIT is non-deterministic without sorting
-        // So even if there is already an "order by" clause, secondarily sort on unique ID
+        // So even if there is already an "order by" clause, secondarily sort on unique ID.
+        // Exception: COUNT(*) queries should not require sorting
         let orderByString = "";
-        if (this.orderByClause || this.limitNum) {
-            orderByString = `ORDER BY ${this.orderByClause}${
+        if (this.orderByClause || (this.limitNum && !this.selectStatement.includes("COUNT(*)"))) {
+            orderByString = `ORDER BY ${this.orderByClause || ""}${
                 this.orderByClause && this.limitNum ? ", " : ""
             }${DatabaseService.HIDDEN_UID_ANNOTATION}`;
         }


kmitcham (approved):
I once heard an anecdote about an API team intentionally randomizing returned data as a distinct step since the specs didn't claim the response would be ordered, and the team didn't want clients to start relying on the default order of returned data.  I don't think that was the source of this issue, though.

PR #569, opened 2025-08-20
Bugfix/error on load datasource

aswallace:
## Context
For our open-source datasets, the app sometimes flashes an error status before loading the dataset. Reproduce by going [here](https://bff.allencell.org/datasets) and loading a large dataset (e.g., "Structural organization and gene...").
This is because when we load a dataset, the app tries to add the query via dispatch ("Process A") and also by reading the query search params ("Process B"), which creates parallel processes that conflict with each other:
1. User selects a dataset to load
2. "Process A" starts loading the dataset, and creates a table in the database with the source name. Meanwhile, the page redirects to the app's current search params (usually empty). 
3. Process A updates the search params. The app registers that the params change, reads the query params, and starts also trying to load the dataset via "Process B"
5. Process B sees that the dataset is already present, has a little panic (`Table with name <name> already exists`), tries to clean itself up, and accidentally deletes the dataset loaded from Process A
6. Process A now can't find its data (because B deleted it), and has its own panic (`Table with name <name> does not exist`) --> this is the error the user sees briefly
7. This is somewhat of a guess since hard to track: The query args are still present in the url, so we try one more time to load the query, and this time it succeeds

These conflicting processes were also creating a secondary bug, where if you loaded a dataset, navigated backwards to the dataset page, and then added a new dataset, old queries would erroneously get added multiple times.

## Changes
When we initialize the app, instead of immediately adding a new query from the search params, first check if the query has already been added via redux. If it has, just reload the existing query instead of adding it again.

Semi-related changes: 
- When a new query was being loaded, the file count sometimes showed a stale count. Now it displays the loading state until a new count comes back
- Collapsed query titles were overflowing their bounds instead of using ellipses 
Before vs after:
image
image

### DIFF
--- a/packages/core/components/FileList/index.tsx
+++ b/packages/core/components/FileList/index.tsx
@@ -131,6 +131,8 @@ export default function FileList(props: FileListProps) {
 
     // Get a count of all files in the FileList, but don't wait on it
     React.useEffect(() => {
+        // Show loading state in UI instead of stale count (could be from unrelated query set)
+        setTotalCount(null);
         let cancel = false;
         fileSet
             .fetchTotalCount()
--- a/packages/core/state/interaction/logics.ts
+++ b/packages/core/state/interaction/logics.ts
@@ -78,12 +78,21 @@ const initializeApp = createLogic({
         // into the query to render (ex. when refreshing a page)
         if (isOnWeb && window.location.search) {
             try {
-                dispatch(
-                    selection.actions.addQuery({
-                        name: DEFAULT_QUERY_NAME,
-                        parts: SearchParams.decode(window.location.search),
-                    })
-                );
+                // If the query is already present, we shouldn't add it again, just refresh it
+                const equivalentExists = queries.some((query) => {
+                    return `?${SearchParams.encode(query?.parts)}` === window.location.search;
+                });
+                // Refresh existing query
+                if (queries.length && equivalentExists) {
+                    dispatch(refresh);
+                } else {
+                    dispatch(
+                        selection.actions.addQuery({
+                            name: DEFAULT_QUERY_NAME,
+                            parts: SearchParams.decode(window.location.search),
+                        })
+                    );
+                }
             } catch (err) {
                 // Parsing error from SearchParams.decode
                 dispatch(
--- a/packages/core/components/QuerySidebar/Query.module.css
+++ b/packages/core/components/QuerySidebar/Query.module.css
@@ -72,11 +72,16 @@
     justify-content: space-between;
 }
 
-.header > h4 {
+.header h4 {
     overflow: hidden;
     text-overflow: ellipsis;
     width: 100%;
     white-space: nowrap;
+    line-height: 1.5;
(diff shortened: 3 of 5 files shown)


kmitcham (approved):

packages/core/components/FileList/index.tsx
```diff
@@ -131,6 +131,8 @@ export default function FileList(props: FileListProps) {
 
     // Get a count of all files in the FileList, but don't wait on it
     React.useEffect(() => {
+        // Show loading state in UI instead of stale count from previously rendered query results
+        setTotalCount(null);
```
BrianWhitneyAI, line 135: The comment to `setTotalCount(null)` doesnt seem easy to interpret to me.

packages/core/state/interaction/logics.ts
```diff
@@ -70,20 +70,30 @@ const initializeApp = createLogic({
...
-                        parts: SearchParams.decode(window.location.search),
-                    })
-                );
+                // If the exact query already exists in state on initialization,
+                // we shouldn't add it again, just refresh it
+                const equivalentExists = queries.some((query) => {
+                    return `?${SearchParams.encode(query?.parts)}` === window.location.search;
+                });
+                // Refresh existing query
+                if (queries.length && equivalentExists) {
```
BrianWhitneyAI, line 87: does equivalentExists imply queries.length here?

BrianWhitneyAI (approved):

packages/core/state/interaction/logics.ts
aswallace: ah yes, I think this was a legacy from debugging. fixed in [be28c24](https://github.com/AllenInstitute/biofile-finder/pull/569/commits/be28c24513166bd9e1749a56f180ee0a0b04d217)

packages/core/components/FileList/index.tsx
aswallace: updated in [be28c24](https://github.com/AllenInstitute/biofile-finder/pull/569/commits/be28c24513166bd9e1749a56f180ee0a0b04d217)

packages/core/state/interaction/logics.ts
```diff
@@ -70,20 +70,31 @@ const initializeApp = createLogic({
...
-                );
+                // If the exact query already exists in state on initialization,
+                // we shouldn't add it again-- just refresh the existing query
+                if (
+                    queries.some(
+                        (query) =>
+                            `?${SearchParams.encode(query?.parts)}` === window.location.search
+                    )
+                ) {
+                    dispatch(refresh);
```
SeanLeRoy, line 89: Why do we need the refresh?

SeanLeRoy (approved):

packages/core/state/interaction/logics.ts
aswallace: It was a precautionary measure in case the data had become out of date and/or the correct query results weren't being rendered, but may not be necessary -- will test

PR #577, opened 2025-08-28
Feature/Handle server errors when loading data sources

### LINKED ISSUE #563: Error handling for csv or parquet loading
## Use Case

We are looking at generating BFF tables on the fly from IDR / OMERO, which means that we occasionally get errors. Currently this shows up like:

Image

## Suggest a Solution

It would be nice if we could give users (or even devs/testers) some idea of what went wrong, or maybe a link to click on to display the error in another page. 

E.g. return a 500 response with some text to display to the user.

aswallace:
@will-moore 
## Context

Resolves #563 . When DuckDB attempts to create a table from a data source, if the server fails to return data, DuckDB does not pass along the error status or provide informative messages. This means users can't see what went wrong -- it just triggers the "Missing File Path column" error.

## Changes

Unfortunately, there's not a clear way to make DuckDB more descriptive. Instead, when the step to create a table fails, if the source is a url (i.e., not a `File` object), we can run our own `axios` check to retrieve error info from the server directly.

For this PR I'm just clipping the error message to a maximum of 200 characters (approx. 3 lines), but we may consider making the message expandable or allowing users to open the message in a separate dialog/tab
image

## Testing

Added unit tests to make sure errors are dispatched for different stubbed error responses. 

Also manually tested by using a tiny fake server that sends different error formats for different endpoints.

### DIFF
--- a/packages/core/state/selection/test/logics.test.ts
+++ b/packages/core/state/selection/test/logics.test.ts
@@ -4,28 +4,32 @@ import {
     mergeState,
     ResponseStub,
 } from "@aics/redux-utils";
+import axios from "axios";
 import { expect } from "chai";
 import { get as _get, shuffle } from "lodash";
 import sinon from "sinon";
 
 import {
     addFileFilter,
-    selectFile,
+    addQuery,
+    changeDataSources,
+    changeSourceMetadata,
+    decodeSearchParams,
+    expandAllFileFolders,
     reorderAnnotationHierarchy,
     removeFileFilter,
     removeFromAnnotationHierarchy,
+    setAnnotationHierarchy,
+    selectFile,
+    selectNearbyFile,
+    ADD_DATASOURCE_RELOAD_ERROR,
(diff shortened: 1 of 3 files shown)


packages/core/state/selection/test/logics.test.ts
```diff
@@ -680,6 +684,180 @@ describe("Selection logics", () => {
...
+                    type: ADD_DATASOURCE_RELOAD_ERROR,
+                    payload: {
+                        dataSourceName,
+                        error: "Unknown error while adding query",
+                    },
+                })
+            ).to.be.true;
+        });
+    });
+
```
kmitcham, line 860: A test with almost no value would be confirming the 200 character truncation works.

aswallace: Will add, can't hurt! It's an imported method and I suppose dependencies do change

BrianWhitneyAI (approved):

PR #579, opened 2025-09-04
Bugfix/error selecting queries before loaded

### LINKED ISSUE #571: Error on  clicking datasets before they have loaded
## Description

When users load large datasets, if they click on the query before it is loaded, an error appears because the source hasn't been registered to the database yet

Image

## Expected Behavior

Users should not see the error. Either they shouldn't be able to click while it's loading, or clicking on it should just continue to show a loading state

## Reproduction

Add a large dataset (e.g., 100MB). Click on the query that appears before it loads on its own

## Environment
- [x] Web
- [x] Desktop

aswallace:
## Context
Resolves #571. When users load large files that take a long time to load, if they try to click on the newly created query before DuckDB has finished creating a new table for that data source, an error temporarily appears because the database doesn't recognize the table yet. 

## Changes
Per UX recommendation, adds an overlay that prevents user interactions with the app *just* while the new table is being created. Then we allow user interactions and display the usual loading UI while the query retrieves results from that table.

Note that the overlay is currently in `App.tsx`, so the header stays clickable even while the data source is loading. If this is not desired behavior, I should be able to move the overlay to the root component. 

## Testing
Tested manually
- Loading a large data source when other queries are present
- Loading a large data source when no other queries are present
- Selecting a data source that will fail (e.g., has no file path column) to make sure error status messages are still displayed and that the loading overlay goes away
- Selecting small data sources that load quickly 

## Screen recording 
Before: With bug where error appears on user interaction

[before-video](https://github.com/user-attachments/assets/ac263e3a-4947-47a3-acad-33bc07cb4876)

After: With overlay until DuckDB has created the new table

[after-video](https://github.com/user-attachments/assets/9d8120cb-f05b-4415-8326-84e674aef7ef)


packages/core/components/Icons/LoadingIcon.tsx
```diff
@@ -16,7 +17,8 @@ export default function LoadingIcon(props: Props) {
     return (
         <Spinner
             className={classNames(props.className, styles.spinner, {
-                [styles.invertColor]: props.invertColor,
+                [styles.invertColor]: props.invertColor && !props.highlightColor,
+                [styles.highlightColor]: props.highlightColor,
```
SeanLeRoy, line 21: maybe color should just be an enum at this point? seems odd highlightColor being true results in the same effect regardless of whether invertColor is true otherwise

SeanLeRoy (commented):
This _**might be too technically difficult to make it worth it**_, but what do you think about making it so when the "Loading" screen leaves the file list is done loading as well? It feels a little weird to have a loading screen and then still be loading

aswallace:
> This might be too technically difficult to make it worth it, but what do you think about making it so when the "Loading" screen leaves the file list is done loading as well? It feels a little weird to have a loading screen and then still be loading

I agree that it's weird, I had briefly looked into keeping it there while files are loading but ran into some issues with it. I'll take another look, I think it should be doable

kmitcham (approved):

aswallace:
Closing in favor of #589

PR #586, opened 2025-09-11
Feature/process files modal

### LINKED ISSUE #230: <Placeholder>
> This might be too unrelated, but in a lot of beta user interviews they ask about where and how we do some of our images modifications alluding to tools like bioio and segmentation algs. I’ve been wondering if it would be powerful to provide a UI in BFF that would provide users the exact Python script to run to use on their files or perhaps create a backend that just runs the script for them and streams back the image data.

An example workflow of a user wanting to convert to .zarr would be:
- Upload .csv containing metadata/references to files
- Select files to convert
- Right click
- Select a new menu option: “Process” (might not be the best word here)
- Select which type of file processing, in this case: “File type conversion”
- A modal appears with some prompts about the files to convert (like optional options about physical pixel size for example)
- A Python script appears that is automatically filtered to just the selected files and will perform the .zarr conversion as selected

BrianWhitneyAI:
## Description 

The purpose of this PR is to open an initial branch to start developing processing workflows from BFF. The goal is to explore and resolve #230. This branch will be the base for additional changes.  

This PR adds in a code snippet generation for extracting standard metadata from images. Standard metadata is a suite of metadata defined by the `bioio` package [here](https://bioio-devs.github.io/bioio/generated/bioio.StandardMetadata.html#). The modal extracts the file type (.lif,.ome.zarr, etc..) so that it can recommend necessary `bioio` pluigins in setup. Then generates a script that extracts the metadata from each file and inserts it into a csv. 

### Screenshot of Inital Selection 
Screenshot 2025-09-11 at 10 14 26 AM

### Screenshot of Generated Snippet
Screenshot 2025-09-11 at 10 14 37 AM

### Example of Generated CSV
[selected_files_metadata.csv](https://github.com/user-attachments/files/22280515/selected_files_metadata.csv)

### DIFF
--- a/packages/core/components/DataSourcePrompt/index.tsx
+++ b/packages/core/components/DataSourcePrompt/index.tsx
@@ -59,6 +59,7 @@ export default function DataSourcePrompt(props: Props) {
                 selection.actions.addQuery({
                     name: `New ${dataSource.name} Query`,
                     parts: { sources: [dataSource], sourceMetadata: metadataSource },
+                    loading: true,
                 })
             );
         }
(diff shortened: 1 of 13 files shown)


aswallace (commented):
Looks good to me with minor nits

PR #609, opened 2025-11-05
Add Vol-E as .tiff viewer by default

SeanDuHare:
Vol-E can load "small enough" tiff files. This adds Vol-E as a viewer for .tiff files. In the future, it would be great to determine the max size and remove it as an option when too big.

### DIFF
--- a/packages/core/hooks/useOpenWithMenuItems/index.tsx
+++ b/packages/core/hooks/useOpenWithMenuItems/index.tsx
@@ -12,6 +12,8 @@ import styles from "./useOpenWithMenuItems.module.css";
 import useOpenInCfe from "./useOpenInCfe";
 import useRemoteFileUpload from "../useRemoteFileUpload";
 
+const ONE_MEGABYTE = 1024 * 1024;
+
 enum AppKeys {
     AGAVE = "agave",
     BROWSER = "browser",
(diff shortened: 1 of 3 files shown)


SeanLeRoy:
> Side note: I don't think agave, cfe, or vole know what to do with a .dvi file. (?) Also I don't think cfe knows what to do if it is passed a tiff file.
> 
> But on the main point, I sort of remember talking this over with UX and it was really based on the idea that microscopy tiffs can be very large, and VolE can't load a tiff in chunks, it has to pull the whole thing down.

@ShrimpCryptid Added cfe for .tiff and .dvi [here](https://github.com/AllenInstitute/biofile-finder/commit/bbf15f73f9eb8604ae7d013422550bd5b37eab31) I removed it in this [8500489](https://github.com/AllenInstitute/biofile-finder/pull/609/commits/8500489742740094ef545893d703de99a5944d51)

I added .dvi accidentally a while back from a neuroglancer change - I changed it up

SeanLeRoy:
@toloudis I was avoiding doing the work of setting a maximum file size for using Vol-E for the moment - but went ahead and just did it in https://github.com/AllenInstitute/biofile-finder/pull/609/commits/3aeaeb764a59e0eacbad8d4f19381496c69ed76c just in case

PR #611, opened 2025-11-06
Add a target zarr resolution to gallery view

aswallace:
## Context
We now use `ome-zarr.js` to generate zarr thumbnails, which allows us to specify a target resolution for the images. 
For the gallery view, the current target is `undefined`, which means that the lowest possible resolution is used. This has been helpful for keeping our rendering effort low, but also means that for datasets like [this one](https://bff.allencell.org/app?c=File+Path%3A0.25%2CFile+Name%3A0.25%2Ctype%3A0.25%2Cgrid%3A0.25&v=3&source=%7B%22name%22%3A%22pub+%285%2F5%2F2025+10%3A18%3A08+AM%29%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22https%3A%2F%2Fdocs.google.com%2Fspreadsheets%2Fd%2Fe%2F2PACX-1vRiUH2WgB_jHDKys9uvRz-dOimbIe4WsvS_doVE-LNTlcvS1zogDL2D8JlXE_5jt7rBk26vpjXff6wg%2Fpub%3Fgid%3D200533885%26single%3Dtrue%26output%3Dcsv%22%7D) (provided by @krokicki), thumbnails can still be inconsistent in the gallery view (e.g., some are blank or fuzzy, and others are high resolution).

## Changes
This adds a target size for the gallery view. It's using a semi-regular value (the number of columns in the grid rather than the calculated screen size) so that we can avoid having extra network calls to re-render thumbnails on screen changes. 

Since I'm less familiar with the library, I'd like to check with @will-moore to verify that there are no concerns with specifying a resolution that may not exist on the file-- namely, does it default to a lower resolution if the requested size is too high? + does it default to the lowest possible resolution if the requested one is smaller than the smallest available? Also, I didn't notice a big change in performance, but it's worth keeping an eye on. 

## Testing
Tested with the above dataset, and made sure that our other datasets still generate appropriate thumbnails. 

#### Before
image

#### After
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -95,26 +95,6 @@
         "npm": "10.x"
       }
     },
-    "node_modules/@75lb/deep-merge": {
-      "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/@75lb/deep-merge/-/deep-merge-1.1.1.tgz",
-      "integrity": "sha512-xvgv6pkMGBA6GwdyJbNAnDmfAIR/DfWhrj9jgWh3TY7gRm3KO46x/GPjRg6wJ0nOepwqrNxFfojebh0Df4h4Tw==",
-      "dependencies": {
-        "lodash.assignwith": "^4.2.0",
-        "typical": "^7.1.1"
-      },
-      "engines": {
-        "node": ">=12.17"
-      }
-    },
-    "node_modules/@75lb/deep-merge/node_modules/typical": {
-      "version": "7.1.1",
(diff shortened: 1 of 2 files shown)


will-moore:
Yes, the dataset resolution will be chosen to be as close as possible to the targetSize. It'll choose the smallest if the target is smaller.
The logic for that is at https://github.com/BioNGFF/ome-zarr.js/blob/b18174666069306ab8bb859171574e29a4b41b1a/src/render.ts#L56

PR #566, opened 2025-08-12
feat: Open selections in CFE

ShrimpCryptid:
## Context
Adds support for opening files in CFE! This change makes the first change required for https://github.com/allen-cell-animated/cell-feature-explorer/issues/226.

## Changes
- Adds a new `getManifest()` API for the `FileService` interface, which allows callers to get a CSV file without downloading (like in `download()`).
- Adds a hook for uploading files to an external server, currently set up on Dan's dev server (`useRemoteFileUpload`). 
- Adds a menu option for opening a selection of OME-Zarr and OME-TIFF files in Cell Feature Explorer.
  - The menu option is hidden if the upload endpoint cannot be reached.
  - The selection is saved as a CSV, which is uploaded to the endpoint. A UUID file ID is returned and can be used to access the file in CFE. 

## Testing
Tested running locally and opening selections of OME-Zarr files in CFE. See attached video!

**Video talk-through (🔊):**

https://github.com/user-attachments/assets/7e325796-f766-4a1d-b61c-2d4b2cceae88

PR #621, opened 2025-11-28
Open file selections and metadata in Vol-E v2

frasercl:
## Context

For a while, we've wanted a way to send more data to an opened Vol-E window than could practically fit in URL query parameters &mdash; namely, large selections of images and image metadata. Previous passes at this include #516, which got caught up in arcane browser security policies; and #566, which requires an intermediate server, and targets CFE rather than Vol-E. The implementation in this PR happens entirely locally, appears unlikely to run afoul of any cross-origin isolation, and has the convenient bonus of being backwards-compatible with versions of Vol-E that aren't expecting the extra protocol.

## Changes

Here's how it works:
- The new `openInVole` handler always includes *something* in the `url` query param of the Vol-E URL it opens, which ensures Vol-E will always show *something* even if one of the steps below fails.
- ...but it also adds new query params: `storageid` uniquely identifies the message BFF will send to Vol-E, and `msgorigin` is the origin of the opening BFF window (the slice of the URL from the beginning through the port, i.e. https://bff.allencell.org), which helps with security.
- On load, Vol-E recognizes that those params are present and uses `postMessage` to send the value of `storageid` back to BFF, indicating that it's ready to receive the message content.
- Once BFF receives that message back from Vol-E, it `postMessage`s back the extra data.

allen-cell-animated/vole-app#467 is the counterpart PR on the Vol-E side. It gets into the details of how Vol-E handles the received message.

## Testing

With BFF on this branch...
- Make a large selection of image files and choose "Open in Vol-E." **Expectation:** only the "focused" image should open in Vol-E. `msgorigin` and `storageid` will be in the URL but unused.
- Select a small number of image files (~3) and choose "Open in Vol-E". **Expectation:** this branch is willing to put small multi-scene collections in the URL, so all these images should open at once (the "Scene" slider will be visible in Vol-E's "Clipping" panel). `msgorigin` and `storageid` should still be there to send image metadata.
- Change `VOLE_BASE_URL` to http://localhost:9020/viewer, and start Vol-E locally on the branch `feature/receive-from-bff`. Make a large selection of image files and choose "Open in Vol-E." **Expectation:** subject to the caveats described in allen-cell-animated/vole-app#467, Vol-E will open the whole image collection, starting on the image that was "focused" in BFF, and data from other columns in BFF will appear in Vol-E's "Metadata" panel.


packages/core/hooks/useOpenWithMenuItems/index.tsx
```diff
@@ -299,6 +347,65 @@ export default (fileDetails?: FileDetail, filters?: FileFilter[]): IContextualMe
         fileService
     );
 
+    // custom hook this, like `useOpenInCfe`?
+    const openInVole = React.useCallback(async (): Promise<void> => {
+        const allDetails = await fileSelection.fetchAllDetails();
+        const details = allDetails.filter((detail) => {
+            const fileExt = getFileExtension(detail);
+            return fileExt === "zarr" || fileExt === "";
```
SeanDuHare: Sldy files are sometimes captured in the community as `.sldy/` which would become `""` here

packages/core/hooks/useOpenWithMenuItems/index.tsx
```diff
@@ -267,6 +276,44 @@ function getFileExtension(fileDetails: FileDetail): string {
...
+ *    validation, and `storageid` uniquely identifies the message we want to send.
+ * 2. *The opened window must check if these params are present* and post the value of `storageid`
+ *    back to us (via `window.opener`, validated using `msgorigin`) once it's loaded and ready.
+ * 3. Once we receive that message, we send over `entry`.
+ *
+ * This is currently only used by `openInVole`, but is broken out into a separate function to
+ * emphasize that this protocol is both message- and receiver-agnostic, and could be used to send
+ * large bundles of data to other apps as well.
+ */
+function openWindowWithMessage(openUrl: URL, message: any): void {
```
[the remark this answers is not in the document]
frasercl: That `any` is there to match the signature for `postMessage`, which can also take `any` type. Certainly no reason we can't make our own function more specific though.

packages/core/hooks/useOpenWithMenuItems/index.tsx
frasercl: `getFileExtension` is unaware of `/` and would get this right.

...which implies a bigger problem: this will fall over for valid Zarr images without an extension with a `.` somewhere else in their file path: `.../john.doe/valid_zarr`. I'll look at making that more robust, including special-casing `.sldy/`.

PR #629, opened 2026-01-03
Bugfix/Query duplication logic

### LINKED ISSUE #626: Duplicate query button does not include filters
## Description

When users click the "Duplicate query" button, a new query is created that uses the same source, but does not include the applied filters/groups/sorts.

## Expected Behavior

The query should be duplicated exactly, including any applied filters etc.

## Reproduction

Start querying a data source, and apply any set of filters and groups. Then click "Duplicate query" in the footer of the subpanel. 

## Environment
- [x] Web
- [x] Desktop
   - Version: [e.g. 8.1.1]

aswallace:
## Context
Resolves #626 
Currently, when users attempt to duplicate existing queries by using the button in the query footer, the new query that is created doesn't always end up with all of the same filters/groups/properties as the original. 

Also, the logic for adding new queries sometimes fails to generate unique names, which can lead to odd behavior. This is because the current logic just tacks on an incremented count of the total occurrences for a given query name, which doesn't account for the user deleting or renaming queries.

To replicate:
- Create query named "Test query"
- Duplicate that query twice, so that there's a query called "Test query (1)" and one called "Test query (2)"
- Delete "Test query (1)"
- Duplicate the original query
- There will now be two queries called "Test query (2)". Now, when you attempt to switch between them or make changes to either of the duplicates, both will be affected

## Changes
- Includes ALL original query parts in the duplication logic
- Changes the query naming logic to find the first available unused number instead of incrementing a total count

## Testing
Added a few unit tests for the naming logic, and otherwise tested manually

### DIFF
--- a/packages/core/state/selection/test/logics.test.ts
+++ b/packages/core/state/selection/test/logics.test.ts
@@ -10,6 +10,7 @@ import { get as _get, shuffle } from "lodash";
 import sinon from "sinon";
 
 import {
+    addDataSourceReloadError,
     addFileFilter,
     addQuery,
     changeDataSources,
@@ -23,6 +24,7 @@ import {
     selectFile,
     selectNearbyFile,
     ADD_DATASOURCE_RELOAD_ERROR,
+    ADD_QUERY,
     SET_ANNOTATION_HIERARCHY,
     SET_AVAILABLE_ANNOTATIONS,
     SET_FILE_FILTERS,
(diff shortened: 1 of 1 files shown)


kmitcham:
Nit: is it better to do the 'first available number' or 'add one to the highest existing number?'  The latter seems more intuitive to me.  I guess it would make the code more complicated to try and detect the highest copy number, since who knows if they made 100 duplicates and deleted 99.

kmitcham (approved):

PR #633, opened 2026-01-08
ACM REVIEW #0(Base): Feature/all cells mask segmentation

### LINKED ISSUE #702: Feature Request: Bff to Compute - ACM1 pipeline
## Description 

We want to add a way for users to select files and run the ACM pipeline on slurm.

BrianWhitneyAI:
## Context
The purpose of this PR is to create a POC for AICS compute pipelines. It resolves #702   It is NOT intended to be merged. The goal of this design is to deliver something that people can try out so that we can verify interest in this project and identify more requirements from the user’s side. The proof of concept is limited to running “All Cells Mask” on central IT’s Slurm cluster.

### Demo 
Heres a demo of the feature. The video was too large to embed
https://drive.google.com/file/d/1a2FQpDDN_wR4q_2TuYaZ4SP9AxZOecW3/view?usp=sharing 

### Development notes.
1) The waiting states for the CSV are very fragile and hacky they should definitely not be merged, they are included to smooth over the POC. 
2) This pipeline currently requires files to have local vast paths, I would recommend removing that complexity if we decide to move this into a real feature
3) theres no  process failure error support currently and minimal response error handling 

### Build 
This feature is Desktop only so heres the build if you want to try it out! 
https://github.com/AllenInstitute/biofile-finder/actions/runs/20826578098


packages/core/state/interaction/logics.ts
```diff
@@ -864,6 +866,64 @@ const copyFilesLogic = createLogic({
...
+            });
+
+            dispatch(
+                interaction.actions.processInfo(
+                    processId,
+                    `Waiting for CSV to be written:\n${manifestCsvPath}`
+                )
+            );
+
+            await httpFileService.waitForPath(manifestCsvPath);
```
SeanDuHare: This method has a "this is bad we should never do it" comment on it

packages/desktop/src/services/DatabaseServiceElectron.ts
```diff
@@ -12,6 +12,7 @@ export default class DatabaseServiceElectron extends DatabaseService {
     constructor() {
         super();
         this.database = new duckdb.Database(":memory:");
+        this.database.exec("INSTALL 'httpfs';");
```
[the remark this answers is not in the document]
BrianWhitneyAI: OK this was the bug that Anya and I were running into. I couldn't add data sources without this installed. super weird. Hopefully her pr on main will completely remove this.

packages/core/state/interaction/logics.ts
BrianWhitneyAI: yeah. This is what I was referencing in the dev notes. We ping to wait for the csv to exist but checking a non-existant file will look like a CORS block since it doesn't divulge if it exists or not. We 100% should not do this specific logic and have some other tracking of job progress but without it BFF cannot ingest the empty csv path

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -221,4 +223,76 @@ export default class HttpFileService extends HttpServiceBase implements FileServ
...
+
+    // This is bad and we should never do it
+    public async waitForPath(
+        url: string,
+        options?: {
+            intervalMs?: number;
+            timeoutMs?: number;
+        }
+    ): Promise<void> {
+        const intervalMs = options?.intervalMs ?? 3000;
```
saeliddp: Did you consider using a backoff algorithm?

BrianWhitneyAI: No, Didnt put a whole lot of thought into this given that I dont think this request method is something we actually want to do, better to either initially create the csv or subscribe to jss

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+        setStatusMessage(null);
+
+        try {
+            const details = await fileSelection.fetchAllDetails();
+            if (!details.length) {
+                throw new Error("No files selected.");
+            }
+
+            const files: string[] = details.map((d) => {
+                // TODO: ideally switch to file IDs instead of local paths
```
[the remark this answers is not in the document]
saeliddp: > There is some added complexity using local paths

How complex are we talking? Reading from VAST paths is faster than downloading images from S3, so moving to cloud paths would probably be a significant sacrifice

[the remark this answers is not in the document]
kmitcham: Possibly a resolver that goes one way for a path value and talks to FSS (and gets the location, local or S3) for an id? (assume anything the right length and without a path separator is an id?)

packages/core/hooks/useProcessFilesMenuItems.ts
```diff
@@ -40,5 +40,18 @@ export default function useProcessMenuItems(
                 );
             },
         },
+        {
```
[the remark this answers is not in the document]
BrianWhitneyAI: my mistake It does work on web! Now the question Do we want it to?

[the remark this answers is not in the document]
SeanDuHare: I think having it be web is fine we do **not** want it enabled for non-aics users (there is a state property for that called something like isAicsEmployee)

packages/core/state/interaction/logics.ts
SeanDuHare: Perhaps create a ticket requesting to replace this with something better then

BrianWhitneyAI: Would we want to merge this with the kind of jankey status checking? I would think that we might want to stay on branch until another option is worked out.

packages/desktop/src/services/DatabaseServiceElectron.ts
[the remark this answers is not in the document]
BrianWhitneyAI: Have not verified. I will have to do a rebase for this branch, will do this after we have a project discussion with Jennifer.

saeliddp (approved):

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
kmitcham, line 1: Is the idea to get this version working, then abstract it, or just go with dedicated components for each workflow?

kmitcham (approved):

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: We are headed towards pipeline discovery. so BFF will ask for available pipelines then ingest those into modals for each one. Unsure how specific each one will be right now.

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+import { useDispatch, useSelector } from "react-redux";
+
+import { ModalProps } from "..";
+import BaseModal from "../BaseModal";
+import { PrimaryButton, SecondaryButton } from "../../Buttons";
+import FileSelection from "../../../entity/FileSelection";
+import { interaction, selection } from "../../../state";
+
+import styles from "./AllCellsMaskSegmentation.module.css";
+import AnnotationName from "../../../entity/Annotation/AnnotationName";
```
SeanDuHare: This import should be moved up

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+type Status = "idle" | "submitting" | "submitted" | "error";
+
+type UIOpts = {
+    channelIndex: string;
+};
+
+export default function AllCellsMaskSegmentation({ onDismiss }: ModalProps) {
```
SeanDuHare: A description of this component would be good

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+                const shouldBeInLocal = d.getFirstAnnotationValue(
+                    AnnotationName.SHOULD_BE_IN_LOCAL
+                );
+
+                const localPath = d.getFirstAnnotationValue(AnnotationName.LOCAL_FILE_PATH);
+
+                return !shouldBeInLocal || typeof localPath !== "string";
+            });
+
+            setHasNonLocalFiles(hasAnyNonLocal);
```
SeanDuHare: This has to do two `fetchAllDetails` calls to grab local paths and in doing so has to check both times for if the path is defined for each file. Rather than doing that perhaps you could just change this setState to `setLocalFilePaths` **only** after making sure no files are missing a path ex.
```
const fileDetails = await fileSelection.fetchAllDetails();
 if (cancelled) return;

const localPaths = fileDetails
    .filter(file => file.getFirstAnnotationValue(AnnotationName.SHOULD_BE_IN_LOCAL) && !!file.getFirstAnnotationValue(AnnotationName.LOCAL_FILE_PATH)
    .map(file => file.getFirstAnnotationValue(AnnotationName.LOCAL_FILE_PATH));
if (localPaths.length === fileDetails.path) {
    setLocalFilePaths(localPaths);
}
```

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+            if (cancelled) return;
+
+            const hasAnyNonLocal = details.some((d) => {
+                const shouldBeInLocal = d.getFirstAnnotationValue(
+                    AnnotationName.SHOULD_BE_IN_LOCAL
+                );
+
+                const localPath = d.getFirstAnnotationValue(AnnotationName.LOCAL_FILE_PATH);
+
+                return !shouldBeInLocal || typeof localPath !== "string";
```
SeanDuHare: Would `localPath` ever not be a string? Could this just check if it is defined?

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+                return localPath;
+            });
+
+            const scene = 0;
+
+            const channel =
+                opts.channelIndex.trim() && /^\d+$/.test(opts.channelIndex)
+                    ? parseInt(opts.channelIndex, 10)
+                    : 0;
```
SeanDuHare: Hmm this should probably error out if you can parse an int when _something_ was supplied rather than defaulting to 0.

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+                files,
+                scene,
+                channel,
+            });
+
+            setComputeTaskId(computeTaskId ?? null);
+            setDashboardUrl(dashboardUrl ?? null);
+            setStatus("submitted");
+
+            dispatch(
```
SeanDuHare: Does this happen fast enough that we shouldn't do a `processStarted` action at the beginning of this call?

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+                            <h4>Submitting job</h4>
+                            <p className={styles.text}>Your request is being submitted.</p>
+                        </>
+                    )}
+
+                    {status === "submitted" && computeTaskId && (
+                        <>
+                            <h4>Job submitted</h4>
+                            <p className={styles.text}>
+                                <strong>Job ID:</strong> {computeTaskId}
```
SeanDuHare: Should this be a link to the JSS dashboard?

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -221,4 +223,35 @@ export default class HttpFileService extends HttpServiceBase implements FileServ
...
+                })
+            );
+
+            return {
+                computeTaskId: result.computeTaskId,
+                dashboardUrl: result.dashboardUrl,
+            };
+        } catch (error) {
+            console.error("Failed to submit All Cells Mask job:", error);
+            throw error;
```
SeanDuHare: Are there any errors we are aware of that we could catch and output as something actionable to the user? For instance, can a 422 happen?

packages/core/state/interaction/logics.ts
```diff
@@ -940,6 +942,55 @@ const copyFilesLogic = createLogic({
...
+
+        const httpFileService = interactionSelectors.getHttpFileService(getState());
+        const processId = uniqueId("all-cells-mask-");
+
+        try {
+            dispatch(
+                interaction.actions.processStart(processId, "Submitting All Cells Mask job...")
+            );
+
+            const { computeTaskId, dashboardUrl } = await httpFileService.submitAllCellsMaskJob({
```
SeanDuHare: This logic isn't used nor are the actions

SeanDuHare (changes requested):

packages/core/services/FileService/HttpFileService/index.ts
BrianWhitneyAI: rawPost should already handle this since the error will surface with its message.

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: THe button is a link to the dashboard. the id is something they could query the dashboard landing page for if they loose the link

packages/core/state/interaction/logics.ts
BrianWhitneyAI: ah! whoops a relic of this being open for so long. Will remove!

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: no we should do a processStart will add

packages/core/state/interaction/logics.ts
BrianWhitneyAI: [4dc1384](https://github.com/AllenInstitute/biofile-finder/pull/633/commits/4dc13844f82787b760cd5b83eacac2c0563c1faf)

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: [4dc1384](https://github.com/AllenInstitute/biofile-finder/pull/633/commits/4dc13844f82787b760cd5b83eacac2c0563c1faf)

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: [4dc1384](https://github.com/AllenInstitute/biofile-finder/pull/633/commits/4dc13844f82787b760cd5b83eacac2c0563c1faf)

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: changed this around a bit in [4dc1384](https://github.com/AllenInstitute/biofile-finder/pull/633/commits/4dc13844f82787b760cd5b83eacac2c0563c1faf)

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: [4dc1384](https://github.com/AllenInstitute/biofile-finder/pull/633/commits/4dc13844f82787b760cd5b83eacac2c0563c1faf)

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: [4dc1384](https://github.com/AllenInstitute/biofile-finder/pull/633/commits/4dc13844f82787b760cd5b83eacac2c0563c1faf)

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: [4dc1384](https://github.com/AllenInstitute/biofile-finder/pull/633/commits/4dc13844f82787b760cd5b83eacac2c0563c1faf)

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -0,0 +1,259 @@
...
+                    "Failed to submit All Cells Mask job."
+                )
+            );
+        }
+    };
+
+    const submitDisabled = status !== "idle" || hasNonLocalFiles;
+
+    const submitTooltip = hasNonLocalFiles
+        ? "All selected files must be available locally before submitting this job."
```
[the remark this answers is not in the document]
BrianWhitneyAI: [1f87232](https://github.com/AllenInstitute/biofile-finder/pull/633/commits/1f87232a60e8e624d1574e7bcda8c6cd219e2480)

packages/core/services/FileService/HttpFileService/index.ts
SeanDuHare: Yea I was thinking it might be useful to reshape some potentially if there are messages are that server side wording specific

SeanDuHare (approved):

PR #643, opened 2026-01-16
Remove call to fetch total count from query footer

### LINKED ISSUE #636: Remove duplicate `SELECT COUNT(*) ...` query
Significance: filtering large tables (e.g., 30M rows) can take 30s+ (measured on the `demo-large-parquet` branch), and a big chunk of that time is due to running the query to count the number of resulting files twice. For more context, see #620.

Steps to reproduce:
1. Edit BFF source code to enable logging of each duckdb query. For example, I used [these 4 lines](https://github.com/AllenInstitute/biofile-finder/compare/demo-large-parquet?expand=1#diff-4c1d6c7d6d0485bb57292cbb053231f90ff1432fdf1fed28d1406a7b82c07efcR63-R66)
1. With any data source, click "Filter +"
2. Select any column
3. Open the dev console and clear the log output
4. Select any value from that column
5. Check the dev console for `SELECT COUNT(*) ...` queries

Actual result: two identical `SELECT COUNT(*) ...` queries are run.

Expected result: the query is just triggered once.

aswallace:
## Context
Resolves #636: Every time the user updates filters on a data source, an extra `SELECT COUNT(*)` query runs in the background, which can be slow and expensive.

The query panel footer includes a button that allows users to save the entire query result as a csv/json/parquet file. 
When we initially implemented this, there were some concerns about whether we should allow folks to download results for a huge number of files (e.g., hundreds of thousands +). To address this, we placed a somewhat arbitrary limit on the total file count (250,000) and disabled the button when the file count exceeded this number.

The problem is that this total file count has to happen separately from the query that populates the file count at the bottom of file lists. When there are groups applied, we don't actually know the file count for each group until we open a group "folder." 

It seems like the best solution in this case is to fully remove the check/restriction from the query footer: 
1. We don't actually have evidence yet of issues with users needing to save CSVs for huge numbers of files (aside from the note below re FES)
2. Even if we did, the restriction doesn't have to (and probably shouldn't) happen that early on in the UI. We could move it to the manifest download modal instead, so that the count only occurs when the user actually intends to download the manifest instead of on every query update.

## Changes
For now, fully removes `fetchTotalCount` from the query footer, and enables manifest downloads for any number of files. 
If we want to restrict this in the future, we can consider implementing (2) above

## Testing
Tested manually:
 - Not seeing the extra `SELECT COUNT(*)` query anymore
 - Can still download manifests (for <12,000 files... see below)

## Note
Found an incidental bug in FES where the manifest query cuts off the results for >12,000 files. There's nothing for us to do about this on the BFF side right now, but depending on the source of the issue, we may need to consider restricting manifest download sizes anyway

### DIFF
--- a/packages/core/components/QuerySidebar/QueryFooter.tsx
+++ b/packages/core/components/QuerySidebar/QueryFooter.tsx
@@ -5,18 +5,14 @@ import { useDispatch, useSelector } from "react-redux";
 
 import { TertiaryButton } from "../Buttons";
 import { ModalType } from "../Modal";
-import Tooltip from "../Tooltip";
 import Tutorial from "../../entity/Tutorial";
 import IncludeFilter from "../../entity/FileFilter/IncludeFilter";
-import FileSet from "../../entity/FileSet";
 import { SearchParamsComponents } from "../../entity/SearchParams";
 import useSaveMetadataOptions from "../../hooks/useSaveMetadataOptions";
 import { interaction, selection } from "../../state";
 import { Query } from "../../state/selection/actions";
 
 import styles from "./QueryFooter.module.css";
-
-const MAX_MANIFEST_FILE_COUNT = 250000;
 
 interface Props {
     isDeletable?: boolean;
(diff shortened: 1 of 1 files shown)


kmitcham (approved):
Deleted code remains the best code.

PR #664, opened 2026-02-04
Separate download logic from S3 handling

SeanDuHare:
## Context
We have regularly seen different behavior across the app when trying to interact with S3 due to separated efforts and logic handling.

## Changes
This attempts to remove as much S3 specific logic from the download service as possible - no new functionality or bugs were intentionally fixed, added, or changed as part of this specific change set.

### DIFF
--- a/packages/core/components/FileList/Header.module.css
+++ b/packages/core/components/FileList/Header.module.css
@@ -23,13 +23,10 @@
   background-color: var(--primary-background-color);
   border-radius: var(--small-border-radius);
   box-shadow: var(--box-shadow);
-  color: var(--primary-text-color);
-  padding: var(--inner-header-padding) 0;
 }
 
 .header:hover {
   background-color: var(--accent-dark);
-  color: var(--primary-text-color);
 }
 
 .header > div {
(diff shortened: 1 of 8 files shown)


packages/core/services/S3StorageService/index.ts
```diff
@@ -0,0 +1,216 @@
+import { parseS3Url, isS3Url } from "amazon-s3-url";
+import axios from "axios";
+
+import HttpServiceBase, { ConnectionConfig } from "../HttpServiceBase";
+
+/**
+ * Return true if the URL seems to point to a multi object file like Zarr
+ */
+export const isMultiObjectFile = (url: string) =>
+    [".zarr", ".zarr/", ".sldy", ".sldy/"].some((ext) => url.endsWith(ext));
```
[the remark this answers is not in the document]
aswallace: Ah okay, I was for some reason remembering there being additional path after that (e.g. .zarr/0/a/b/c/)

packages/web/src/services/FileDownloadServiceWeb.ts
```diff
@@ -1,55 +1,51 @@
...
+import { isMultiObjectFile } from "../../../core/services/S3StorageService";
 
 export default class FileDownloadServiceWeb extends FileDownloadService {
     isFileSystemAccessible = false;
-    private readonly activeRequestMap: ActiveRequestMap = {};
+
+    // TODO: Test this in unit tests
+    public static isLocalPath(filePath: string): boolean {
+        const uriPattern = /^(https?|ftp):\/\/|^[a-zA-Z]:\\/;
+        return filePath.startsWith("file://") || !uriPattern.test(filePath);
```
kmitcham: would this return true for s3://some.bucket.file ?  Is that even a use case?

packages/core/services/S3StorageService/index.ts
```diff
@@ -0,0 +1,216 @@
...
+            parsedHost = hostname.slice(bucket.length + 1);
+        }
+        return { hostname: parsedHost, bucket, key };
+    }
+
+    constructor(config: ConnectionConfig = {}) {
+        super({ ...config, includeCustomHeaders: false });
+    }
+
+    public async formatAsHttpResource(url: string): Promise<string | undefined> {
```
aswallace: the second half of this appears to duplicate the private `formatAsHttpResource(parsedUrl: ParsedUrl)`? comment explaining difference might be helpful

PR #670, opened 2026-02-10
bugfix/dataset-creation-deregister

BrianWhitneyAI:
## Description

Updates the name variables used in table creation to fix a bug being seen on the datasets page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,8 +45,7 @@
         "redux": "4.0.x",
         "redux-logic": "3.x",
         "reselect": "4.0.x",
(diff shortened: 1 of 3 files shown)


pgarrison (approved):
This bug was introduced in #666 where I updated the logic above to register files under a different name from the data source name.

The scope of this bug is much larger: I expect all CSV and JSON reading to be broken.

I tested this fix locally by opening the Open Source Datasets page and seeing that multiple datasets are listed.

kmitcham (approved):

PR #671, opened 2026-02-10
Register input files under the same name as their data source

pgarrison:
## Context
In #666 I updated BFF to register input files to DuckDB under different file names than their corresponding table/view. That introduced a major bug that broke CSV and JSON inputs. #670 resolved the bug. But I noticed that it actually is not necessary for any of the BFF features to use different names for the registered file and the table/view.

## Changes
Remove the distinction between view/table name and input file's registered name.

Pro: Simpler code.

Con: Maybe it improves debugging to keep these logically separated. 

## Testing
Manual tests conducted (most of these copied from #666):
- Open the "Enhancer AAVs" Open Source Dataset
- Open the 2 BFF links going into the EMT paper
- Open parquet table with parquet metadata descriptor file and confirm that column descriptions are displayed
- Set a group by and a sort, both on columns with many repeated values, then use the arrow keys to navigate down one of the resulting file lists: the order of the displayed files should not change.
- Set a filter, then save to CSV: check that the correct files are downloaded.
- Open a [parquet from S3](http://localhost:9001/app?c=File+Name%3A0.25%2Cabundance_stoichiometry%3A0.25%2CCell+Line%3A0.25%2CColony+Position%3A0.25&source=%7B%22name%22%3A%22ibid.parquet+%282%2F9%2F2026+12%3A39%3A37+PM%29.parquet%22%2C%22type%22%3A%22parquet%22%2C%22uri%22%3A%22https%3A%2F%2Fstaging-biofile-finder-datasets.s3.us-west-2.amazonaws.com%2Fibid.parquet%22%2C%22mode%22%3A%22DIRECT_FROM_PARQUET%22%7D), then copy-paste the URL into a new tab
- Open a parquet table with FMS-formatted column names like `file_id`


packages/core/services/DatabaseService/index.ts
```diff
@@ -543,11 +536,11 @@ export default class DatabaseService {
         });
         selectParts.push(`"file_row_number" AS "${DatabaseService.HIDDEN_UID_ANNOTATION}"`);
         // 4. Create the view for this data source
-        const createViewSql = `CREATE VIEW "${viewName}"
+        const createViewSql = `CREATE VIEW "${name}"
```
BrianWhitneyAI: Is there a potential name convention overlap here having a view be the same name as the table?

aswallace (approved):

PR #675, opened 2026-02-13
Bugfix/generate-parquet-file-name-from-path

BrianWhitneyAI:
## Description 

The purpose of this PR is to resolve a bug where when reading parquet files without a "File Name" selecting the files would crash. Previously we generated this column in its absence off of the path URl. 

### Description of changes

- This PR reintroduces the column generation and does some basic filtering of the values.
- Adds test for parquet missing column cases

### DIFF
--- a/packages/core/hooks/useOpenWithMenuItems/index.tsx
+++ b/packages/core/hooks/useOpenWithMenuItems/index.tsx
@@ -17,6 +17,7 @@ const ONE_MEGABYTE = 1024 * 1024;
 enum AppKeys {
     AGAVE = "agave",
     BROWSER = "browser",
+    FIJI = "fiji",
     NEUROGLANCER = "neuroglancer",
     SIMULARIUM = "simularium",
     VALIDATOR = "validator",
@@ -28,6 +29,7 @@ enum AppKeys {
 interface Apps {
     [AppKeys.AGAVE]: IContextualMenuItem;
     [AppKeys.BROWSER]: IContextualMenuItem;
+    [AppKeys.FIJI]: IContextualMenuItem;
     [AppKeys.NEUROGLANCER]: IContextualMenuItem;
     [AppKeys.SIMULARIUM]: IContextualMenuItem;
     [AppKeys.VALIDATOR]: IContextualMenuItem;
@@ -39,6 +41,7 @@ interface Apps {
 type AppOptions = {
     openInCfe: () => void;
     openInVolE: () => void;
+    openInFiji: () => void;
 };
 
 const SUPPORTED_APPS_HEADER = {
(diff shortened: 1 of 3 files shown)


packages/core/services/DatabaseService/test/DatabaseService.test.ts
```diff
@@ -165,4 +165,45 @@ describe("DatabaseService", () => {
...
+            ]);
+
+            // Act
+            const result = getParquetFileNameSelectPart(actualToPreDefined);
+
+            // Assert
+            expect(result).to.be.null;
+        });
+
+        it("returns null when File Path column does not exist", () => {
```
SeanDuHare: I suppose this makes sense to include as a test, but we have bigger issues if there is no File Path

packages/core/services/DatabaseService/index.ts
```diff
@@ -49,6 +49,38 @@ function getActualToPreDefinedColumnMap(columns: string[]): Map<string, string>
     return map;
 }
 
+/**
+ * Derive a "File Name" from a path-like column (local path or URL).
+ */
+export function getFileNameFromPathExpression(quotedPathColumn: string): string {
```
SeanDuHare: No need to export it seems
```suggestion
function getFileNameFromPathExpression(quotedPathColumn: string): string {
```

packages/core/services/DatabaseService/test/DatabaseService.test.ts
BrianWhitneyAI: I think these are part of Philips TDD, we can remove em if you think they arnt relevant.

packages/core/services/DatabaseService/index.ts
```diff
@@ -49,6 +49,38 @@ function getActualToPreDefinedColumnMap(columns: string[]): Map<string, string>
...
+    const cleaned = `REGEXP_REPLACE(
+        REGEXP_REPLACE(${quotedPathColumn}, '[?#].*$', ''),
+        '/+$',
+        ''
+    )`;
+    const basename = `REGEXP_EXTRACT(${cleaned}, '([^/]+)$', 1)`;
+    const stripOme = `REGEXP_REPLACE(${basename}, '(?i)\\\\.ome$', '')`;
+
+    return `COALESCE(NULLIF(${stripOme}, ''), ${quotedPathColumn})`;
```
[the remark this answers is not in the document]
BrianWhitneyAI: I tried to split it up a bit more than our previous "One go" regex. Again I think we can probably iterate here its not optimal

packages/core/services/DatabaseService/index.ts
```diff
@@ -49,6 +49,38 @@ function getActualToPreDefinedColumnMap(columns: string[]): Map<string, string>
...
+/**
+ * For parquets: add a computed "File Name" when we have "File Path" but not "File Name".
+ */
+export function getParquetFileNameSelectPart(
+    actualToPreDefined: Map<string, string>
+): string | null {
+    const hasFileName = [...actualToPreDefined.values()].includes(PreDefinedColumn.FILE_NAME);
+    if (hasFileName) return null;
+
+    const pathColumn = [...actualToPreDefined.entries()].find(
```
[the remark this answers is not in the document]
BrianWhitneyAI: I dont have too much of a preference but I think findKey is more readable so im fine with it

SeanDuHare:
Need this fix for paper reviewers so merging in with just Brian & I for now - will make sure others see this later and likely iterate on some of the decisions made here

PR #687, opened 2026-03-06
Feature/stream zarr downloads to zip

SeanDuHare:
## Context
This PR started as an attempt to fix an issue Jianxu's lab had where we were unable to download his Zarrs since they weren't providing an S3 interface. While adding the capability to address that I discovered a way to stream our Zarr downloads (or any file download) meaning we could download very large Zarrs, something we currently can't do.

## Changes
This does the following major changes:
1) Adds an entity responsible for iterating over a Zarr. First it will attempt to read the S3 object dir listing (which is improved now to handle pagination), and if it can't read the object dir for whatever reason it will attempt to iterate over the Zarr using the coordinate systems laid out within the metadata files Zarrs have (ex. zarr.json).
2) Downloads are now streamed to zip rather than bundling them in memory. This happens by fetching chunks of each file and piping the response directly into the zip effectively avoiding the browser's memory. 

### Where to start reviewing???
The new entity files (`Zarr`, `ConcurrentTaskQueue`, & `StreamedZipDownloader`) all contain the real heft of this change so I'd start with the lowest level of them, `ConcurrentTaskQueue`, and move up so `StreamedZipDownloader` then `Zarr` then everywhere else.

## Testing
For the first change, I tried downloading Zarrs from various AICS, AIBS, and external datasets with success!
For the second change, I generated some large Zarrs & used some real ones from the EMT dataset (which are ~45gb).
& for both added a bunch of unit tests

### DIFF
--- a/packages/core/services/S3StorageService/test/S3StorageService.test.ts
+++ b/packages/core/services/S3StorageService/test/S3StorageService.test.ts
@@ -43,12 +43,12 @@ describe("S3StorageService", () => {
         it("parses virtualized s3 url", async () => {
             // Act
             const reformattedUrl = await s3StorageService.formatAsHttpResource(
-                "https://example.com/directory-name/path/to/file.zarr"
+                "https://animatedcell-test-data.s3.us-west-2.amazonaws.com/variance/10005.zarr"
             );
 
             // Assert
             expect(reformattedUrl).to.equal(
-                "https://example.com/directory-name%2Fpath%2Fto%2Ffile.zarr"
+                "https://s3.us-west-2.amazonaws.com/animatedcell-test-data/variance%2F10005.zarr"
             );
         });
     });


packages/core/state/interaction/logics.ts
```diff
@@ -69,6 +69,18 @@ import { fetchWithTimeout } from "../../hooks/useRemoteFileUpload";
...
 export const DEFAULT_QUERY_NAME = "New Query";
 
+/**
+ * Function for creating a message to display representing the total bytes to download
+ */
+const numberFormatter = annotationFormatterFactory(AnnotationType.NUMBER);
+function getBytesDisplay(bytes: number, hasUnknownSize = false): string {
+    const bytesWithUnits = numberFormatter.displayValue(bytes, "bytes");
+    if (hasUnknownSize) {
+        return `Unknown, but at least ${bytesWithUnits}`;
```
aswallace: I'm not sure if this text will make sense in context (it looks like atm this only gets rendered the `onProgress` message in line 360 since the conditional prevents it from rendering in 368)

PR #715, opened 2026-04-01
Fix plentiful file list fetches

### LINKED ISSUE #714: Fix file list fetch
File list fetches are occurring way more than they should - figure out why and fix it. Recreate this by scrolling a several thousand+ list

SeanDuHare:
## Context
When scrolling the file list in BFF many _many_ requests were being sent rather than just one or two when the user stops scrolling for a moment

Resolves #714

## Changes
The call for fetching files was wrapped in a `debounce` but the function was recreated on every render so it was effectively never accumulating any. This now wraps that in a React memoization.

## Testing
Tested via scrolling a ton - not really sure of a good unit test (open to ideas)


packages/core/components/FileList/index.tsx
```diff
@@ -146,7 +147,7 @@ export default function FileList(props: FileListProps) {
                 // Data source may not be prepared if the data source is taking longer to load
                 // than the component does to render. In this case, we can ignore the error.
                 // The component will re-render when the data source is prepared.
-                if (!(err as Error)?.message?.includes("Data source is not prepared")) {
+                if (!err?.message.includes("Data source is not prepared")) {
```
aswallace: I think I might have added the extra ? for `message` at some point because I was getting an uncaught error "cannot read property 'includes' of undefined" or something like that

Issue #712, opened 2026-04-01
Add support for date & datetime types


SeanDuHare:
BFF currently casts everything to strings - lets add date & datetime types to that if possible

Issue #711, opened 2026-04-01
Add support for duration types


SeanDuHare:
BFF currently casts everything to strings - lets add duration types to that if possible

Issue #710, opened 2026-04-01
Add support for boolean types


SeanDuHare:
BFF currently casts everything to strings - lets add boolean types to that if possible

Issue #709, opened 2026-04-01
Add support for number types (BIGINT, INTEGER, etc).


SeanDuHare:
BFF currently casts everything to strings - lets add number types to that if possible

PR #716, opened 2026-04-02
Add data type support for Parquet and JSON data sources (number, boolean, duration, date, datetime)

### LINKED ISSUE #60: Support all data types from CSV data sources
Currently only INTEGER, BIGINT, VARCHAR, and TEXT are definitely parsed to their correct type. We want all the other types to parse to the format that the explorer has gotten used to with FES

BrianWhitneyAI:
## Summary

Closes #60 (sub-issues #709, #710, #711, #712)

Adds data type support for Parquet and JSON data sources by mapping DuckDB column types to `AnnotationType` and generating correct SQL for each type. CSV is intentionally excluded — it uses `all_varchar=true` so all columns arrive as VARCHAR regardless of content. (opening a separate PR for support)

### Changes by type

- **Number (#709)**: Maps `DOUBLE`/`FLOAT`/`INTEGER`/etc. → `AnnotationType.NUMBER`; uses `NumberRangePicker`
- **Boolean (#710)**: Maps `BOOLEAN` → `AnnotationType.BOOLEAN`; uses list picker with true/false options
- **Duration (#711)**: Maps `INTERVAL` → `AnnotationType.DURATION`; **Parquet only** — JSON has no native interval type and `read_json_auto` does not infer INTERVAL from strings. Works around an Apache Arrow JS bug where `month_day_nano_interval` (typeId=11) loses nanoseconds — INTERVAL columns are re-queried as `CAST(EXTRACT(epoch FROM col) * 1000 AS BIGINT)` before Arrow serialization, in both `DatabaseService.runQuery()` (desktop) and `duckdb-worker.worker.ts queryWorker()` (web). Filter SQL uses `EXTRACT(epoch FROM col)::BIGINT * 1000 = ms`.
- **Date/Datetime (#712)**: Maps `DATE` → `AnnotationType.DATE`, `TIMESTAMP`/`TIMESTAMPTZ` → `AnnotationType.DATETIME`; fixes formatters to handle numeric-string ms-since-epoch values returned by duckdb-wasm; date range filters generate `CAST(col AS TIMESTAMPTZ) >= ... AND < ...` SQL.

### Bug fixes

- `AnnotationFilterForm`: `filter.value` is stored as a number after `valueOf()` coercion but `annotationValues` are always strings — stringifying both sides of the `Set.has()` lookup fixes DURATION (and BOOLEAN) filters not appearing selected after being applied.
- `FileFilter`: Passes `annotationType` through to enable type-aware SQL dispatch.

## Test plan

Smoke test files for each supported format are linked in the comments below.

- [x] Load `smoke_test_types.parquet` as a data source
- [x] Load `smoke_test_types.json` as a data source
- [x] For each: verify `Cell Count` → number range picker, `Is Aligned` → list picker with `true`/`false`, `Experiment Date` / `Captured At` → date range picker
- [x] Parquet only: verify `Acquisition Duration` → list picker showing `0M`–`19M`; selecting a value should appear checked
- [x] Run unit tests: `npm test` in `packages/core`

### DIFF
--- a/packages/core/components/DateRangePicker/index.tsx
+++ b/packages/core/components/DateRangePicker/index.tsx
@@ -116,7 +116,9 @@ export default function DateRangePicker(props: DateRangePickerProps) {
                     placeholder="Start of date range"
                     onSelectDate={(v) => (v ? onDateRangeSelection(v, null) : onReset())}
                     defaultDate={
-                        props.type == AnnotationType.DATE ? addTimeZoneOffset(startDate) : startDate
+                        props.type === AnnotationType.DATE
+                            ? addTimeZoneOffset(startDate)
+                            : startDate
                     }
                 />
                 <div className={styles.dateRangeSeparator}>
@@ -126,7 +128,7 @@ export default function DateRangePicker(props: DateRangePickerProps) {
                     placeholder="End of date range"
                     onSelectDate={(v) => (v ? onDateRangeSelection(null, v) : onReset())}
                     defaultDate={
-                        props?.type == AnnotationType.DATE ? addTimeZoneOffset(endDate) : endDate
+                        props?.type === AnnotationType.DATE ? addTimeZoneOffset(endDate) : endDate
                     }
                 />
                 <TertiaryButton


BrianWhitneyAI:
## Smoke test assets

For manual testing, a pre-built Parquet file with all supported column types is attached to a [pre-release](https://github.com/AllenInstitute/biofile-finder/releases/tag/test-assets-pr716):

- **[smoke_test_types.parquet](https://github.com/AllenInstitute/biofile-finder/releases/download/test-assets-pr716/smoke_test_types.parquet)** — contains `File Name` (STRING), `Cell Count` (NUMBER/DOUBLE), `Is Control` (BOOLEAN), `Acquisition Duration` (INTERVAL), `Date Created` (DATE), and `Last Modified` (TIMESTAMP) columns

To regenerate the parquet, the generator script is at: https://gist.github.com/BrianWhitneyAI/b1babccee917c4fc1c6d87ec22e6f9aa

```bash
pip install pyarrow duckdb
python generate_smoke_test_parquet.py
```

**What to verify:**
- Load `smoke_test_types.parquet` as a data source
- `Cell Count` → number range picker
- `Is Control` → list picker with `true`/`false`
- `Acquisition Duration` → list picker showing `1M`, `2M`, ..., `19M`; selecting a value should show it as checked
- `Date Created` / `Last Modified` → date range picker

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -173,22 +185,17 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
                     />
                 );
             case AnnotationType.NUMBER:
-                // File size is a special case where we don't have
-                // the ability to filter by range in the backend yet
-                // so we'll just let that case fall through to the string below
-                if (props.annotation.name !== AnnotationName.FILE_SIZE) {
```
BrianWhitneyAI: This is moved above.

packages/core/entity/FileFilter/index.ts
```diff
@@ -85,7 +96,38 @@ export default class FileFilter {
...
+                        this.annotationValue
+                    );
+                    return `CAST("${
+                        this.annotationName
+                    }" AS TIMESTAMPTZ) >= CAST('${startDate?.toISOString()}' AS TIMESTAMPTZ) AND CAST("${
+                        this.annotationName
+                    }" AS TIMESTAMPTZ) < CAST('${endDate?.toISOString()}' AS TIMESTAMPTZ)`;
+                }
+                if (this.annotationType === AnnotationType.DURATION) {
+                    return `EXTRACT(epoch FROM "${this.annotationName}")::BIGINT * 1000 = ${Number(
```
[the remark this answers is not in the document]
BrianWhitneyAI: DuckDB's INTERVAL type comes through Apache Arrow JS  as Arrow Type.Interval (typeId 11).  Arrow JS only reads the first 8 bytes of the interval value so it drops nanoseconds. To work around this, we re-run any query containing INTERVAL columns with those columns cast to milliseconds as BIGINT using EXTRACT(epoch FROM ...) * 1000.   So for example, an interval of 1 day 2 hours would arrive as 93600000 (ms).

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -154,9 +157,18 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
...
+    // FILE_SIZE is excluded: range filtering is not yet supported for it in the backend.
+    const typeHasDedicatedPicker =
+        props.annotation.name !== AnnotationName.FILE_SIZE &&
+        [AnnotationType.NUMBER, AnnotationType.DATE, AnnotationType.DATETIME].includes(
+            props.annotation.type as AnnotationType
+        );
+
     const searchFormType = () => {
-        // Use the checkboxes if values exist and are few enough to reasonably scroll through
-        if (items.length > 0 && items.length <= 100) {
```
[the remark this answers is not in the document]
aswallace: I'd be curious if there's a point at which the list picker becomes preferable for dates/numbers/etc. too (e.g., less than 10). That might be a UX/user question though

packages/core/entity/AnnotationFormatter/test/formatters.test.ts
```diff
@@ -70,6 +70,12 @@ describe("Annotation formatters", () => {
                 expect(dateTimeFormatter.displayValue(testCase.input)).to.equal(testCase.expected);
             })
         );
+
+        // the formatter coerces all-digit strings to Number before constructing the Date.
+        it("formats a numeric string ms-since-epoch value as a datetime", () => {
+            expect(dateTimeFormatter.displayValue("1645833600000")).to.not.equal("Invalid Date");
+            expect(dateTimeFormatter.displayValue("1645833600000")).to.be.a("string").and.not.empty;
```
SeanDuHare: Can this just be added to the list of test cases above (same for the Date annotation formatter)

packages/core/entity/FileFilter/index.ts
```diff
@@ -85,7 +96,39 @@ export default class FileFilter {
...
+                    DATE_RANGE_OPERATOR_REGEX.test(this.annotationValue)
+                ) {
+                    const { startDate, endDate } = extractDatesFromRangeOperatorFilterString(
+                        this.annotationValue
+                    );
+                    return `CAST("${
+                        this.annotationName
+                    }" AS TIMESTAMPTZ) >= CAST('${startDate?.toISOString()}' AS TIMESTAMPTZ) AND CAST("${
+                        this.annotationName
+                    }" AS TIMESTAMPTZ) < CAST('${endDate?.toISOString()}' AS TIMESTAMPTZ)`;
```
aswallace: I'm not sure if it's actually related to this, but I'm noticing a small bug where when you set a date filter and then click to edit the range, it bumps the end date forward by one each time the date picker renders. I remember seeing this bug before for FMS, and it had to do with the way we do exclusive end bound--I think ended up needing to do some funky -1/+1 on the dates to prevent it


https://github.com/user-attachments/assets/354430a4-536d-41c4-adb0-a1011c387ea6

Issue #718, opened 2026-04-02
Recognize Delta Lake tables


kmitcham:
## Use Case

Users should be able to drag and drop a Delta Table or pyarrow parquet file into BFF and have it work like parquet or csv files.

## Suggest a Solution

Update BFF to recognize and process Delta Tables correctly.

## Additional Details
Related to https://github.com/aics-int/fms-mongo-etl/issues/70

kmitcham:
Some time testing for opening, numbers in seconds for a 10,000,000 row file
delta:  82.0   104 
csv:    20.3    30
parq:   2.2       3
This was not using the native DuckDB read of delta, but rather the hack from https://github.com/AllenInstitute/biofile-finder/pull/738

SeanDuHare:
Closing this out in favor of detecting delta lake via URL

PR #725, opened 2026-04-07
Feature/718 read deltalake

kmitcham:
## Context
Update to allow opening delta lake tables from URIs
https://github.com/AllenInstitute/biofile-finder/issues/718 actual issue
https://github.com/aics-int/fms-mongo-etl/issues/70 x-repo parent
https://github.com/aics-int/fms-mongo-etl/issues/69 Parent

## Changes
Some changes to file type detection

## Testing
Limited manual testing was done pointing at 
https://staging-biofile-finder-datasets.s3.us-west-2.amazonaws.com/delta-table-example-file_path/
Which should have contents similar to
Index,Name,Age,file_path
1,Alice,25,/tmp/Alice
2,Bob,30,/tmp/Bob
3,Charlie,28,/tmp/Charlie
4,Diana,35,/tmp/Diana
5,Ethan,22,/tmp/Ethan
6,Fiona,31,/tmp/Fiona
7,George,27,/tmp/George
8,Scott,29,/tmp/Scott

### DIFF
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -107,11 +107,21 @@ export async function initializeDuckDB(logLevel: duckdb.LogLevel): Promise<duckd
         },
     });
 
-    // Enable native Delta Lake support for DuckDB queries.
+    // Try enabling native Delta Lake support for DuckDB queries.
+    // In duckdb-wasm, this extension may not be published for every runtime build.
     const setupConnection = await db.connect();
     try {
         await setupConnection.query("INSTALL delta");
         await setupConnection.query("LOAD delta");
+    } catch (err) {
+        const message = (err as Error)?.message || String(err);
+        if (message.includes("delta.duckdb_extension.wasm is not available")) {
+            console.warn(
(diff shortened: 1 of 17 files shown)


packages/core/services/DatabaseService/index.ts
```diff
@@ -422,11 +461,18 @@ export default abstract class DatabaseService {
         if (this.parquetDirectViewNames.has(dataSource)) {
             this.parquetDirectViewNames.delete(dataSource);
             await this.execute(`DROP VIEW IF EXISTS "${dataSource}"`);
+        } else if (this.deltaDirectViewNames.has(dataSource)) {
+            this.deltaDirectViewNames.delete(dataSource);
```
SeanDuHare: Could `parquetDirectViewNames` contain these delta files (since they are parquet?)

SeanDuHare:
It feels like this isn't necessarily reading Delta Lake files natively within DuckDB and instead we are managing the Delta Lake files at ingestion time and treating them as a partitioned block of Parquet files - is that accurate? I'm wondering if that'll have some negative side effects performance wise and around the comment I left about what "active" files means. How is the Delta Lake you generated partitioned (on what column)?

PR #732, opened 2026-04-09
Fix loading metadata descriptor files

### LINKED ISSUE #727: Metadata descriptor file inconsistently applied
## Description

The metadata descriptors don't always get fully ingested. [Here's an example](https://bff.allencell.org/app?c=Plate%3A0.25%2CSite%3A0.25%2CWell%3A0.25%2CBatch%3A0.25&source=%7B%22name%22%3A%22cpg0005-gerry-bioactivity_harmonized_metadata_v0_1.parquet+%284%2F6%2F2026+2%3A54%3A13+PM%29%22%2C%22type%22%3A%22parquet%22%2C%22uri%22%3A%22https%3A%2F%2Fstaging-biofile-finder-datasets.s3.us-west-2.amazonaws.com%2Fcpg0005-gerry-bioactivity_harmonized_metadata_v0_1.parquet%22%7D&sourceMetadata=%7B%22name%22%3A%22Broad%2BColumn%2BDescriptions.parquet+%284%2F6%2F2026+2%3A54%3A02+PM%29%22%2C%22type%22%3A%22parquet%22%2C%22uri%22%3A%22https%3A%2F%2Fstaging-biofile-finder-datasets.s3.us-west-2.amazonaws.com%2FBroad%2BColumn%2BDescriptions.parquet%22%7D) where the descriptions don't show up.

We're also currently unable to allow two queries to have different metadata descriptor sources; the database service expects a single source of truth and deletes the metadata tables when users switch between queries.

## Expected Behavior

Metadata descriptors should always get applied when the file is present and correctly formatted. See [here](https://bff.allencell.org/app?c=Plasmid+ID%3A0.11466848319709355%2CPlasmid+name%3A0.1809718437783833%2CTarget+Cell+Population%3A0.25%2CLABELED+REGION+COARSE%3A0.25%2CLABELED+REGION+FINE%3A0.25%2CLABELED+CELL+POPULATION%3A0.25&group=Enhancer+ID&group=Cargo&group=Delivery+Method&group=Experiment+Type&openFolder=%5B%22AiE0025h%22%5D&openFolder=%5B%22AiE0025h%22%2C%22SYFP2%22%5D&openFolder=%5B%22AiE0025h%22%2C%22SYFP2%22%2C%22RO%22%5D&openFolder=%5B%22AiE0025h%22%2C%22SYFP2%22%2C%22RO%22%2C%22EPI%22%5D&source=%7B%22name%22%3A%22Enhancer_AAVs%28in%29.csv+%284%2F8%2F2026+11%3A21%3A28+AM%29%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22https%3A%2F%2Fallen-genetic-tools.s3.us-west-2.amazonaws.com%2Fbio_file_finder%2FBFF_collections_csvs%2FEnhancer_AAVs%28in%29.csv%22%7D&sourceMetadata=%7B%22name%22%3A%22BFF_Viral_Description%28in%29.csv+%283%2F16%2F2025+2%3A16%3A05+PM%29%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22s3%3A%2F%2Fallen-genetic-tools%2Fbio_file_finder%2FBFF_collections_csvs%2FBFF_Viral_Description%28in%29.csv%22%7D) for an example with descriptors properly applied (e.g., the neuroglancer link works)

## Reproduction
May be worth noting that the example that works is two csvs, and the one that fails is a parquet

## Environment
- [x] Web
- [x] Desktop

aswallace:
## Context
The metadata descriptor files weren't being ingested/loaded for parquet files, because the only time we were properly checking for metadata descriptor files was during the normalization step (which doesn't happen for parquet). 
These descriptors *should* be loaded during the `fetchAnnotation` call, but we had an unsatisfiable conditional check for whether the data source should have descriptions: `dataSourceNames.includes(this.SOURCE_METADATA_TABLE)` is never true, because we don't include the metadata descriptor in the list of data sources for a query -- if we did, our current system would expect to see an aggregate table called "\<_name of actual data source_\>, \<_name of metadata descriptor file_\>".

We were also previously unable to allow two queries to have different metadata descriptor sources, because the `DatabaseService` always reused the name `this.SOURCE_METADATA_TABLE = "source_metadata"`.

resolves #727 

## Changes
Now includes the currently selected metadata source in the `DatabaseAnnotationService` so that we can pass it into the `fetchAnnotation` call to correctly check whether descriptors should be present. This also allows users to provide a different metadata descriptor file per query, and lets us keep track of which file we're currently referencing.

## Testing
Manually tested: 
- The parquet file mentioned in the tagged issue. Made sure descriptions show up now
- Open source datasets that have descriptions & custom links still work
- AICS FMS descriptions still show up
- Tested having two queries open with different metadata descriptor files, and verified can switch back and forth between them without losing access to the files.

### DIFF
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -470,11 +470,9 @@ export default abstract class DatabaseService {
         }
     }
 
-    public async deleteSourceMetadata(name?: string): Promise<void> {
-        // If a specific metadata file name is passed, delete it
-        // Otherwise, delete the currently selected one
-        if (name) await this.deleteDataSource(name);
-        else if (this.sourceMetadataName) this.deleteDataSource(this.sourceMetadataName);
+    public async deleteSourceMetadata(): Promise<void> {
+        // Avoid trying to delete a file that doesn't exist
+        if (this.sourceMetadataName) await this.deleteDataSource(this.sourceMetadataName);
         this.dataSourceToAnnotationsMap.clear();
     }
 
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -374,7 +374,7 @@ export default abstract class DatabaseService {
         if (isPreviousSource && this.hasDataSource(sourceMetadata.name)) {
             return;
         }
-        // If the data source needs replacing (e.g., we've been passed a new source with a URI), delete the old instance before preparing the new one
+        // If the metadata source is being replaced, delete the old instance before preparing the new one
         if (sourceMetadata.uri) {
             await this.deleteSourceMetadata();
             // Make sure we don't still have a cached version of the metadata source
@@ -388,6 +388,7 @@ export default abstract class DatabaseService {
                 );
             }
         }
+        // If the source doesn't have a uri, we should instead try to use the cached table
         this.sourceMetadataName = sourceMetadata.name;
     }
 
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -110,10 +110,9 @@ export async function initializeDuckDB(logLevel: duckdb.LogLevel): Promise<duckd
  */
 export default abstract class DatabaseService {
     public static readonly LIST_DELIMITER = ",";
-    protected readonly SOURCE_METADATA_TABLE = "source_metadata";
     protected readonly SOURCE_PROVENANCE_TABLE = "source_provenance";
     private static readonly ANNOTATION_TYPE_SET = new Set(Object.values(AnnotationType));
-    private sourceMetadataName?: string;
+    protected sourceMetadataName?: string;
     public sourceProvenanceName?: string;
     private currentAggregateSource?: string;
     // Initialize with AICS FMS data source name to pretend it always exists
@@ -288,7 +287,7 @@ export default abstract class DatabaseService {
 
         if (!type || !uri) {
             throw new DataSourcePreparationError(
-                `Lost access to the data source.\
+                `Lost access to data source "${name}".\
                 </br> \
                 Local data sources must be re-uploaded with each \
                 page refresh to gain access to the data source file \
@@ -334,7 +333,7 @@ export default abstract class DatabaseService {
 
         if (!type || !uri) {
             throw new DataSourcePreparationError(
-                `Lost access to the data source.\
+                `Lost access to data source "${name}".\
                 </br> \
                 Local data sources must be re-uploaded with each \
                 page refresh to gain access to the data source file \
@@ -372,18 +371,23 @@ export default abstract class DatabaseService {
 
     public async prepareSourceMetadata(sourceMetadata: Source): Promise<void> {
         const isPreviousSource = sourceMetadata.name === this.sourceMetadataName;
-        if (isPreviousSource) {
+        if (isPreviousSource && this.hasDataSource(sourceMetadata.name)) {
             return;
         }
-
-        await this.deleteSourceMetadata();
-        await this.prepareDataSourceWrapper(
-            {
-                ...sourceMetadata,
-                name: this.SOURCE_METADATA_TABLE,
-            },
-            true
-        );
+        // If the data source needs replacing (e.g., we've been passed a new source with a URI), delete the old instance before preparing the new one
+        if (sourceMetadata.uri) {
+            await this.deleteSourceMetadata();
+            // Make sure we don't still have a cached version of the metadata source
+            if (!this.hasDataSource(sourceMetadata.name)) {
+                await this.prepareDataSourceWrapper(
+                    {
+                        ...sourceMetadata,
+                        name: sourceMetadata.name,
+                    },
+                    true
+                );
+            }
+        }
         this.sourceMetadataName = sourceMetadata.name;
     }
 
@@ -411,8 +415,11 @@ export default abstract class DatabaseService {
         }
     }
 
-    public async deleteSourceMetadata(): Promise<void> {
-        await this.deleteDataSource(this.SOURCE_METADATA_TABLE);
+    public async deleteSourceMetadata(name?: string): Promise<void> {
+        // If a specific metadata file name is passed, delete it
+        // Otherwise, delete the currently selected one
+        if (name) await this.deleteDataSource(name);
+        else if (this.sourceMetadataName) this.deleteDataSource(this.sourceMetadataName);
         this.dataSourceToAnnotationsMap.clear();
     }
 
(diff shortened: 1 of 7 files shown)


packages/web/src/services/DatabaseServiceWeb/index.ts
```diff
@@ -202,14 +202,13 @@ export default class DatabaseServiceWeb extends DatabaseService {
                 payload: { name: dataSource },
             });
         });
-        const wrappedPromise = promise.finally(() => {
-            this.existingDataSources.delete(dataSource);
-            this.dataSourceToAnnotationsMap.delete(dataSource);
-        });
-        return wrappedPromise;
+        return promise;
```
aswallace, line 205: Unrelated change: This was duplicate behavior; these two deletes already happen by default during the database method

packages/core/services/DatabaseService/index.ts
```diff
@@ -426,18 +425,24 @@ export default abstract class DatabaseService {
 
     public async prepareSourceMetadata(sourceMetadata: Source): Promise<void> {
         const isPreviousSource = sourceMetadata.name === this.sourceMetadataName;
-        if (isPreviousSource) {
+        if (isPreviousSource && this.hasDataSource(sourceMetadata.name)) {
```
SeanDuHare, line 428: Do we still need to delete the source metadata a few lines below even when this data source is cached?
Wondering if it could be like this? Lmk your thoughts:
```
// Skip if this is just the same data source
const isPreviousSource = sourceMetadata.name === this.sourceMetadataName;
let isAlreadyCached = this.hasDataSource(sourceMetadata..name);
if (isPreviousSource && isAlreadyCached) return;

// Always delete old source metadata
await this.deleteSourceMetadata();

// Re-check if data source is already in memory so we can avoid re-fetching it
// necessary since the delete may have taken a while
isAlreadyCached = this.hasDataSource(sourceMetadata..name);
if (isAlreadyCached) {
     this.sourceMetadataName = sourceMetadata.name;
     return;
}

// Should never happen, but here to check to avoid failing silently
if (!sourceMetadata.uri) {
    throw new Error(`Lost reference to source metadata table for ${sourceMetadata.name}`);
}

// Actually prepare the data source
await this.prepareDataSourceWrapper(
                    {
                        ...sourceMetadata,
                        name: sourceMetadata.name,
                    },
                    true
);
```

Q: f using a cached source do we need to do the prepare data source step? doesn't seem like it?
Q: When would it not have a URI? Just when we are expecting to switch to a cached table?
Q: Don't we always want to delete the only source metadata if changing it out?

packages/core/services/DatabaseService/index.ts
```diff
@@ -465,8 +470,11 @@ export default abstract class DatabaseService {
         }
     }
 
-    public async deleteSourceMetadata(): Promise<void> {
-        await this.deleteDataSource(this.SOURCE_METADATA_TABLE);
+    public async deleteSourceMetadata(name?: string): Promise<void> {
```
SeanDuHare: Should we just force users to always supply it and they have to supply `this.sourceMetadataName`?

packages/core/services/DatabaseService/index.ts
```diff
@@ -1024,14 +1032,19 @@ export default abstract class DatabaseService {
...
         const aggregateDataSourceName = dataSourceNames.sort().join(", ");
         const hasAnnotations = this.dataSourceToAnnotationsMap.has(aggregateDataSourceName);
         const hasDescriptions = this.dataSourceToAnnotationsMap
             .get(aggregateDataSourceName)
             ?.some((annotation) => !!annotation.description);
-        const shouldHaveDescriptions = dataSourceNames.includes(this.SOURCE_METADATA_TABLE);
+        const shouldHaveDescriptions =
+            metadataSource && this.existingDataSources.has(metadataSource?.name);
         if (!hasAnnotations || (!hasDescriptions && shouldHaveDescriptions)) {
+            this.sourceMetadataName = metadataSource?.name;
```
SeanDuHare, line 1045: This could be undefined if we don't have annotations right?

packages/core/state/selection/logics.ts
```diff
@@ -642,10 +642,19 @@ const changeSourceMetadataLogic = createLogic({
...
-            await databaseService.prepareSourceMetadata(selectedSourceMetadata);
-        } else {
-            await databaseService.deleteSourceMetadata();
+        try {
+            if (selectedSourceMetadata) {
+                await databaseService.prepareSourceMetadata(selectedSourceMetadata);
+            } else {
+                await databaseService.deleteSourceMetadata();
+            }
+        } catch (err) {
```
SeanDuHare, line 651: 🙌

packages/web/src/services/DatabaseServiceWeb/duckdb-worker.worker.ts
```diff
@@ -310,8 +318,11 @@ export default class DatabaseServiceWebWorker extends DatabaseService {
             .get(aggregateDataSourceName)
             ?.some((annotation) => !!annotation.description);
 
-        const shouldHaveDescriptions = dataSourceNames.includes(this.SOURCE_METADATA_TABLE);
+        const shouldHaveDescriptions =
+            metadataSource && this.existingDataSources.has(metadataSource.name);
         if (!hasAnnotations || (!hasDescriptions && shouldHaveDescriptions)) {
+            // Make sure dbservice is referencing the correct metadata source
+            this.sourceMetadataName = metadataSource?.name;
```
SeanDuHare, line 325: Same comment as before - `metadataSource` seems like it could be undefined when `hasAnnotations` is false

packages/core/services/DatabaseService/index.ts
aswallace: It can be, but that's fine-- it just means that the fetchAnnotationDescriptions/fetchAnnotationTypes calls will return `{}` because there isn't a metadata descriptor file (e.g., [here](https://github.com/AllenInstitute/biofile-finder/pull/732/changes#diff-896daea5ed105e8ad194a958d065ff1d912bfc978d52c843f29dd9f638b18156R1095)), so then this function is only fetching the annotations themselves and not their descriptions

packages/core/services/DatabaseService/index.ts
aswallace: I think this almost works, the issue is just when switching between queries:
Let's say we have a query (Query 1) with metadata source 1 that we manually added (so it's a source where uri = the file)
- Add a new query (Query 2) with metadata source 2. Now we're switched to Query 2, and Query 1 has been dehydrated into its encoded SearchParams. When we encode the sources of queries, we only encode the uri if it's a link/string, not if it's a file/object.
- Currently, when we switch back to Query 1, we rehydrate its SearchParams, and we hit `prepareSourceMetadata`. We no longer have access to its uri/file because it didn't get encoded, but the metadata description table still exists. In this case, we don't want to `deleteSourceMetadata` because that removes the table, and then we'll hit the error and fail.

I think you're right that this means that if we know we're switching to a query where the table already exists, we shouldn't need to do the prepareDataSource step. But we also need to make sure we're only deleting the source if we've been given a new uri to replace it with, even when `sourceMetadata.name !== this.sourceMetadataName;`

SeanDuHare: Okay that makes sense!

SeanDuHare (approved):

packages/core/services/DatabaseService/index.ts
aswallace: I'm realizing I think I got rid of the use case where I needed to pass in the name, so I think this can actually be reverted. will test and make sure

aswallace: Reverted method signature in [f53330a](https://github.com/AllenInstitute/biofile-finder/pull/732/commits/f53330afa44eb0e79b5aa4e5fb7f88f04e2a9502)

kmitcham (approved):

pgarrison (approved):

BrianWhitneyAI (approved):

PR #734, opened 2026-04-09
Fix number range filter breaking on page refresh

BrianWhitneyAI:
## Summary
Fixes range filters returning empty results on page refresh. Allows for ingestion of old URLs.

Key changes:
- `FileFilter.annotationType` made `public readonly` so enrichment sites can read it
- `metadata/reducer`: `annotationsLoaded` flag — reset on `REQUEST_ANNOTATIONS`, set via new `ANNOTATIONS_READY` action
- `metadata/logics`: enrich `selection.filters` via `setFileFilters` in `receiveAnnotationsLogic`, then dispatch `annotationsReady()`
- `FileList`: guard `fetchTotalCount` behind `annotationsLoaded`

## Test plan
- [x] Apply a number range filter, refresh the page — confirm results are correct immediately without needing to re-submit the query
- [x] Use a shared URL containing a number range filter — confirm it resolves correctly on load

Depends on #735

### DIFF
--- a/packages/core/components/FileList/index.tsx
+++ b/packages/core/components/FileList/index.tsx
@@ -57,7 +57,7 @@ export default function FileList(props: FileListProps) {
     const fileGridColumnCount = useSelector(selection.selectors.getFileGridColCount);
     const isDisplayingSmallFont = useSelector(selection.selectors.getShouldDisplaySmallFont);
     const isColumnWidthOverflowing = useSelector(selection.selectors.isColumnWidthOverflowing);
-    const annotationsLoaded = useSelector(metadata.selectors.getAnnotationsLoaded);
+    const annotationsLoaded = useSelector(metadata.selectors.areAnnotationsLoaded);
     const [measuredNodeRef, measuredHeight, measuredWidth] = useLayoutMeasurements<
         HTMLDivElement
     >();
--- a/packages/core/components/FileList/index.tsx
+++ b/packages/core/components/FileList/index.tsx
@@ -16,7 +16,7 @@ import { FileView } from "../../entity/SearchParams";
 import FileSet from "../../entity/FileSet";
 import useLayoutMeasurements from "../../hooks/useLayoutMeasurements";
 import useFileAccessContextMenu from "../../hooks/useFileAccessContextMenu";
-import { selection } from "../../state";
+import { metadata, selection } from "../../state";
 
 import styles from "./FileList.module.css";
 
@@ -57,6 +57,7 @@ export default function FileList(props: FileListProps) {
     const fileGridColumnCount = useSelector(selection.selectors.getFileGridColCount);
     const isDisplayingSmallFont = useSelector(selection.selectors.getShouldDisplaySmallFont);
     const isColumnWidthOverflowing = useSelector(selection.selectors.isColumnWidthOverflowing);
+    const annotationsLoaded = useSelector(metadata.selectors.getAnnotationsLoaded);
     const [measuredNodeRef, measuredHeight, measuredWidth] = useLayoutMeasurements<
         HTMLDivElement
     >();
@@ -130,8 +131,10 @@ export default function FileList(props: FileListProps) {
         }
     }, [fileSelection, fileSet, height, fileGridColumnCount, rowHeight]);
 
-    // Get a count of all files in the FileList, but don't wait on it
+    // Get a count of all files in the FileList, but don't wait on it.
+    // Wait for annotations to load first so filters have correct annotationType before querying.
     React.useEffect(() => {
+        if (!annotationsLoaded) return;
         // Set the file count to null so that the UI will display the loading message until a
         // new count is retrieved. Otherwise, may display stale count from previous fileSet
         setTotalCount(null);
@@ -154,7 +157,7 @@ export default function FileList(props: FileListProps) {
         return () => {
             cancel = true;
         };
-    }, [fileSet]);
+    }, [annotationsLoaded, fileSet]);
 
     const fileFetchWrapper = React.useCallback(
         async (startIndex: number, endIndex: number) => {
--- a/packages/core/components/FileList/index.tsx
+++ b/packages/core/components/FileList/index.tsx
@@ -57,7 +57,7 @@ export default function FileList(props: FileListProps) {
     const fileGridColumnCount = useSelector(selection.selectors.getFileGridColCount);
     const isDisplayingSmallFont = useSelector(selection.selectors.getShouldDisplaySmallFont);
     const isColumnWidthOverflowing = useSelector(selection.selectors.isColumnWidthOverflowing);
-    const annotationsLoaded = useSelector(metadata.selectors.areAnnotationsLoaded);
+    const areAnnotationsLoaded = useSelector(metadata.selectors.areAnnotationsLoaded);
     const [measuredNodeRef, measuredHeight, measuredWidth] = useLayoutMeasurements<
         HTMLDivElement
     >();
@@ -134,7 +134,7 @@ export default function FileList(props: FileListProps) {
     // Get a count of all files in the FileList, but don't wait on it.
     // Wait for annotations to load first so filters have correct annotationType before querying.
     React.useEffect(() => {
-        if (!annotationsLoaded) return;
+        if (!areAnnotationsLoaded) return;
         // Set the file count to null so that the UI will display the loading message until a
         // new count is retrieved. Otherwise, may display stale count from previous fileSet
         setTotalCount(null);
@@ -157,7 +157,7 @@ export default function FileList(props: FileListProps) {
         return () => {
             cancel = true;
         };
-    }, [annotationsLoaded, fileSet]);
+    }, [areAnnotationsLoaded, fileSet]);
 
     const fileFetchWrapper = React.useCallback(
         async (startIndex: number, endIndex: number) => {
--- a/packages/core/components/FileList/test/FileList.test.tsx
+++ b/packages/core/components/FileList/test/FileList.test.tsx
@@ -14,7 +14,7 @@ import FileList from "..";
 
 describe("<FileList />", () => {
     it("Calls getCountOfMatchingFiles() on mount and properly updates state afterwards", async () => {
-        const state = mergeState(initialState, {});
+        const state = mergeState(initialState, { metadata: { annotationsLoaded: true } });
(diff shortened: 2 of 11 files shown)


SeanDuHare:
>annotationType was never included in FileFilter.toJSON(), so it was dropped on every persist/serialize cycle. On refresh, deserialized filters had no annotation type context, causing toSQLWhereString() to fall through to a regex match on the literal RANGE(...) string instead of generating the correct numeric range SQL.

I'm worried about guessing the annotation type - could we instead match the filter types with the annotation types on ingestion? I kind of feel like technically we don't need to encode the filter types since the data source will tell us the type and the filters should only ever be of that type. Does that make any sense?

Ah actually with your newest commit you are like 50% of the way there. I think if we just grab the annotation type from the DB rather than inferring it this would be more reliable

packages/core/state/index.ts
```diff
@@ -48,11 +48,7 @@ export const reduxLogicDependencies: Partial<ReduxLogicDeps> = {
     httpClient: axios,
 };
 
-export const reduxLogics = [
```
BrianWhitneyAI, line 51: This is getting autoformatted by prettier so Idk why it was like this originally.

SeanDuHare (approved):

aswallace (approved):

packages/core/state/metadata/test/logics.test.ts
```diff
@@ -11,18 +11,21 @@ import {
     requestDatasetManifest,
     RECEIVE_PASSWORD_MAPPING,
     requestPasswordMapping,
+    receiveAnnotations,
 } from "../actions";
 import metadataLogics from "../logics";
 import { initialState, interaction } from "../../";
 import DatasetService, { DataSource } from "../../../services/DataSourceService";
 import DatabaseServiceNoop from "../../../services/DatabaseService/DatabaseServiceNoop";
+import Annotation from "../../../entity/Annotation";
```
SeanDuHare, line 21: This order of imports is slightly off:

```
...etc
import { initialState, interaction } from "../../";
import { SET_COLUMNS, SET_FILE_FILTERS } from "../../selection/actions";
import DatasetService, { DataSource } from "../../../services/DataSourceService";
import Annotation from "../../../entity/Annotation";
import { AnnotationType } from "../../../entity/AnnotationFormatter";
import FileFilter, { FilterType } from "../../../entity/FileFilter";
```

packages/core/state/metadata/actions.ts
```diff
@@ -192,6 +192,20 @@ export function storeNewAnnotation(annotation: AnnotationResponseMms): StoreNewA
     };
 }
 
+/**
+ * ANNOTATIONS_READY
+ *
+ * Dispatched after RECEIVE_ANNOTATIONS has been processed and any filter enrichment
+ * has completed. Used to gate file queries until annotationType is correct on all filters.
+ */
+export const ANNOTATIONS_READY = makeConstant(STATE_BRANCH_NAME, "annotations-ready");
```
SeanDuHare: actions are marked by a verb usually so I think style wise this should be something like `mark-annotations-ready`

packages/core/components/FileList/index.tsx
```diff
@@ -57,6 +57,7 @@ export default function FileList(props: FileListProps) {
     const fileGridColumnCount = useSelector(selection.selectors.getFileGridColCount);
     const isDisplayingSmallFont = useSelector(selection.selectors.getShouldDisplaySmallFont);
     const isColumnWidthOverflowing = useSelector(selection.selectors.isColumnWidthOverflowing);
+    const annotationsLoaded = useSelector(metadata.selectors.getAnnotationsLoaded);
```
SeanDuHare: How does this function when switching between data sources, it relies on getting reset by requestAnnotations()? Would it be possible to just ask the databaseService if the annotations are loaded rather than having an additional action? My thought is that the databaseService should know this information already as part of its state so we could wrap that in a selector (see other comment)

packages/core/state/metadata/selectors.ts
```diff
@@ -6,6 +6,7 @@ import AnnotationName from "../../entity/Annotation/AnnotationName";
 
 // BASIC SELECTORS
 export const getAnnotations = (state: State) => state.metadata.annotations;
+export const getAnnotationsLoaded = (state: State) => state.metadata.annotationsLoaded;
```
SeanDuHare: If you do roll with [this comment](https://github.com/AllenInstitute/biofile-finder/pull/734/changes#r3074787949) I would do something like this (down below) to replace this selector:

```
export const areAnnotationsLoaded = createSelector(getPlatformDependentServices, getDataSources, ({ databaseService }, dataSources: Source[]) => (
    !!this.databaseService.getAnnotationsBySources(dataSources)
);
```


You would need to add the `getAnnotationsBySources` method (or something similar) to the databaseService of course

packages/core/components/FileList/index.tsx
SeanDuHare: Also FWIW this should be (I forget the SW word) but boolean-y like `areAnnotationsLoaded`

aswallace: I like this, but it's also making me wonder now if this is slightly misnamed? In the `RECEIVE_ANNOTATIONS` action, what we're actually waiting for isn't technically the annotations themselves but for the `FileFilters` to have `annotation.type` added to them

packages/core/state/metadata/actions.ts
BrianWhitneyAI: removed ANNOTATIONS_READY in [d4e2ec8](https://github.com/AllenInstitute/biofile-finder/pull/734/commits/d4e2ec8079dc9a0597b3c4506ac1cbf24fc4c6b8)

packages/core/state/metadata/selectors.ts
BrianWhitneyAI: used `createSelector(getAnnotations, annotations => annotations.length > 0)` in [d4e2ec8](https://github.com/AllenInstitute/biofile-finder/pull/734/commits/d4e2ec8079dc9a0597b3c4506ac1cbf24fc4c6b8)

packages/core/components/FileList/index.tsx
BrianWhitneyAI: resovled in [d4e2ec8](https://github.com/AllenInstitute/biofile-finder/pull/734/commits/d4e2ec8079dc9a0597b3c4506ac1cbf24fc4c6b8)

packages/core/state/metadata/test/logics.test.ts
BrianWhitneyAI: resolved in [d4e2ec8](https://github.com/AllenInstitute/biofile-finder/pull/734/commits/d4e2ec8079dc9a0597b3c4506ac1cbf24fc4c6b8)

packages/core/components/FileList/index.tsx
BrianWhitneyAI: resolved in [d4e2ec8](https://github.com/AllenInstitute/biofile-finder/pull/734/commits/d4e2ec8079dc9a0597b3c4506ac1cbf24fc4c6b8)

packages/core/components/FileList/test/FileList.test.tsx
```diff
@@ -61,4 +75,39 @@ describe("<FileList />", () => {
...
+        // pre-check
+        expect(getCountSpy.called).to.equal(false);
+        expect(() => getByText("Counting files...")).not.to.throw();
+
+        // act
+        store.dispatch(metadata.actions.receiveAnnotations([FILE_NAME_ANNOTATION]));
+        await logicMiddleware.whenComplete();
+
+        // assert
+        expect(getCountSpy.called).to.equal(true);
```
SeanDuHare, line 110: Might be cool to add onto this or as another test that this gets reset between annotation requests ex:

```
       // Pre-check
        expect(getCountSpy.called).to.equal(false);
        expect(() => getByText("Counting files...")).not.to.throw();

        // Act: first check
        store.dispatch(metadata.actions.receiveAnnotations([FILE_NAME_ANNOTATION]));
        await logicMiddleware.whenComplete();

       // Assert: count called once
        expect(getCountSpy.callCount).to.equal(1);
        expect(() => getByText("Counting files...")).to.throw;

       // Reset and sanity check
        store.dispatch(metadata.actions.requestAnnotations());
        await logicMiddleware.whenComplete();
        expect(getCountSpy.callCount).to.equal(1);
        expect(() => getByText("Counting files...")).to.not.throw;

        // Act: second check
        store.dispatch(metadata.actions.receiveAnnotations([FILE_NAME_ANNOTATION]));
        await logicMiddleware.whenComplete();

       // Assert: count called second time
        expect(getCountSpy.callCount).to.equal(2);
        expect(() => getByText("Counting files...")).to.throw;
    
```

packages/core/components/FileList/index.tsx
```diff
@@ -57,6 +57,7 @@ export default function FileList(props: FileListProps) {
     const fileGridColumnCount = useSelector(selection.selectors.getFileGridColCount);
     const isDisplayingSmallFont = useSelector(selection.selectors.getShouldDisplaySmallFont);
     const isColumnWidthOverflowing = useSelector(selection.selectors.isColumnWidthOverflowing);
+    const annotationsLoaded = useSelector(metadata.selectors.areAnnotationsLoaded);
```
SeanDuHare: this variable can (should?) be changed to match selector `areAnnotationsLoaded`

packages/core/state/metadata/selectors.ts
```diff
@@ -12,6 +12,12 @@ export const getEdgeDefinitions = (state: State) => state.metadata.edgeDefinitio
 export const getPasswordToProgramMap = (state: State) => state.metadata.passwordToProgramMap;
 
 // COMPOSED SELECTORS
+
+export const areAnnotationsLoaded = createSelector(
```
SeanDuHare, line 16: If this works then cool 🤷‍♂️ nice find :)

SeanDuHare (approved):

packages/core/components/FileList/index.tsx
BrianWhitneyAI: ah whoops thought I fixed that. [9bb5019](https://github.com/AllenInstitute/biofile-finder/pull/734/commits/9bb5019d54bb8c4f263a137fa73a3cca7df60664)

PR #735, opened 2026-04-10
Fix number range picker blocking input for out-of-bounds intermediate values

BrianWhitneyAI:
## Summary
Fixes an issue where typing in the number range picker would reset the input on every keystroke, making it impossible to enter values.

The root cause was an early `return` in `validateInput` that prevented `props.onChange` from firing when a value was temporarily out of bounds. React treats this as a controlled input with no change, so it reset the field back to its previous value on each keystroke.


BrianWhitneyAI:
I separated this out from the other PR as they resolve different issues. This is the most pressing issue!

aswallace (approved):

SeanDuHare (approved):

PR #738, opened 2026-04-13
Feature/718 read deltalake

kmitcham:
## Context
This PR adds Delta Lake support for URL-based data sources in environments where DuckDB-WASM may not have the native delta extension available, while preserving existing behavior for CSV/JSON/parquet sources.

## Changes

1. Added a new Delta resolver service that:
detects Delta table roots via _delta_log
resolves active parquet files by replaying Delta add/remove actions
supports probing via checkpoint/version discovery for object-store access patterns
2. Extended DatabaseServiceto support delta source types:
load delta roots by resolving active parquet files and creating direct views over parquet_scan(
track direct view kind (parquet vs delta) with unified direct-view bookkeeping
add delta-aware fast-path metadata checks for large direct views (avoids expensive full blank-row scans when metadata is decisive)
3. Updated source URL resolution flow:
centralized async type/name resolution through resolveNameAndTypeFromSourceUrl
removed duplicated inline delta-probing logic from UI call sites
removed unused export (parseSourceUrl now internal)

## Testing
Added test coverage for the new code, and manually tested against https://staging-biofile-finder-datasets.s3.us-west-2.amazonaws.com/delta-table-example-file_path/

This PR is at the outside limits of my understanding of Typescript, so if something seems wrong, it likely is.

### DIFF
--- a/packages/core/components/DataSourcePrompt/FilePrompt.tsx
+++ b/packages/core/components/DataSourcePrompt/FilePrompt.tsx
@@ -6,8 +6,7 @@ import { useDropzone } from "react-dropzone";
 
 import { SecondaryButton, TertiaryButton, TransparentIconButton } from "../Buttons";
 import Tooltip from "../Tooltip";
-import { Source, parseSourceUrl } from "../../entity/SearchParams";
-import DeltaTableService from "../../services/DeltaTableService";
+import { Source, resolveNameAndTypeFromSourceUrl } from "../../entity/SearchParams";
 
 import styles from "./FilePrompt.module.css";
 
@@ -78,27 +77,8 @@ export default function FilePrompt(props: Props) {
         async (evt?: React.FormEvent) => {
             evt?.preventDefault();
             if (dataSourceURL) {
-                const parsed = parseSourceUrl(dataSourceURL);
-                const deltaTableService = new DeltaTableService();
-                let type: Source["type"];
-
-                if (parsed.shouldProbeForDelta) {
-                    const isDelta = await deltaTableService.isDeltaTableRoot(
-                        parsed.pathWithoutParams
-                    );
-                    type = isDelta ? "delta" : "csv";
-                } else {
-                    type = parsed.extensionGuess || "csv";
-                    if (!parsed.extensionGuess) {
-                        console.warn(
-                            "Assuming the source is csv since no extension was recognized"
-                        );
-                    }
-                }
-
                 props.onSelectFile({
-                    name: parsed.name,
-                    type,
+                    ...(await resolveNameAndTypeFromSourceUrl(dataSourceURL)),
                     uri: dataSourceURL,
                 });
             }
--- a/packages/core/components/DataSourcePrompt/FilePrompt.tsx
+++ b/packages/core/components/DataSourcePrompt/FilePrompt.tsx
@@ -6,7 +6,12 @@ import { useDropzone } from "react-dropzone";
 
 import { SecondaryButton, TertiaryButton, TransparentIconButton } from "../Buttons";
 import Tooltip from "../Tooltip";
-import { Source, getNameAndTypeFromSourceUrl } from "../../entity/SearchParams";
+import {
+    Source,
+    getNameAndTypeFromSourceUrl,
+    resolveNameAndTypeFromSourceUrl,
+    shouldProbeForDeltaSourceUrl,
+} from "../../entity/SearchParams";
 
 import styles from "./FilePrompt.module.css";
 
@@ -74,9 +79,17 @@ export default function FilePrompt(props: Props) {
     }, [fileRejections]);
 
     const onEnterURL = throttle(
-        (evt?: React.FormEvent) => {
+        async (evt?: React.FormEvent) => {
             evt?.preventDefault();
             if (dataSourceURL) {
+                if (shouldProbeForDeltaSourceUrl(dataSourceURL)) {
+                    props.onSelectFile({
+                        ...(await resolveNameAndTypeFromSourceUrl(dataSourceURL)),
+                        uri: dataSourceURL,
+                    });
+                    return;
+                }
+
                 props.onSelectFile({
                     ...getNameAndTypeFromSourceUrl(dataSourceURL),
                     uri: dataSourceURL,
--- a/packages/core/components/DataSourcePrompt/FilePrompt.tsx
+++ b/packages/core/components/DataSourcePrompt/FilePrompt.tsx
@@ -6,12 +6,8 @@ import { useDropzone } from "react-dropzone";
 
 import { SecondaryButton, TertiaryButton, TransparentIconButton } from "../Buttons";
 import Tooltip from "../Tooltip";
-import {
-    Source,
-    getNameAndTypeFromSourceUrl,
-    resolveNameAndTypeFromSourceUrl,
-    shouldProbeForDeltaSourceUrl,
-} from "../../entity/SearchParams";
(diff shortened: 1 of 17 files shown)


packages/core/components/DataSourcePrompt/FilePrompt.tsx
```diff
@@ -74,11 +74,11 @@ export default function FilePrompt(props: Props) {
     }, [fileRejections]);
 
     const onEnterURL = throttle(
-        (evt?: React.FormEvent) => {
+        async (evt?: React.FormEvent) => {
```
BrianWhitneyAI, line 77: Throttle with an async event?

packages/core/services/DatabaseService/index.ts
```diff
@@ -81,6 +86,24 @@ export function getParquetFileNameSelectPart(
     return `${getFileNameFromPathExpression(`"${pathColumn}"`)} AS "${PreDefinedColumn.FILE_NAME}"`;
 }
 
+function getDirectViewColumnProjection(
```
BrianWhitneyAI, line 89: What is a Direct View here?

BrianWhitneyAI: what does this do?

packages/core/services/DatabaseService/index.ts
```diff
@@ -64,6 +65,10 @@ function getFileNameFromPathExpression(quotedPathColumn: string): string {
     return `COALESCE(NULLIF(${stripOme}, ''), ${quotedPathColumn})`;
 }
 
+function escapeSqlString(value: string): string {
```
BrianWhitneyAI, line 68: Why is this needed?

packages/core/components/DataSourcePrompt/test/DataSourcePrompt.test.tsx
```diff
@@ -20,7 +20,7 @@ describe("<DataSourcePrompt />", () => {
         expect(loadButton?.hasAttribute("disabled")).to.be.true;
     });
 
-    it("enables the load button after a file is selected", () => {
+    it("enables the load button after a file is selected", async () => {
```
BrianWhitneyAI, line 23: Why is this async now? what changed?

BrianWhitneyAI: same for the rest of the tests.

packages/core/entity/SearchParams/index.ts
```diff
@@ -73,20 +82,71 @@ export const DEFAULT_AICS_FMS_QUERY: SearchParamsComponents = {
...
+    const { extensionGuess, name } = parseSourceUrl(dataSourceURL);
     if (!extensionGuess) {
         console.warn("Assuming the source is csv since no extension was recognized");
-        extensionGuess = "csv";
     }
-    return { name, type: extensionGuess };
+    return { name, type: extensionGuess || "csv" };
 };
 
+export const resolveNameAndTypeFromSourceUrl = async (
```
BrianWhitneyAI, line 108: This needs some comments. Its not super clear why we need an async function to resolve details from a url. (I would argue its counterintuitive actully).

packages/core/entity/SearchParams/index.ts
```diff
@@ -73,20 +82,71 @@ export const DEFAULT_AICS_FMS_QUERY: SearchParamsComponents = {
...
+    dataSourceURL: string,
+    deltaTableService: Pick<DeltaTableService, "isDeltaTableRoot"> = new DeltaTableService()
+) => {
+    const { extensionGuess, name, pathWithoutParams } = parseSourceUrl(dataSourceURL);
+    if (extensionGuess) {
+        return { name, type: extensionGuess };
+    }
+
+    if (
+        pathWithoutParams.endsWith("/") &&
```
BrianWhitneyAI, line 118: ends with a "/'? Is that standard?/ a valid identifier?

packages/core/entity/SearchParams/index.ts
```diff
@@ -73,20 +82,71 @@ export const DEFAULT_AICS_FMS_QUERY: SearchParamsComponents = {
...
+        (await deltaTableService.isDeltaTableRoot(pathWithoutParams))
+    ) {
+        return { name, type: "delta" as const };
+    }
+
+    console.warn("Assuming the source is csv since no extension/format was recognized");
+    return { name, type: "csv" as const };
+};
+
+export const resolveSourcesFromUrls = async (sources: Source[]): Promise<Source[]> =>
```
BrianWhitneyAI, line 128: Also confused what is happening here.

packages/core/services/DatabaseService/index.ts
```diff
@@ -101,6 +124,28 @@ export async function initializeDuckDB(logLevel: duckdb.LogLevel): Promise<duckd
             forceFullHTTPReads: false,
         },
     });
+
+    // Try enabling native Delta Lake support for DuckDB queries.
+    // In duckdb-wasm, this extension may not be published for every runtime build.
+    // If unavailable, we fall back to manual Delta table support via DeltaTableService,
+    // which recovers active parquet files by replaying _delta_log entries.
+    const setupConnection = await db.connect();
```
BrianWhitneyAI, line 132: Just checking here. This tries to set up "delta" if it fails theres some default that is used but that default is already in place, nothing from here is used then?

packages/core/services/DatabaseService/index.ts
```diff
@@ -205,13 +253,42 @@ export default abstract class DatabaseService {
...
+                    ? duckdb.DuckDBDataProtocol.S3
+                    : duckdb.DuckDBDataProtocol.HTTP;
+
+                await this.database.registerFileURL(fileAlias, parquetUri, protocol, false);
+                parquetRefs.push(fileAlias);
+            }
+
+            await this.createDeltaDirectView(name, parquetRefs);
+            return;
+        }
```
BrianWhitneyAI, line 290: A description of the flow of delta would be very helpful here. If Im understanding correctly. You try and hit the deltalake file and get a collection of parquets then register each one, then create a view from the list of files.

packages/core/services/DeltaTableService/index.ts
```diff
@@ -0,0 +1,263 @@
...
+    return `${stripTrailingSlash(tableRootUri)}/_delta_log`;
+}
+
+function buildDeltaLogVersionFileUrl(deltaLogDir: string, version: number): string {
+    return `${stripTrailingSlash(deltaLogDir)}/${String(version).padStart(20, "0")}.json`;
+}
+
+function isMissingObjectError(err: unknown): boolean {
+    const status = (err as any)?.response?.status;
+    if (status === 404 || status === 403) {
```
BrianWhitneyAI, line 83: 403 is permission error right?

packages/core/services/DeltaTableService/index.ts
```diff
@@ -0,0 +1,263 @@
...
+                        : pathJoin(normalizedRootUri, removedPath);
+                    activeFiles.delete(encodeURI(resolvedPath));
+                }
+            }
+        }
+
+        return [...activeFiles].sort();
+    }
+
+    private listDeltaLogJsonFilesDefault = async (rootUri: string): Promise<string[]> => {
```
BrianWhitneyAI, line 185: per AI - "listDeltaLogJsonFilesDefault throws for non-S3 URLs, but isDeltaTableRoot catches all errors and returns false. The result is that a valid Delta table at a non-S3 HTTPS URL will be quietly treated as CSV by resolveNameAndTypeFromSourceUrl with no user-visible explanation. "

Issue #728, opened 2026-04-08
Add performance benchmarks


SeanDuHare:
We are adding more and more complex features into the application that could potentially cause slowdowns (or speedups!).

We need to come up with a way to test performance against all of the queries BFF performs against DuckDB-WASM. This can happen in multiple different ways - the most common I've seen is putting in a benchmark.js script somewhere that runs a bunch of tasks and outputs a passing or failing score for each that a GitHub workflow can then check and validate for any incoming PR but there are lots of ways this could be done.
**Some requirements**

- Must test the DuckDB-WASM version which relies on being constrained to the browser (this is probably hard to do in a script, kind of unsure at the moment)
- Must run all types of queries BFF does at varying scales (ex. 100k, 1m, 10m, 100m rows), column count, column value cardinality with some scores chosen for each for what is considered "acceptable" performance
- Needs to check performance across cloud and locally uploaded files
- Must watch out for caching and DB initialization time - this can cause queries to either appear slow or appear faster than they actually are
- Ideally would be able to be ran from a GH workflow
- Must be relative to network speed for cloud locations

PR #739, opened 2026-04-15
feat: query benchmark system with DuckDB-WASM

BrianWhitneyAI:
## Summary

Adds three tools for measuring and monitoring DuckDB-WASM query performance.

  - **Local benchmark** (`run-local.js`) — runs the task suite in headless Chromium against local or S3 fixtures, prints a p50/p95 table, and writes a result JSON for comparison
  - **CI regression workflow** (`benchmark.yml`) — `workflow_dispatch` that benchmarks two branches sequentially on the same VM and posts a Markdown diff to the workflow summary                                                                                                                
  - **Dev console timing** — enable per-query DuckDB timing in the running app via `localStorage.setItem("bff_query_timing", "1")`

### See [`dev-docs/07-query-benchmarking.md`](../blob/feature/query-benchmark/dev-docs/07-query-benchmarking.md) for full usage. 
---
## Description of added code

  CI / Workflow                                                                                                                                                                                                                                                                               
  - benchmark.yml — new workflow_dispatch workflow; checks out two branches sequentially on the same VM and posts a Markdown comparison to the step summary                                                                                                                                      
                                                                                                                                                                                                                                                                                                 
  Browser-side benchmark bundle (packages/web/benchmark/)                                                                                                                                                                                                                                        
  - src/index.ts — Playwright page entry; registers sources, runs warmup + timed iterations in shuffled order, signals results back                                                                                                                                                              
  - src/tasks.ts — 11 benchmark tasks defined at the service layer (same code paths as the real app)                                                                                                                                                                                             
  - src/types.ts — shared config/result types                                                                                                                                                                                                                                                    
  - index.html — minimal HTML host for the benchmark bundle                                                                                                                                                                                                                                      
  - webpack.config.js — webpack config for the benchmark bundle
                                                                                                                                                                                                                                                                                                 
  Node.js scripts (packages/web/scripts/)          
  - run-local.js — developer tool; local or S3 fixtures, any scale, side-by-side cloud vs local (--full)                                                                                                                                                                                         
  - run-regression.js — CI tool; local fixtures only, writes branch-stamped JSON                                                                                                                                                                                                                 
  - compare-results.js — diffs two JSONs into a Markdown table with regression badges                                                                                                                                                                                                            
  - summarize-results.js — pretty-prints a single result file                                                                                                                                                                                                                                    
  - lib/run-benchmark-page.js — shared Playwright runner used by the two script tools above                                                                                                                                                                                                      
 
## Important changes to review                                                                                                                                                                                                                                                                                                
  Core service layer (packages/core/)                                                                                                                                                                                                                                                            
  - DatabaseAnnotationService, DatabaseFileService, DatabaseService — SQL-building logic extracted into exported functions so the benchmark calls the exact same SQL paths as the app                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                 
  DuckDB web worker (packages/web/src/services/DatabaseServiceWeb/)                                                                                                                                                                                                                              
  - duckdb-worker.worker.ts — adds query timing instrumentation: queryTimingEnabled flag, accumulatedTimings map, per-query console logging, enableQueryTiming() / clearTimings() / sumTimings() / clearAnnotationCache() for benchmark use
  - index.ts — wires localStorage flag to timing activation on init                                                                                                                                                                                                                              
  - types.ts — minor type updates for the INIT payload


packages/web/benchmark/src/queries.ts
```diff
@@ -0,0 +1,96 @@
...
+                            FilterType.DEFAULT,
+                            AnnotationType.NUMBER
+                        ),
+                    ],
+                }),
+                0,
+                100
+            ),
+    },
+    // DatabaseFileService.getFiles() — multiple FileFilters
```
SeanDuHare: Perhaps a date range test too?

packages/web/scripts/run-benchmark.js
```diff
@@ -0,0 +1,271 @@
...
+ * run-benchmark.js
+ *
+ * Builds the benchmark bundle, serves it in a local HTTP server with the
+ * Cross-Origin isolation headers required for SharedArrayBuffer (and therefore
+ * DuckDB-WASM's pthread bundle), then uses Playwright to run the benchmark
+ * in real Chromium and collect the results.
+ *
+ * The run happens in two phases:
+ *   Phase 1 – In-memory: synthetic tables at multiple scales × schema configs.
+ *              The browser also exports a small parquet fixture at the end.
```
SeanDuHare: What does this line mean?

packages/web/scripts/run-benchmark.js
```diff
@@ -0,0 +1,271 @@
+/**
+ * run-benchmark.js
+ *
+ * Builds the benchmark bundle, serves it in a local HTTP server with the
+ * Cross-Origin isolation headers required for SharedArrayBuffer (and therefore
+ * DuckDB-WASM's pthread bundle), then uses Playwright to run the benchmark
+ * in real Chromium and collect the results.
+ *
+ * The run happens in two phases:
+ *   Phase 1 – In-memory: synthetic tables at multiple scales × schema configs.
```
SeanDuHare: (Double check this) but I think we always load in Parquet files via cloud out of memory now so perhaps these tests aren't necessary (_or_ perhaps better yet could be only run if a flag to run them is provided to the test)

packages/web/scripts/run-benchmark.js
```diff
@@ -0,0 +1,271 @@
...
+ * We use COEP: credentialless (rather than require-corp) so that the
+ * cross-origin jsDelivr CDN resources used by @duckdb/duckdb-wasm load
+ * without needing a Cross-Origin-Resource-Policy header from the CDN.
+ *
+ * Range requests are supported so DuckDB-WASM can perform partial reads from
+ * the parquet fixture file (mirroring real S3 / cloud parquet access).
+ */
+function startServer() {
+    return new Promise((resolve, reject) => {
+        const server = http.createServer((req, res) => {
```
SeanDuHare: Hmm do we have any way of knowing how accurate this is? A major part of why these tests would be useful is to know when a query is too slow to be actually run by a user so if this isn't accurate then we might hit some issues

SeanDuHare (commented):
didn't finish, but here are some thoughts so far

packages/web/benchmark/src/index.ts
```diff
@@ -0,0 +1,354 @@
...
+    // --- Cloud benchmark (one pass per scale) ---
+    const cloudResults: CloudQueryResult[] = [];
+
+    for (const scale of CLOUD_FIXTURE_SCALES) {
+        const cloudFixtureUrl = cloudFixtureUrls[scale];
+        if (!cloudFixtureUrl) continue;
+
+        const cloudFile = `cloud_fixture_${scale}.parquet`;
+        setStatus(`Registering cloud fixture ${scale.toLocaleString()} rows: ${cloudFixtureUrl}`);
+        await db.registerFileURL(
```
SeanDuHare: Is this registration not included in the benchmark result? I think a critical test is the init of the parquet file in the cloud _**something to note**_ I have no idea why this is, maybe DuckDB explains somewhere, but I tend to see the init for cloud files be quick and then the first query against that file be slow. I think it waits to request the actual header of the parquet until then

packages/web/scripts/run-benchmark.js
BrianWhitneyAI: When doing a cloud out of memory we do bound ourselves by the network connection. (Ive even seen variance on the same runner). So running in memory is a little more isolated against specific query efficiency.  We could do more specific flags, not sure exactly what we would want

packages/web/scripts/run-benchmark.js
BrianWhitneyAI: At the end, the browser serializes the generated parquet fixture back to Node.js via CDP so it can be served in Phase 2. Its basically just a time save

SeanDuHare:
Remembered I wasn't sure if I checked the following in this, can you add the following if it isn't yet (or just lmk your thoughts):
* add to the test matrix testing for cardinality? "Metadata cardinality defines the number of unique values in a data field (column) or the numeric relationship between entities in a dataset"
* add testing varying column amounts (I think it is always 100 right now?)

packages/web/benchmark/src/index.ts
BrianWhitneyAI: Registration is included! (This comment may have predated me adding it im not sure). I was also able to observe this behavior with cloud files.

packages/web/scripts/run-benchmark.js
BrianWhitneyAI: I added in an option to turn on query timings from the console so you can compare the benchmark to real timing. Theyre pretty accurate now

BrianWhitneyAI:
**BFF Query Benchmark — Local Run**
`feature/query-benchmark` @ `4d209f5b` · 2026-04-24 · DuckDB init: 822ms
_Timings shown as p50 (1 iteration). Scales: local fixtures served over localhost; cloud via S3._

| Query | 100k-cloud | 100k-local | 1m-cloud | 1m-local | 10m-cloud | 10m-local |
|---|---|---|---|---|---|---|
| **registration** | 55.9ms | 6.91ms | 589.3ms | 17.0ms | 1023.2ms | 118.6ms |
| fetch_annotations | 1.82ms | 2.12ms | 1.64ms | 2.02ms | 1.63ms | 2.00ms |
| open_filter_picker_low_cardinality | 2.49ms | 3.85ms | 14.1ms | 22.0ms | 97.2ms | 203.5ms |
| open_filter_picker_medium_cardinality | 2.97ms | 7.73ms | 57.1ms | 22.3ms | ⚠️ 18218.9ms | 185.3ms |
| open_filter_picker_high_cardinality | 6.06ms | 6.29ms | 1679.8ms | 65.3ms | ⚠️ 40263.8ms | 1029.3ms |
| browse_file_list | 40.1ms | 85.9ms | 55.4ms | 93.1ms | ⚠️ 8676.3ms | 201.2ms |
| sort_file_list | 82.5ms | 178.4ms | ⚠️ 111064.3ms | 1594.8ms | ⚠️ 288693.8ms | 5817.3ms |
| sort_file_list_large_page | 85.5ms | 178.7ms | ⚠️ 120892.8ms | 1914.3ms | ⚠️ 1932389.7ms | 18555.9ms |
| filter_count | 11.4ms | 9.67ms | 65.1ms | 74.9ms | 611.1ms | 729.9ms |
| filter_browse | 41.2ms | 75.7ms | 54.1ms | 91.3ms | ⚠️ 7109.0ms | 209.6ms |
| null_group_count | 2.89ms | 2.32ms | 6.87ms | 6.57ms | 61.5ms | 55.4ms |
| change_grouping | 299.5ms | 412.8ms | 777.6ms | 906.8ms | 5549.5ms | 5848.6ms |
| expand_folder | 9.78ms | 11.3ms | 68.7ms | 88.8ms | 652.7ms | 868.9ms |

**Observations:**
- Cloud vs. local gap widens sharply at 10M rows for sort/browse queries — expected since cloud reads require S3 range requests per query while local reads hit the in-memory buffer. The sort queries at 10m-cloud (288s, 1932s) are likely doing full-table downloads on each iteration with no warm cache.
- Filter and count queries scale well — `filter_count` and `null_group_count` stay fast even at 10M because DuckDB can prune row groups.
- `open_filter_picker_high_cardinality` at 1m-cloud (1679ms) and 10m-cloud (40s) confirms that high-cardinality `DISTINCT` scans are the most expensive user-facing operation at scale.

packages/core/services/DatabaseService/index.ts
```diff
@@ -686,7 +690,7 @@ export default abstract class DatabaseService {
 
     // Create a view over the parquet file that exposes columns under predefined names (e.g. "File Path")
     // and adds hidden_bff_uid.
-    private async createParquetDirectView(name: string): Promise<void> {
+    protected async createParquetDirectView(name: string): Promise<void> {
```
BrianWhitneyAI: Loosened from private to protected so the benchmark subclass in packages/web/benchmark can call it directly to set up fixtures.

SeanDuHare (approved):
okay so I played with a bit and it seems to work really well! I'd like to try this against the structured metadata branch and see how it does in a more real scenario however I think this PR as is is good and any adjustment that come up could happen after

packages/web/scripts/run-local.js
```diff
@@ -0,0 +1,150 @@
...
+    ]);
+} else {
+    sources = Object.entries(URL_MAP)
+        .filter(([label, url]) => Boolean(url) && (!scaleArg || label === scaleArg))
+        .map(([label, url]) => ({ label, url }));
+}
+
+if (sources.length === 0) {
+    console.error(
+        "No real parquet URLs provided.\n" +
```
pgarrison, line 72: This error message seems to leave out the possibility that the user just forgot to set `--scale`?

packages/web/scripts/lib/run-benchmark-page.js
```diff
@@ -0,0 +1,227 @@
...
+                        fs.createReadStream(fixturePath).pipe(res);
+                    }
+                } catch {
+                    res.writeHead(404);
+                    res.end("Not found: " + relPath);
+                }
+                return;
+            }
+
+            const fullPath = path.join(DIST_DIR, relPath);
```
pgarrison, line 85: Probably not a real issue since this web server isn't exposed to the internet, but just want to flag that this pattern is a security risk, since `relPath` could be something like `../../../../.aws/credentials`.

packages/web/benchmark/src/index.ts
```diff
@@ -0,0 +1,189 @@
...
+            );
+        }
+        await service.prepareDataSources(
+            [{ name: source.label, type: "parquet", uri: localFile ?? source.url }],
+            /* skipNormalization */ true
+        );
+        const registrationMs = performance.now() - regStart;
+
+        const queries = await benchmarkSource(service, source.label, iterations, warmupRounds);
+        sources.push({ label: source.label, registrationMs, queries });
```
pgarrison, line 169: I think I mainly want to test cold start scenarios, since that's what users will encounter first. I can do that with warmups = 0 and iterations = 1, but there's also significant noise to these measurements, so I want multiple iterations. Unfortunately, with warmups = 0 and iterations > 1, only the first one is a cold start, because there's a single duckdb instance shared by all the iterations.

So for my use case, I think I'd prefer `iterations` to be handled at the `run-regression.js` level, so that each iteration has a separate page load, instead of here. Ideally I'd want each iteration to start a separate chromium instance, since I am also wary of potential browser-level caching of network responses. That isn't consistent with also having warmup rounds, so if we want both patterns you could just do the route I'm suggesting if `warmups = 0`.

BrianWhitneyAI: Is this covered enough by being able to run multiple single run no warm up runs then use the comparison tool to compare?

packages/web/benchmark/src/tasks.ts
```diff
@@ -0,0 +1,155 @@
...
+        name: "fetch_annotations",
+        resetAnnotationCache: true,
+        run: (a) => a.fetchAnnotations(),
+    },
+
+    // Opening a filter picker — three cardinality tiers because query time varies
+    // significantly depending on how many distinct values DuckDB must collect.
+    // cell_line:     5 distinct values  (low)
+    // experiment_id: 100 distinct values (medium)
+    // focus_score:   ~unique per row     (high — near-continuous float)
```
pgarrison, line 62: FYI, at 10M rows, 64-bit floats should actually be unique per row, 32-bit floats should have less than 1% collisions, and 16-bit floats would be inappropriate.

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -10,6 +10,28 @@ import IncludeFilter from "../../../entity/FileFilter/IncludeFilter";
 import { Source } from "../../../entity/SearchParams";
 import SQLBuilder from "../../../entity/SQLBuilder";
 
+/**
+ * SQL used by fetchFilteredValuesForAnnotation — exported so the benchmark can run the same query.
+ */
+export function buildDistinctValuesSQL(
```
pgarrison: I'm confused because this function doesn't appear to be used in `tasks.ts` or anywhere outside this file.

packages/core/services/DatabaseService/index.ts
```diff
@@ -81,6 +81,16 @@ export function getParquetFileNameSelectPart(
     return `${getFileNameFromPathExpression(`"${pathColumn}"`)} AS "${PreDefinedColumn.FILE_NAME}"`;
 }
 
+/** SQL used by fetchAnnotations — exported so the benchmark can run the same query. */
+export function buildFetchAnnotationsSQL(tableName: string): string {
```
pgarrison: This function doesn't appear to be used in `tasks.ts` or anywhere outside this file.

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -18,6 +18,31 @@ import FileDetail from "../../../entity/FileDetail";
...
+/** SQL used by getCountOfMatchingFiles — exported so the benchmark can run the same query. */
+export function buildGetCountSQL(dataSourceNames: string | string[], fileSet: FileSet): string {
+    return fileSet
+        .toQuerySQLBuilder()
+        .select("COUNT(*) AS num_files")
+        .from(dataSourceNames)
+        .removeOrderBy()
+        .toSQL();
+}
+
```
pgarrison: These functions don'tt appear to be used in `tasks.ts` or anywhere outside this file.

packages/web/benchmark/src/tasks.ts
```diff
@@ -0,0 +1,155 @@
...
+    {
+        name: "open_filter_picker_medium_cardinality",
+        run: (a) => a.fetchValues("experiment_id"),
+    },
+    {
+        name: "open_filter_picker_high_cardinality",
+        run: (a) => a.fetchValues("focus_score"),
+    },
+
+    // File list: default view, no filter or sort.
```
pgarrison, line 76: I want to highlight a couple aspects of this design for discussion.

1. The `tasks.ts` file needs to correspond to the actual tasks relevant to users. I imagine BFF might see drift, where a couple years from now the tasks here don't quite match tasks users do.
2. The current tasks defined here are substantially more granular than real user tasks. A user might be interested in time from adding a new data source to the first files are displayed, which is a combination of multiple tasks here. This design comes with some tradeoffs.
    1. Performance improvements that work by changing which combination of queries are part of a high-level task will not be captured here.
    2. The risk of drift between the task definitions and the actual codebase is partially mitigated: the tasks here are each a single function call to a function used in the codebase.

If we go ahead with this, I think it would be useful to write some design rationale and instructions for future devs, perhaps in a comment at the top of `tasks.ts`.

The main alternative design pattern I imagine would be to use playwright to send user events to BFF in the browser and thereby get metrics that reflect all the work that actually happens. This would have its own set of tradeoffs.
1. Brittle to UI changes. For example, something like the tutorial popup might need to be closed before getting started.
2. Useful for frontend correctness testing in addition to performance testing.

Pinging @SeanDuHare also

packages/core/services/FileService/DatabaseFileService/index.ts
BrianWhitneyAI: I may have mislead you through slack, I was referencing these queries but what I meant was the service layer functions. I had them mixed up in my mind. My original design was to test the queries we built against some database, but that doesn't actually emulate the experience of the user since they're waiting on the actual service call. So I moved the tasks up to the service layer so calling through the service layer gives you the SQL build + the query + the result mapping. The exported functions were essentially offering a way to call just the middle part, which didn't really work. We could revert this change but I actually kind of like it stylistically. Open to suggestions here though.

packages/web/benchmark/src/tasks.ts
BrianWhitneyAI: I think that what you're describing is a different scope. #728 is the parent issue here.  The core request is "We need to come up with a way to test performance against all of the queries BFF performs against DuckDB-WASM".

packages/web/src/services/DatabaseServiceWeb/duckdb-worker.worker.ts
```diff
@@ -260,8 +263,25 @@ export default class DatabaseServiceWebWorker extends DatabaseService {
...
+                // can be grouped by logical operation rather than raw SQL text.
+                const labelMatch = sql.match(/^--\s*(.+)\n/);
+                const label = labelMatch ? labelMatch[1].trim() : "query";
+                const existing = accumulatedTimings.get(label) ?? [];
+                accumulatedTimings.set(label, [...existing, elapsed]);
+                // Strip the label comment from the first line, collapse whitespace,
+                // and truncate so the console log stays readable.
+                const body = sql
+                    .replace(/^--[^\n]*\n/, "") // remove `-- label\n` prefix
+                    .replace(/\s+/g, " ") // collapse newlines/indentation
```
pgarrison, line 281: Queries often come with whitespace at the start after removing the prefix. 
```suggestion
                    .replace(/\s+/g, " ") // collapse newlines/indentation
                    .trim()
```

packages/web/src/services/DatabaseServiceWeb/duckdb-worker.worker.ts
```diff
@@ -260,8 +263,25 @@ export default class DatabaseServiceWebWorker extends DatabaseService {
...
+                    .slice(0, 100);
```
pgarrison: I encountered a query I was interested in but I couldn't tell which one it was from just the first 100 characters. I suggest increasing or removing this.
```suggestion
                    .slice(0, 200);
```

packages/web/src/services/DatabaseServiceWeb/duckdb-worker.worker.ts
```diff
@@ -260,8 +263,25 @@ export default class DatabaseServiceWebWorker extends DatabaseService {
...
+                console.log(`[duckdb] ${elapsed.toFixed(1)}ms — [${label}] ${body}`);
```
pgarrison:
```suggestion
                console.log(`[duckdb] ${elapsed}ms — [${label}] ${body}`);
```

Since `elapsed` is computed from `performance.now()`, which returns an integer, the `toFixed(1)` would always add `.0` to the string.

packages/web/scripts/run-local.js
```diff
@@ -0,0 +1,150 @@
...
+// fixtures, single scale or all scales, and side-by-side cloud vs local comparison (--full).
+
+"use strict";
+
+const path = require("path");
+const fs = require("fs");
+const { execSync } = require("child_process");
+const { runBenchmarkPage } = require("./lib/run-benchmark-page");
+
+const LOCAL_FIXTURE_MAP = {
```
pgarrison, line 11: This is duplicated with `run-regression.js` FYI.

packages/core/services/FileService/DatabaseFileService/index.ts
pgarrison: I don't think it makes sense to export a function that is only used locally.

Without the export, there's a subjective style question, where I lean toward smaller PR diffs and code locality.

packages/web/benchmark/src/tasks.ts
pgarrison: Gotcha. Then the main point of my comment is probably:

> If we go ahead with this, I think it would be useful to write some design rationale and instructions for future devs, perhaps in a comment at the top of tasks.ts.

packages/web/benchmark/src/index.ts
pgarrison: I think the comparison tool only supports two input files? So there's another step required where the multiple run outputs would need to be aggregated? And how would that work on github actions?

packages/core/services/FileService/DatabaseFileService/index.ts
BrianWhitneyAI: ok Ill revert these then!

packages/web/benchmark/src/tasks.ts
BrianWhitneyAI: added in [ba763a0](https://github.com/AllenInstitute/biofile-finder/pull/739/commits/ba763a00466d7ac94b84c66803f80cdb06479f22)

packages/core/services/FileService/DatabaseFileService/index.ts
BrianWhitneyAI: removed in [ba763a0](https://github.com/AllenInstitute/biofile-finder/pull/739/commits/ba763a00466d7ac94b84c66803f80cdb06479f22)

dev-docs/07-query-benchmarking.md
```diff
@@ -0,0 +1,118 @@
...
+
+```bash
+cd packages/web
+npx playwright install chromium --with-deps
+```
+
+**Download local fixtures** (one time; ~500 MB total)
+
+```bash
+BASE=https://staging-biofile-finder-datasets.s3.us-west-2.amazonaws.com/benchmark-fixtures/v1
```
pgarrison, line 23: @BrianWhitneyAI can you elaborate on how you see this versioning scheme being used? 

For context, here's what I did getting this to work for testing my aggregate parquet changes.
1. I wanted to add a new test file `synthetic-20m.parquet`. I uploaded to the `v1/` folder because there wasn't anything with that name already to create a conflict.
2. To get the gh action to recognize the new test file, I changed the name of the key in `benchmark.yml` to `v1.1`
3. I realized that I prepared the test parquet file incorrectly, so I regenerated it and uploaded the new file to the same path `v1/synthetic-20m.parquet`. Again, I kept using the `v1` folder because this was all for work on my feature branch, and the first version of the file was no good.

My approach has some potential challenges with concurrent work. In step 2, maybe two people working on different features simultaneously get the idea to bump the cache key. I'd recommend cache key version numbers like `v1.1-83fc7ad` instead. In step 3, it's possible (not so likely) that another dev had seen the original `synthetic-20m.parquet` and started using it before I changed it.

There's also the option of object-level versioning provided by S3. I don't know if this bucket has versioning enabled (I don't have permission to check).

SeanDuHare: I'm in favor of this change. It does seem unlikely given our current usage that we'd encounter this issue, it seems like a small enough change to be worth the potential headache/confusion/wasted dev hours in the future should we encounter it.

packages/web/benchmark/src/index.ts
SeanDuHare: I agree that benchmarking cold start scenarios is key since we want to try to capture what the worst case scenarios could be - perhaps rather than warmups it could just be a boolean of coldStart and it'll spin up a new instance on each iteration (including when iterations > 1). I'm also ok with Philip's suggestion or the comparison tool getting an upgrade to compare multiple files (though that might be the least favorable long term, unsure).

pgarrison: I started work on cold start benchmarking in this PR https://github.com/AllenInstitute/biofile-finder/pull/808

packages/web/scripts/run-local.js
pgarrison: In #806 I rewrote this logic in both `run-local` and `run-regression`, so if we want to DRY it, I'd suggest doing it as a follow-up to that PR?

packages/web/scripts/summarize-results.js
```diff
@@ -0,0 +1,94 @@
...
+            process.env.BENCHMARK_BRANCH ||
+            execSync("git rev-parse --abbrev-ref HEAD", { stdio: "pipe" }).toString().trim();
+        const slug = branch.replace(/[^a-zA-Z0-9._-]/g, "-").replace(/-+/g, "-");
+        const stamped = path.join(__dirname, "..", `benchmark-results-${slug}.json`);
+        if (fs.existsSync(stamped)) return stamped;
+    } catch {
+        /* fall through */
+    }
+
+    return path.join(__dirname, "..", "benchmark-results.json");
```
pgarrison:
```suggestion
    } catch (e) {
        if (e instanceof SomeSpecificError) {
            return path.join(__dirname, "..", "benchmark-results.json");
        } else {
            throw e;
        }
    }
```

This silent error-swallowing is an LLM anti-pattern. If there's an error, we generally would like it to bubble up so that a dev can inspect and address it directly. If there's a specific error case that the catch block is for, that should be caught narrowly using the pattern in the suggestion here.

pgarrison (approved):
I have been able to build on this successfully in #806 and #809, and it's been useful for benchmarking my code, so I think it's a good starting point that we should move forward with.

Some of my feedback are things that I've addressed already in #806, so I think we can just move forward from there.

At a high level, though, I have a concern about large LLM-generated PRs. They are difficult to review due to their length, and code readability is hampered by verbosity, poor use of DRY, vague or over-abbreviated variable names, and inconsistent comment quality. If we want to maintain high readability standards and thorough understanding of the codebase, there is a risk that future PRs like this shift effort from PR authors to reviewers.

For these reasons, I have opted not to fully read and understand the details of all code here. Instead, I've focused my reading on looking for security concerns.

PR #741, opened 2026-04-16
Add provenance file option to data source prompt

aswallace:
## Context
Adds back the ability for users to load provenance files. There's a more significant redesign coming, so this is mostly a temporary measure

## Changes
In the data source prompt, adds a provenance file option that matches what we currently provide for metadata descriptor files, with minor styling changes that are more similar to what the eventual final design looks like. Made sure that users can add both types of optional files for the same data source

### To do:
We don't currently allow users to add a metadata descriptor or provenance file on its own without also adding a data source, which I imagine folks will want (e.g., if they're adding it to an existing query). Allowing this is going to rely on/interfere with some of the changes in #732, so to keep things clean, I'll wait on that & then make a separate ticket (#742).

## Testing
Manually tested creating new queries with one or both of the optional files. Made sure the provenance graph still gets generated and that descriptors do get added.

## Screenshots 
image

### DIFF
--- a/packages/core/components/DataSourcePrompt/index.tsx
+++ b/packages/core/components/DataSourcePrompt/index.tsx
@@ -43,24 +43,23 @@ export default function DataSourcePrompt(props: Props) {
 
     const [dataSource, setDataSource] = React.useState<Source>();
     const [metadataSource, setMetadataSource] = React.useState<Source>();
-    const [showAdvancedOptions, setShowAdvancedOptions] = React.useState(false);
+    const [provenanceSource, setProvenanceSource] = React.useState<Source>();
+    const [showMetadataPrompt, setShowMetadataPrompt] = React.useState(false);
+    const [showProvenancePrompt, setShowProvenancePrompt] = React.useState(false);
     const [isDataSourceDetailExpanded, setIsDataSourceDetailExpanded] = React.useState(false);
 
     const onDismiss = () => {
         dispatch(interaction.actions.hideVisibleModal());
     };
 
-    const onSubmit = (dataSource: Source, metadataSource?: Source) => {
-        if (sourceType === DataSourceType.provenance) {
-            if (dataSource) {
-                dispatch(selection.actions.changeProvenanceSource(dataSource));
-            }
-            // To do: include provenance source in query as with metadatasource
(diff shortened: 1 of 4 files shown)


packages/core/components/DataSourcePrompt/index.tsx
```diff
@@ -43,36 +43,40 @@ export default function DataSourcePrompt(props: Props) {
...
-    const onSubmit = (dataSource: Source, metadataSource?: Source) => {
-        if (sourceType === DataSourceType.provenance) {
-            if (dataSource) {
-                dispatch(selection.actions.changeProvenanceSource(dataSource));
-            }
-            // To do: include provenance source in query as with metadatasource
-            return onDismiss();
-        }
+    const onSubmit = () => {
+        if (!dataSource) return; // TO DO: Allow users to add just a prov/metadata descriptor files without adding a dataSource
```
SeanDuHare: Is this another ticket?

SeanDuHare (approved):

packages/core/components/DataSourcePrompt/index.tsx
```diff
@@ -133,21 +170,34 @@ export default function DataSourcePrompt(props: Props) {
...
+                        />
+                    )}
+                </div>
+                {showMetadataPrompt && metadataDescriptorPrompt}
+                {showProvenancePrompt && provenanceFilePrompt}
                 <div className={styles.loadButtonContainer}>
                     <PrimaryButton
                         className={classNames(styles.loadButton)}
-                        disabled={!dataSource && !metadataSource}
+                        disabled={!dataSource} // TO DO: Allow users to upload just metadata/prov files
```
SeanDuHare: (same comment here, link to the issue would be good)

packages/core/components/DataSourcePrompt/index.tsx
aswallace: Created ticket #742 and added to comments in [dc8add6](https://github.com/AllenInstitute/biofile-finder/pull/741/commits/dc8add61a2ac4839a76c457c5c327b1f564a3629)

BrianWhitneyAI (approved):

PR #748, opened 2026-04-22
fix: cast File Path to VARCHAR before TRIM and regex ops to handle BI…

### LINKED ISSUE #747: trim(BIGINT) error when loading CSV with numeric File Path values
### Bug
When loading a CSV data source where DuckDB infers the \"File Path\" column as `BIGINT` (e.g. all values are numeric IDs), the app throws:

> `Binder Error: No function matches the given name and argument types 'trim(BIGINT)'`

### Root Cause
`getRowsWhereColumnIsBlank` calls `TRIM(A."File Path")` directly. If DuckDB's `read_csv_auto` with `sample_size=-1` infers the column as `BIGINT`, this fails. `addRequiredColumns` then also runs `REGEXP_REPLACE` on \"File Path\" which would fail for the same reason.

### Fix
- Cast to `VARCHAR` before `TRIM` in `getRowsWhereColumnIsBlank`
- Add `ALTER COLUMN "File Path" TYPE VARCHAR` in `addRequiredColumns`

BrianWhitneyAI:
## Description 

The purpose of this PR is to resolve #747. CSV type inference may produce non-VARCHAR "File Path" (e.g. all-numeric values inferred as BIGINT). Coerce to VARCHAR so regex-based operations succeed.

### DIFF
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -494,6 +494,9 @@ export default abstract class DatabaseService {
                 ADD COLUMN ${HIDDEN_UID_ANNOTATION} INT
             `,
             this.getUpdateHiddenUIDSQL(name),
+            // CSV type inference may produce non-VARCHAR "File Path" (e.g. all-numeric values
+            // inferred as BIGINT). Coerce to VARCHAR so regex-based operations succeed.
+            `ALTER TABLE "${name}" ALTER COLUMN "${PreDefinedColumn.FILE_PATH}" TYPE VARCHAR`,
         ];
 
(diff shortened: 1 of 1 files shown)


aswallace (approved):
Since `addRequiredColumns` only gets used for non-parquet files, it may be worth checking if this will be an issue with parquets

BrianWhitneyAI:
> Since `addRequiredColumns` only gets used for non-parquet files, it may be worth checking if this will be an issue with parquets

Parquets have their datatypes already associated internally so we dont make any inferences like we do with csvs. I guess it could be a problem if a user has typed file_name as numeric. Not sure tbh

SeanDuHare (approved):

PR #775, opened 2026-05-04
Fuzzy search -> Is exactly or contains

### LINKED ISSUE #545: Replace "Fuzzy search" toggle with "Is Exactly / Contains" buttons
## Use Case

When filtering on a text field, the meaning of "Fuzzy search" may be unclear to those unfamiliar with the jargon.

## Suggest a Solution

Replace the "Fuzzy search" toggle with an "Is Exactly / Contains" toggle (I imagine the text is now on the clickable part of the input, instead of next to it)

SeanDuHare:
Fuzzy search is kind of a dev specific term, this updates the label to say "Is exactly or contains" instead.

Resolves #545 

Example
Screenshot 2026-05-04 at 2 07 16 PM

### DIFF
--- a/packages/core/components/AnnotationFilterForm/SearchBoxForm/SearchBoxForm.module.css
+++ b/packages/core/components/AnnotationFilterForm/SearchBoxForm/SearchBoxForm.module.css
@@ -1,54 +1,13 @@
-.toggle {
+.checkbox {
     margin-bottom: 6px;
 }
 
-.toggle-hidden {
+.checkbox > label > span {
+    font-size: 14px;
+}
+
+.checkbox-hidden {
     display: none;
-}
-
-.toggle input:disabled + label {
-    cursor: not-allowed;
-    opacity: 0.5;
-}
-
-.toggle :is(input, label, span), .toggle-label {
-    color: var(--primary-text-color);
-    font-size: var(--l-paragraph-size);
-    font-weight: 500;
-}
-
-.toggle label:hover > span {
-    color: var(--highlight-text-color) !important;
-}
-
-.toggle-pill-off {
-    border-color: var(--border-color) !important;
-    background-color: unset !important;
-}
-
-.toggle-pill-off {
-    border-color: var(--border-color) !important;
-}
-
-.toggle-pill-off > span {
-    background-color: var(--border-color) !important;
-}
-
-.toggle-pill-off:hover > span {
-    background-color: var(--primary-text-color) !important;
-}
-
-.toggle-pill-on {
-    border: none !important;
-    background-color: var(--aqua) !important;
-}
-
-.toggle-pill-on > span {
-    background-color: var(--highlight-hover-text-color) !important;
-}
-
-.toggle-pill-on:hover {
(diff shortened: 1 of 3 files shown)


aswallace:
I think this is a little confusing, it's unclear what the toggle does when phrased this way (does the user know whether "on" mean BOTH is exactly or contains, and if not, which one)

SeanDuHare:
> I think this is a little confusing, it's unclear what the toggle does when phrased this way (does the user know whether "on" mean BOTH is exactly or contains, and if not, which one)

Hmm good point. Perhaps this should be a checkbox with an updating label rather than a toggle, I'll run this by Lyndsay

BrianWhitneyAI:
This is maybe a different perspective but do we actually care about exact search? with a fuzzy search you already have the power to exact search by going down to the specific name since file name is unique correct? Could just only use fuzzy search

SeanDuHare:
> This is maybe a different perspective but do we actually care about exact search? with a fuzzy search you already have the power to exact search by going down to the specific name since file name is unique correct? Could just only use fuzzy search

Yea because if you want to just capture those from gene "TXY" and the search returns "TXY" and "TXYC" then you are getting misleading results also I think, but would need to double check, that fuzzy search is less performant

BrianWhitneyAI:
> Yea because if you want to just capture those from gene "TXY" and the search returns "TXY" and "TXYC" then you are getting misleading results also I think, but would need to double check, that fuzzy search is less performant

Could this be solved by then switching to list picker and selecting TXY?

SeanDuHare:
Talked with Lyndsay - seems like this is the way to go for now. This checkbox behavior and styling mirrors what users see in the group by pop up



Screenshot 2026-05-07 at 3 50 49 PM
Screenshot 2026-05-07 at 3 50 45 PM

SeanDuHare:
> > Yea because if you want to just capture those from gene "TXY" and the search returns "TXY" and "TXYC" then you are getting misleading results also I think, but would need to double check, that fuzzy search is less performant
> 
> Could this be solved by then switching to list picker and selecting TXY?

Yep! But this searchbox is used for lists of > 100 values & we wouldn't be able to efficiently load a list of thousands of values

BrianWhitneyAI (approved):
I still like the idea of default fuzzy searching and list picking over a toggle but seems fine to me

aswallace:
Looking at this again, I still have a slight gripe- with this checked it makes it look like we are _currently_ disabling fuzzy search. This might just be me though so I'm fine with going forward with it if others are on board
image

aswallace (approved):
Approving pending thoughts on most recent comment

SeanDuHare:
> Looking at this again, I still have a slight gripe- with this checked it makes it look like we are _currently_ disabling fuzzy search. This might just be me though so I'm fine with going forward with it if others are on board
> 
> image
> 
> 

i wonder about this too, but this is the same logical (negative affirmation) pattern that the other checkbox uses - so it's i guess becoming our convention... something to think about and asking users about. Going to merge with the plan of asking UX to talk to users or talk to them myself about this pattern

PR #782, opened 2026-05-10
Feature/show all columns at once

SeanDuHare:
## Context
For, pretty much ever, users have wanted to see all the columns in their dataset at once.

## Description
This, by default, adds all columns to the users view. As a result, we can remove the context menu option for modifying the columns. In turn, I've added the option to send the column to the start or end for ease which on top of the previous PR for drag and drop reordering columns should help users create datasets nicely.

**A big part of this changeset is converting the width tracking of columns from % of total width to px.**

## Testing
Tons of local testing. Making sure the column names get padded and cut off properly is truly a nightmare but seems to be good now.

## Follow up PRs
* #783 Add gradient to top and left of the file sets (they exist on the bottom and right right now)
* #783 Some little display improvements like pixel spacing that don't really have anything to do with this changeset
* #784 Auto calculating optimal column width based on the lenghtiest value for the column


packages/core/entity/SearchParams/index.ts
```diff
@@ -97,11 +102,24 @@ enum URLQueryArgShorthands {
...
-        return columns
-            .map((column) => `${column.name}${ColumnCoder.VALUE_DELIMETER}${column.width}`)
-            .join(ColumnCoder.COLUMN_DELIMETER);
+        return (
+            columns
+                // Encode width as divided by COLUMN_VALUE_PRECISION to shorten the resulting URL;
+                // this is an arbitrary choice to balance URL length with precision of column widths
+                .map(
+                    (column) =>
+                        `${column.name}${ColumnCoder.VALUE_DELIMETER}${Math.ceil(
```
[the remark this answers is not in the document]
SeanDuHare: Prefer it to grow due to being cut off by a factor of 10

packages/core/entity/SearchParams/index.ts
```diff
@@ -97,11 +102,24 @@ enum URLQueryArgShorthands {
...
+                            column.width / ColumnCoder.COLUMN_VALUE_PRECISION
+                        )}`
+                )
+                // Arbitrary limit to prevent URLs from getting too long;
+                // if users have more than 6 columns they can resize and reorder them in-app after loading the URL
+                .slice(0, 6)
```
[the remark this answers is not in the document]
SeanDuHare: This is to avoid columns getting too big

packages/core/entity/SearchParams/index.ts
```diff
@@ -110,7 +128,19 @@ class ColumnCoder {
...
+                // space the column should take up, but this was difficult to work with and unintuitive for users.
+                // Now we encode the actual pixel width, which is more straightforward to understand and work with when manually editing URLs.
+                // To maintain backwards compatibility with existing URLs, we continue to support previously encoded widths as percentages,
+                // but we convert them to pixel widths in the encoding process.
+                // Also, multiply the parsedWidth by COLUMN_VALUE_PRECISION because it is encoded as the actual width divided by COLUMN_VALUE_PRECISION to
+                // shorten the resulting URL; this is an arbitrary choice to balance URL length with precision of column widths.
+                const width =
+                    parsedWidth <= 1
+                        ? DEFAULT_COLUMN_WIDTH
+                        : parsedWidth * ColumnCoder.COLUMN_VALUE_PRECISION;
```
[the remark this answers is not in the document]
SeanDuHare: Fixed in 42ce41c54c7ea7b8fe58fc3e39e2d38d0b78fef9

packages/core/state/selection/reducer.ts
```diff
@@ -208,16 +209,40 @@ export default makeReducer<SelectionStateBranch>(
...
+                    // Check for matching column in special top level annotations like File Name
+                    // and if still no match just skip
+                    const matchingSpecialAnnotation = TOP_LEVEL_FILE_ANNOTATIONS.find(
+                        (a) => reorder.name === a.name || reorder.name === a.displayName
+                    );
+                    if (!matchingSpecialAnnotation) {
+                        continue;
+                    }
+                    moving = {
+                        name: matchingSpecialAnnotation.displayName,
```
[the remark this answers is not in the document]
SeanDuHare: True! Fixed in 9101e9e5297363d3c76d5eecac38abadd34f115e

packages/core/components/FileList/index.tsx
```diff
@@ -83,15 +84,68 @@ export default function FileList(props: FileListProps) {
...
+    // The key change unmounts the scrollable container, creating a new one with scrollLeft=0.
+    // lastScrollLeftRef still holds the previous position since FileList itself isn't remounted.
+    React.useLayoutEffect(() => {
+        const el = outerRef.current;
+        if (!el) return;
+        const target = lastScrollLeftRef.current;
+        if (target > 0 && Math.abs(el.scrollLeft - target) > 1) {
+            el.scrollLeft = target;
+        }
+    });
```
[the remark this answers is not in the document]
SeanDuHare: I avoided this because I'm uncertain I prefer the risk of it not updating, but I suppose by tracking the fileSet.instanceId and fileView it'll be fine. Fixed in bd82f8deceb410d00e67201e886dab6443ecb5c7

packages/core/components/FileList/index.tsx
```diff
@@ -83,15 +84,68 @@ export default function FileList(props: FileListProps) {
...
+            rafIdRef.current = requestAnimationFrame(() => {
+                setHorizontalScroll({ scrollLeft: el.scrollLeft, containerWidth: el.clientWidth });
+            });
+        };
+        el.addEventListener("scroll", onScroll, { passive: true });
+        return () => {
+            el.removeEventListener("scroll", onScroll);
+            cancelAnimationFrame(rafIdRef.current);
+        };
+    }, [totalCount]); // re-attach when totalCount changes (outerRef may remount)
```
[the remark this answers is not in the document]
SeanDuHare: This was a nice catch by the AI. Fixed in 60da447c83a6f0465ff0258e8e5928ccb112bef7

packages/core/components/FileList/index.tsx
```diff
@@ -83,15 +84,71 @@ export default function FileList(props: FileListProps) {
...
+    // lastScrollLeftRef still holds the previous position since FileList itself isn't remounted.
+    React.useLayoutEffect(() => {
+        const el = outerRef.current;
+        if (!el) return;
+        const target = lastScrollLeftRef.current;
+        if (target > 0 && Math.abs(el.scrollLeft - target) > 1) {
+            el.scrollLeft = target;
+        }
+    }, [fileSet.instanceId, fileView]);
+
```
BrianWhitneyAI, line 148: This restores scroll position after the InfiniteLoader remounts (e.g. after a sort), but it fires on any fileSet.instanceId change — including when the user switches to a completely different dataset. If the user was scrolled to 500px in dataset A and then opens dataset B, this will scroll B to 500px immediately. Consider resetting lastScrollLeftRef.current to 0 when the instanceId changes to distinguish "remount due to sort" from "switched to new dataset."

packages/core/components/FileList/index.tsx
```diff
@@ -83,15 +84,71 @@ export default function FileList(props: FileListProps) {
...
+        };
+        el.addEventListener("scroll", onScroll, { passive: true });
+        return () => {
+            el.removeEventListener("scroll", onScroll);
+            cancelAnimationFrame(rafIdRef.current);
+        };
+        // Re-attach listener when totalCount changes, as react-window may remount the outerRef div in that case
+        // but also when the fileView changes so that the scroll position can be reset
+        // and when measuredWidth changes to ensure containerWidth is accurate for useVisibleCells (like if window size changes)
+    }, [fileSet.instanceId, fileView, measuredWidth, totalCount]);
```
BrianWhitneyAI: Per Claude Code - The totalCount dep here looks incorrect. The comment says react-window may remount outerRef when totalCount changes, but FixedSizeList is keyed on fileSet.instanceId and fileView — it remounts on those, not on count changes. Including totalCount means the scroll listener is torn down and containerWidth is re-initialized on every page of data that loads in, which briefly collapses the rendered columns to the first-6 fallback. I think you can drop totalCount from this dep array.

packages/core/entity/SearchParams/index.ts
```diff
@@ -60,10 +65,10 @@ export const PAST_YEAR_FILTER = new FileFilter(
...
 export const DEFAULT_AICS_FMS_QUERY: SearchParamsComponents = {
     columns: [
-        { name: AnnotationName.FILE_NAME, width: 0.4 },
-        { name: AnnotationName.KIND, width: 0.2 },
-        { name: AnnotationName.TYPE, width: 0.25 },
-        { name: AnnotationName.FILE_SIZE, width: 0.15 },
+        { name: AnnotationName.FILE_NAME, width: 300 },
+        { name: AnnotationName.KIND, width: 150 },
+        { name: AnnotationName.TYPE, width: 200 },
+        { name: AnnotationName.FILE_SIZE, width: 100 },
```
BrianWhitneyAI, line 71: I might be missing something here but whats the reasoning for switching from percentages/ proportions to pixels?

BrianWhitneyAI (approved):
So if I am understanding correctly, This removes any column picking in favor of ordering the columns you want to the front. Is that what we want to do rather than just switch to "all columns" option for users? Should we be considering a way to hide columns?

packages/core/entity/SearchParams/index.ts
SeanDuHare: Couple reasons:
a) Proportions got funky when switching between users with different window sizes
b) Proportions relative to an essentially infinite list gets tricky and relies on reaching outside of the component and into the parent to measure the width which felt icky - though it is possible!

aswallace:
I'm conflicted about fully removing the columns menu, it seems like it could be frustrating to only be able to sort/manage columns this way. In particular, if you have hundreds of columns (e.g., the NucMorph dataset, which has 658), you now have to scroll through to find the ones you want since there's no longer a search function. I can also imagine that users who are using this for publishing data might want to have the ability to just show a subset of columns?

SeanDuHare:
> I'm conflicted about fully removing the columns menu, it seems like it could be frustrating to only be able to sort/manage columns this way. In particular, if you have hundreds of columns (e.g., the NucMorph dataset, which has 658), you now have to scroll through to find the ones you want since there's no longer a search function. I can also imagine that users who are using this for publishing data might want to have the ability to just show a subset of columns?


#### Re: searching for columns

I hadn't really considered this! Nobody seemed to bring it up in my demos, but this is a great point. I'm not sure what is the best thing to do here to solve this, the existing "Modify columns" action + column picker UI has never not been confusing to users I've seen using BFF so I'm not sure that is the answer either.


#### Re: hiding columns

I wonder about whether users will want to hide columns as well. None of the datasets that have been published alongside BFF seems to have too many columns & I talked with Antoine about this for the EMT dataset and he wasn't concerned - _but_ it does seem potentially tricky. The reason I didn't include some way to only show a subset in this is because I think showing all columns is better than showing a small random subset by default which seem to be our only two options. Otherwise we get into issues with trying to store too much information in the query arguments, like for example:
* If we store the columns they **want** to show then we get up to the maximum number of columns (ish) in the query args
* If we store the columns they **do not want** to show then we would be storing up to the maximum number of columns (ish)

I can't think of any optimizations we could do that would get better storage performance than up to X / 2 columns in the query args in the worst case (X = number of columns).

My thought was try publishing this and then if we want to add on the ability to hide columns we could add it in. FWIW the way I imagine we could go about storing columns as query args is:
* Use their names not indexes (index-based will get us into issues when the source is edited or datasets are combined)
* Consider an empty list of columns to mean "show all"
* If columns shown is > half of max, store the hidden columns in the arguments (as hidden). Otherwise, store the shown columns in the arguments (as shown)
* Omit widths after X amount of columns - just as a way to reduce character count. Let them be auto-calculated
* Use lazy .zip like compression or similar (if it reduces character count)

Let me know your thoughts!

(fwiw NucMorph is a draft of a draft of a draft of a dataset where most of that metadata is irrelevant, but I totally see your point and is is very true/fair)

aswallace:
**re: searching columns**
Is the order of the columns rn just whatever the user supplied? I wonder if having them sorted in some way would make it easier to find things (but immediate counter argument would be that the user might have provided them in that order for a reason and wouldn't want them re-ordered). Otherwise, my instinct would be to keep the menu available in case someone has been relying on that functionality, but I'm not sure if it would be too logistically complicated to maintain both options

**re: hiding columns**
That's a great point @ query arguments and I totally agree that if we're going to store that info in the args it would need to be limited subset. 
Definitely agree on empty list = show all and on auto-calculating widths
Would "X" need to have a fixed upper limit? since we allow data sources to have any number of columns

I'm on board with moving forward with this as is for now and thinking about hiding columns as a separate issue. Will finish reviewing!

packages/core/components/FileList/Header.tsx
```diff
@@ -52,22 +48,47 @@ function Header(
...
+    const onHeaderNameClick = (evt: React.MouseEvent, columnName: string) => {
         // Prevent this click from bubbling up to the header's onClick
         // which opens the column picker context menu
         evt.stopPropagation();
-        dispatch(selection.actions.sortColumn(column.name));
+        dispatch(selection.actions.sortColumn(columnName));
     };
 
-    const headerCells: CellConfig[] = map(columns, (column) => ({
+    const onHeaderColumnClick = (evt: React.MouseEvent, columnName: string) => {
```
aswallace, line 64: What do you think of combining these so that sort is an option in the menu? When the columns are more narrow it's hard to click the header without widening the column, and I'm not sure if users would know to distinguish between those click targets. I know they were separate before bc one was acting on the column itself and the other was acting on the header as a whole, but now the menu is also column-specific

https://github.com/user-attachments/assets/560eedd0-72cb-42f9-be2f-c5ded0f08150

packages/core/entity/SearchParams/index.ts
```diff
@@ -97,11 +102,24 @@ enum URLQueryArgShorthands {
 class ColumnCoder {
     private static readonly COLUMN_DELIMETER = ",";
     private static readonly VALUE_DELIMETER = ":";
```
aswallace: technically unrelated to this PR, but minor typo in these names, should be `COLUMN_DELIMITER` and `VALUE_DELIMITER` as opposed to delimeter

packages/core/components/FileList/index.tsx
aswallace: I may be misunderstanding, but I don't think I've been able to replicate this comment

aswallace (approved):

packages/core/components/FileList/Header.tsx
SeanDuHare: I think that makes sense! I'll run it by UX and report back if they disagree, otherwise will implement

packages/core/entity/SearchParams/index.ts
SeanDuHare: classic - i can update this

packages/core/components/FileList/Header.tsx
SeanDuHare: Decided against sorting joining the menu (for now) but the move to start / move to end were added to the menu in 728c0705abc02f0efcb06effdc4dea34cafdc815

packages/core/entity/SearchParams/index.ts
SeanDuHare: Done in a4b5a4259ea4acf4649869db1af0cbde28e11593

packages/core/components/FileList/index.tsx
SeanDuHare: Same here :( lmk if you can see it, but I'll move forward for now

packages/core/components/FileList/index.tsx
SeanDuHare: I couldnt create a bug either way so I just removed it since I could no longer recreate the bug I had originally commented anyway 🤷  (see 17ea6d9ca79d8bee2b0b9c86880945292a6c65a7)

PR #786, opened 2026-05-12
Bugfix/non-ascii character display

### LINKED ISSUE #780: Fix non-ASCII character display
Suraj: Hello!! I was wondering if it is possible to allow unicode characters in metadata keys on BFF. The cell height column in attached picture should ideally be "Cell Height (µm)", and the cell volume should be "Cell Volume (µm³)" (these were sent with the csv). Thank you! (edited)

aswallace:
## Context
When addressing #700, we ran into files that had non-ASCII characters in column names that broke our grouping/filtering because the Javascript conversion was causing those columns to no longer match the database. As a blanket fix, I added renaming logic to remove all non-ASCII characters from column names during the data source ingestion process. This is now causing issues for users who include units like `µm³` in their metadata field names. 

Resolves #780 

## Changes
Made the logic for renaming columns with non-ASCII characters more conservative and less of a sledgehammer. Instead of always removing those characters, this checks whether Javascript actually trims them (or leaves them, as is the case with `µm³`), and then only renames columns if the Javascript version no longer matches what's in the DuckDB database.

Potential to do:
There are probably ways to make this smarter? e.g., maybe we don't ever want to remove them, or should just convert them to their characters codes?

## Testing
Tested manually on three cases:
1. Nucmorph dataset: this is the file that has a column named `ÔªøPrefix` that causes filtering and aggregate sources to fail. This is a case where the column should be renamed
2. Saurabh's dataset that uses `µm³` etc. These don't get renamed
3. A dataset where none of the columns have any non-ASCII characters

### DIFF
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -657,11 +657,11 @@ export default abstract class DatabaseService {
             (val) => `${val.column_name}`
         );
         // Determine whether Javascript removes those characters by checking
-        // if the JS version of each column name still exists in the table
+        // if the JS version of each column name still exists in the DuckDB table
         const columnsExist = await this.checkColumnsExist(dataSourceName, columnsWithNonAscii);
         const columnsWithMismatch = columnsExist
             ?.filter((value) => value.column_exists === false)
-            .flatMap((val) => `${val.column_name}`);
+            .flatMap((val) => val.column_name);
         // If list is empty, there may be column names with non-ASCII characters,
         // but they don't get trimmed by JS, so we can return safely
         if (!columnsWithMismatch || columnsWithMismatch.length === 0) return;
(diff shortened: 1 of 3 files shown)


packages/core/services/DatabaseService/index.ts
```diff
@@ -648,18 +648,51 @@ export default abstract class DatabaseService {
...
+            ?.filter((value) => value.column_exists === false)
+            .flatMap((val) => val.column_name);
+        // If list is empty, there may be column names with non-ASCII characters,
+        // but they don't get trimmed by JS, so we can return safely
+        if (!columnsWithMismatch || columnsWithMismatch.length === 0) return;
+
+        // Only rename the columns that have a mismatch after being converted to JS
+        // since these are the ones that won't render correctly & will cause errors
+        const findMismatchedColumnsSql = `
+            WITH nonascii_columns AS (
```
BrianWhitneyAI, line 681: Will a quotation break this?

BrianWhitneyAI (approved):

kmitcham (approved):

packages/core/services/DatabaseService/index.ts
aswallace: good question, will take a look

aswallace: Added escaping single quotes in [00d8674](https://github.com/AllenInstitute/biofile-finder/pull/786/commits/00d8674a48ed79f964b3774bb7a42845eb7298f3), preventing double quotes in [d8c3a3b](https://github.com/AllenInstitute/biofile-finder/pull/786/commits/d8c3a3bfb5acfb1933545d16bc1ae2e7c0bea9fc)

SeanDuHare (approved):

Issue #638, opened 2026-01-12
Direct From Parquet mode


pgarrison:
Goal: read large parquet files (e.g., 85 columns, 30M rows) in the browser.

Proof of concept branch: demo-large-parquet

Disable metadata editing for parquet files. Skip the `CREATE TABLE` and `ALTER TABLE` steps for them. If the user tries to edit metadata, make a popup with language like “Sorry, metadata editing isn’t available for parquet files right now, while we scale BioFile Finder up to support larger parquet files. Future BioFile Finder work will restore this feature.”

# Details
In the proof of concept branch, the user often has to click the "Load" button twice when adding a data source. That is a bug, and resolving it should be part of this ticket.

PR #787, opened 2026-05-12
Support aggregate parquet data sources

pgarrison:
## Context
In #638 we dropped support for aggregating parquet files temporarily in order to ship improved parquet read performance. This PR brings it back.

## Changes
Aggregate parquet sources are implemented using duckdb's internal parquet aggregation feature, passing an array to `read_parquet`.
- This option should allow duckdb full opportunity to optimize by pushing queries into the reader.
- The column renaming semantics aren't as nice as the `CREATE TABLE` implementation for CSVs. Column names are matched exactly: inconsistent casing across inputs creates split columns.

## Performance
I evaluated performance of the aggregate tables by using the benchmarking harness under review in #739.

I conducted 80 trials on each of three cases: a 10M row test file, a 20M row test file, and an aggregate source composed of two copies of the same 10M test file.
* To ensure duckdb doesn't recognize the two copies of the input file as the same and just download one of them, I uploaded a copy of the input file to the test bucket. In the github actions test configuration, the file is loaded from `http://localhost:18765/fixtures/synthetic-10m.parquet` and `http://localhost:18765/fixtures/synthetic-10m-copy.parquet`.
* I ran these trials in 4 batches on two github actions runners, each with the following structure.
  * For each input file case, start a new chromium instance
    * 1 warmup round
    * 20 trials
* The input parquets have a row group size of 122,880
* The 20M test file is just the 10M file, concatenated with itself.
  * Its row group layout looks like: Group 1, Group 2, ..., Group 82, Group 1, Group 2, ..., Group 82.

Naively, we might expect most queries to take about twice as long on the 20M row table or 10M+10M aggregate, compared to the 10M row table. For example, fast queries that just need row group metadata might have to look at metadata from twice as many row groups, and queries that need to scan the full table need to load twice as much data. 

Comparing 10M+10M to 10M, the results indicate that most queries on the 10M+10M aggregate come with moderate performance overhead, slower than linear scaling would suggest.

Comparing 10M+10M to 20M, the 20M file shows consistent, and sometimes substantial, performance improvements. 

Screenshot 2026-05-18 at 10 13 43 AM

Aside from the surprising large sort result, the greatest relative difference is in the "expand group" operation, where the 10M+10M aggregate takes on average 6.0s to the 20M row file's 2.6s. This is a `SELECT DISTINCT ... WHERE` query, so it does require a full table scan. Examining results for this query alone show low noise relative to the magnitude of the performance difference.

Screenshot 2026-05-18 at 10 13 55 AM

### Thoughts for future performance improvements
* The sort performance is surprising. Sorting a 20M row table should take at least as much time as sorting a 10M row table. See more discussion in #802


packages/core/services/DatabaseService/index.ts
```diff
@@ -784,12 +794,16 @@ export default abstract class DatabaseService {
             selectParts.push(fileNameSelectPart);
         }
         selectParts.push(`"file_row_number" AS "${HIDDEN_UID_ANNOTATION}"`);
+        if (sourceNames.length > 1) {
+            selectParts.push(`"filename" AS "${DATA_SOURCE_COLUMN}"`);
+        }
         // 4. Create the view for this data source
-        const createViewSql = `CREATE VIEW "${name}"
+        const quotedNames = sourceNames.map((name) => `'${name}'`).join(", ");
```
BrianWhitneyAI, line 801: this will break of theres a quote in the name. Not sure if there ever would be ?

packages/core/services/DatabaseService/index.ts
```diff
@@ -784,12 +794,16 @@ export default abstract class DatabaseService {
...
+        const createViewSql = `CREATE VIEW "${aggregateName}"
             AS SELECT ${selectParts.join(", ")}
-            FROM parquet_scan('${name}');`;
+            FROM parquet_scan(ARRAY[${quotedNames}], union_by_name = true);`;
         await this.execute(createViewSql);
-        this.parquetDirectViewNames.add(name);
+        this.parquetDirectViewNames.add(aggregateName); // TODO change this for aggregates?
```
BrianWhitneyAI: Is this for future or for this PR? confused what its changing

packages/core/services/DatabaseService/index.ts
```diff
@@ -936,97 +947,114 @@ export default abstract class DatabaseService {
...
+                    viewName,
+                    dataSources.map((source) => source.name)
                 );
+            } else {
+                await this.createAggregateInMemory(viewName, dataSources);
+            }
+        } catch (err) {
+            const formattedError =
+                (err as Error)?.message || `Error while combining data sources. ${err}`;
+            throw new DataSourcePreparationError(formattedError, viewName);
```
BrianWhitneyAI, line 972: This rewraps the `DataSourcePreparationError`

BrianWhitneyAI (approved):

packages/core/services/DatabaseService/index.ts
pgarrison: Whoops, just didn't mean to push the TODO

packages/core/services/DatabaseService/index.ts
pgarrison: Yes, it will. Unfortunately, so will the BFF code for handling single sources, or CSV aggregates. The only-partially-sanitized user input is a potential security risk that may merit further investigation (probably off our public github, in case we find a vulnerability).

packages/core/services/DatabaseService/index.ts
pgarrison: Resolved [ae75299](https://github.com/AllenInstitute/biofile-finder/pull/787/commits/ae75299e98afb3518610fa91cac49fe53ebd57b1)

packages/core/services/DatabaseService/index.ts
aswallace: May or may not be useful in this particular case, but in other places we're escaping single quotes (in duckdb that just looks like regex replacing `'` with two single quotes)

aswallace (approved):

SeanDuHare (approved):

packages/core/services/DatabaseService/index.ts
pgarrison: Resolving as out of scope for this PR: will follow-up separately

PR #793, opened 2026-05-15
ACM REVIEW #2: Add User field to All Cells Mask modal

BrianWhitneyAI:
## Summary
- Adds a required **User** field to the All Cells Mask modal, matching the pattern used in the discoverable-pipelines branch (userId state, inline validation, sent as `X-User-Id` header via `setUserName`)

## Test plan
- [x] Open the All Cells Mask modal with locally cached files selected
- [x] Verify the User field is present and required (submit blocked with error if empty)
- [x] Verify a valid user ID allows submission and the job is submitted successfully
- [x] Verify the `X-User-Id` header is sent with the request

### DIFF
--- a/packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
+++ b/packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
@@ -42,6 +42,8 @@ export default function AllCellsMaskSegmentation({ onDismiss }: ModalProps) {
     const [computeTaskId, setComputeTaskId] = React.useState<string | null>(null);
     const [statusMessage, setStatusMessage] = React.useState<string | null>(null);
     const [localFilePaths, setLocalFilePaths] = React.useState<string[] | null>(null);
+    const [userId, setUserId] = React.useState<string>("");
+    const [userIdError, setUserIdError] = React.useState<string>("");
 
     React.useEffect(() => {
         let cancelled = false;
@@ -78,6 +80,12 @@ export default function AllCellsMaskSegmentation({ onDismiss }: ModalProps) {
     const onSubmit = async () => {
         if (status !== "idle" || !localFilePaths?.length) return;
 
+        const userErr = userId.trim() === "" ? "This field is required." : "";
+        if (userErr) {
+            setUserIdError(userErr);
+            return;
+        }
+
         setStatus("submitting");
         setStatusMessage(null);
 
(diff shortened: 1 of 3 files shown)


BrianWhitneyAI:
Screenshot 2026-05-15 at 10 26 18 AM

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -228,29 +228,27 @@ export default class HttpFileService extends HttpServiceBase implements FileServ
...
-                    channel: params.channel,
-                })
-            );
+        this.setUserName(params.user);
 
-            return {
-                computeTaskId: result.computeTaskId,
-                dashboardUrl: result.dashboardUrl,
-            };
-        } catch (error) {
```
BrianWhitneyAI, line 251: No explicit error handling needed here — rawPost extracts server error messages from axios responses and throws them as plain Errors, which the modal catch block handles.

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
```diff
@@ -78,6 +80,12 @@ export default function AllCellsMaskSegmentation({ onDismiss }: ModalProps) {
     const onSubmit = async () => {
         if (status !== "idle" || !localFilePaths?.length) return;
 
+        const userErr = userId.trim() === "" ? "This field is required." : "";
```
SeanDuHare, line 83: Does this user need to exist in AD or anywhere or is it just for human eyes to see?

SeanDuHare:
Screenshot 2026-05-15 at 10 59 03 AM
Full screen on my computer causes the placeholder to still get cut off - generally placeholders should be small enough to be seen on most window sizes

also the submit button's width changes based on whether it is disabled or not which causes some funky behavior

SeanDuHare (approved):
If the user does need to exist in AD - it is worth checking if there is something we could ping (anywhere) to verify that the user entered exists and adding a validation effect here

aswallace (approved):

packages/core/components/Modal/AllCellsMaskSegmentation/index.tsx
BrianWhitneyAI: just for human eyes to see and for searching in the dashboard. Eventually we will have an auth token for checking creds, not sure exactly how AD user will play in there

BrianWhitneyAI:
> If the user does need to exist in AD - it is worth checking if there is something we could ping (anywhere) to verify that the user entered exists and adding a validation effect here

opening an issue for it!

PR #799, opened 2026-05-15
ACM REVIEW #4: feat: add ComputePipelineModal and useComputePipelineMenuItems

BrianWhitneyAI:
### CORE
- Adds `ComputePipelineModal` — generic modal supporting dynamic pipeline selection, cluster selection, user ID, optional parameters (via Advanced options panel), and job submission
- Adds `useComputePipelineMenuItems` — hook that fetches available pipelines from `PipelineService` and builds a context menu submenu

## Minor
- `BaseModal` gains a `className` prop for wider modal sizing
- `ComboBox` CSS: adds `width: 100%` and fixes placeholder selector scoping

The `AllCellsMaskSegmentation` modal is still present and reachable; this PR is purely additive.

## Test plan
- [x] Open context menu on selected files → Process Files → Run Compute Pipeline → select a pipeline
- [x] Verify modal opens with correct pipeline pre-selected and parameters rendered
- [x] Verify User field is required — submit blocked with inline error if empty
- [x] Verify Run Location dropdown, required params, and advanced options all function
- [x] Expand selected files list and verify file names appear
- [x] Submit an All-Cells-Mask job and verify success banner with job ID
- [x] Verify Run Compute Pipeline submenu is hidden for non-AICS employees
- [x] Verify correct endpoint is hit in staging vs production environments

## How to Review

The core file to look at is [packages/core/components/Modal/ComputePipelineModal/index.tsx](https://github.com/AllenInstitute/biofile-finder/pull/799/files/a4f5360b24fa0e660c54e9af4b066c78dd116170#diff-ddb7f68c13360bdedeaabc3beec3af6b1f194d395c4867b47c9eb9fe8fbdbed5). 


packages/core/components/ComboBox/ComboBox.module.css
```diff
@@ -16,7 +17,7 @@
   color: var(--highlight-text-color) !important;
 }
 
-.combo-box > *::placeholder {
+.combo-box input::placeholder {
   color: var(--secondary-text-color) !important;
   font-style: italic;
 }
```
BrianWhitneyAI, line 23: - width: 100% makes the combo boxes fill their container in the modal layout
  - The placeholder selector fix (> * → input) scopes the style correctly to the input element

packages/core/components/Modal/BaseModal/index.tsx
```diff
@@ -18,14 +20,14 @@ interface BaseModalProps {
...
     const titleId = "base-modal-title";
     return (
         <Modal
             isOpen
             onDismiss={onDismiss}
-            containerClassName={styles.container}
+            containerClassName={classNames(styles.container, className)}
             titleAriaId={titleId}
             overlay={{ className: styles.overlay }}
         >
```
BrianWhitneyAI, line 33: It lets individual modals override the width of the container. Maybe we still want a fixed container but I found it pretty restricting for modal content. Open to debate on this

packages/core/components/Modal/ComputePipelineModal/ComputePipelineModal.module.css
```diff
@@ -0,0 +1,371 @@
+/* ===================== MODAL SIZING ===================== */
```
BrianWhitneyAI, line 1: This is the CSS I added to match the mockup I made in image gen. We dont really need to follow it/ could defer to UX design after its released.

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
+import { IComboBoxOption, Icon, Spinner, TextField } from "@fluentui/react";
```
BrianWhitneyAI, line 1: This is the core file to review

packages/core/hooks/useComputePipelineMenuItems.ts
```diff
@@ -0,0 +1,70 @@
+import { IContextualMenuItem } from "@fluentui/react";
+import * as React from "react";
+import { useDispatch, useSelector } from "react-redux";
+
+import { ModalType } from "../components/Modal";
+import FileFilter from "../entity/FileFilter";
+import { interaction } from "../state";
+
```
BrianWhitneyAI, line 9: This is the "Discovery" portion

BrianWhitneyAI:
https://github.com/user-attachments/assets/5225e67c-8b10-4ae7-9262-03b39450e873

packages/core/components/Modal/ComputePipelineModal/ComputePipelineModal.module.css
SeanDuHare: I'd prefer this to be styled and change it post UX review rather than unstyled and wait for UX

packages/core/components/Modal/ComputePipelineModal/ComputePipelineModal.module.css
```diff
@@ -0,0 +1,371 @@
...
+    padding: 14px 16px;
+    border-radius: 8px;
+    border: 1px solid var(--border-color, rgba(255, 255, 255, 0.12));
+    background: var(--secondary-background-color);
+    cursor: pointer;
+    transition: border-color 0.15s;
+}
+
+.pipeline-card:hover {
+    border-color: var(--primary-color, #00b0d0);
```
SeanDuHare: I dislike fallback parameters in CSS like this since they can be hard to traceback IMO

(this is in a few places)

packages/core/components/Modal/ComputePipelineModal/ComputePipelineModal.module.css
```diff
@@ -0,0 +1,371 @@
...
+    font-size: var(--s-paragraph-size);
+    color: var(--secondary-text-color);
+}
+
+.param-error {
+    font-size: var(--s-paragraph-size);
+    color: var(--error-color, #d13438);
+}
+
+.input-box :global(input.ms-TextField-field) {
```
SeanDuHare: Global selections like this can get messy quickly - better to find another way to style this child

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
...
+import * as React from "react";
+import { useDispatch, useSelector } from "react-redux";
+
+import { ModalProps } from "..";
+import BaseModal from "../BaseModal";
+import { PrimaryButton, SecondaryButton } from "../../Buttons";
+import BaseComboBox from "../../ComboBox";
+import FileSelection from "../../../entity/FileSelection";
+import { Pipeline, PipelineParameter } from "../../../entity/ComputePipeline";
+
```
SeanDuHare: unnecessary line break (order is off too)

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
...
+        </div>
+    );
+}
+
+interface SelectedFilesExpanderProps {
+    count: number;
+    onExpand: () => Promise<string[]>;
+}
+
+function SelectedFilesExpander({ count, onExpand }: SelectedFilesExpanderProps) {
```
SeanDuHare: These components should be split into sibling files of the `ComputePipelineModal`

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
...
+import BaseComboBox from "../../ComboBox";
+import FileSelection from "../../../entity/FileSelection";
+import { Pipeline, PipelineParameter } from "../../../entity/ComputePipeline";
+
+import { interaction, selection } from "../../../state";
+import AnnotationName from "../../../entity/Annotation/AnnotationName";
+
+import styles from "./ComputePipelineModal.module.css";
+
+type ModalPhase = "loading" | "selecting" | "configuring" | "submitting" | "submitted" | "error";
```
SeanDuHare, line 21: Is `loading` different than `submitting`?

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
...
+
+    // Phase state
+    const [phase, setPhase] = React.useState<ModalPhase>("loading");
+    const [pipelines, setPipelines] = React.useState<Pipeline[]>([]);
+    const [selectedPipeline, setSelectedPipeline] = React.useState<Pipeline | null>(null);
+    const [selectedCluster, setSelectedCluster] = React.useState<string | null>(null);
+    const [parameters, setParameters] = React.useState<PipelineParameter[]>([]);
+
+    // File state (count only — paths are fetched at submit time)
+    const fileCount = fileSelection.count();
```
SeanDuHare, line 60: Does this create multiple `count` network calls? The count should be cached but I'm curious if in between it being called the first time and the result coming in if this function is rendering and causing duplicate network calls

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
...
+
+    // Form state
+    const [requiredValues, setRequiredValues] = React.useState<Record<string, unknown>>({});
+    const [fieldErrors, setFieldErrors] = React.useState<Record<string, string>>({});
+
+    // Optional params: add-one-at-a-time
```
SeanDuHare, line 66: what does "add-one-at-a-time" mean here, are these each intended to only have a state value when the others don't?

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
...
+            .catch(() => {
+                if (!cancelled) {
+                    setErrorMessage("Failed to load available pipelines.");
+                    setPhase("error");
+                }
+            });
+        return () => {
+            cancelled = true;
+        };
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
SeanDuHare, line 107: I assume this is about the react state hooks? Is it necessary to avoid putting them into the dep array?

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
...
+            try {
+                const params = await pipelineService.getParameters(
+                    pipeline.id,
+                    defaultCluster ?? ""
+                );
+
+                setParameters(params);
+                const defaults: Record<string, unknown> = {};
+                for (const p of params) {
+                    if (p.required && p.type !== "file_paths" && p.default !== null) {
```
SeanDuHare: `file_paths` as a type is used a few places, should be a constant or enum

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
...
+                        <div className={styles.successJobId}>Job ID: {computeTaskId}</div>
+                    )}
+                    <div className={styles.successHint}>
+                        You can monitor progress in the job dashboard.
+                    </div>
+                </div>
+            );
+        }
+
+        if (!selectedPipeline) return null;
```
SeanDuHare: This could be moved above line 437 and the boolean check for it on 437 could be removed

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,651 @@
...
+                {loading ? <Spinner /> : <Icon iconName={expanded ? "ChevronUp" : "ChevronDown"} />}
+            </div>
+            {expanded && paths !== null && (
+                <div className={styles.expanderBody}>
+                    <div className={styles.expanderNote}>
+                        Files are added automatically from your current selection.
+                    </div>
+                    <ul className={styles.filePathList}>
+                        {paths.map((p) => (
+                            <li key={p} className={styles.filePathItem}>
```
SeanDuHare: This would cause too many elements to be created for large selections. Either virtualize or render a slice

packages/core/hooks/useProcessFilesMenuItems.ts
SeanDuHare: Can this be deleted?

packages/core/hooks/useComputePipelineMenuItems.ts
```diff
@@ -0,0 +1,70 @@
...
+ * the "Run Compute Pipeline" context menu entry. Returns an empty array while
+ * loading and a single disabled item on error.
+ */
+export default function useComputePipelineMenuItems(
+    folderFilters?: FileFilter[]
+): IContextualMenuItem[] {
+    const dispatch = useDispatch();
+    const pipelineService = useSelector(interaction.selectors.getPipelineService);
+
+    const [items, setItems] = React.useState<IContextualMenuItem[]>([
```
SeanDuHare: Haven't testing this yet, does the menu re-render when the loading state is done or do you have to open the context menu again to see the pipelines?

packages/core/hooks/useComputePipelineMenuItems.ts
```diff
@@ -0,0 +1,70 @@
...
+                            text: "Failed to load pipelines",
+                            disabled: true,
+                        },
+                    ]);
+                }
+            });
+        return () => {
+            cancelled = true;
+        };
+    }, [dispatch, folderFilters, pipelineService]);
```
SeanDuHare: This makes it so we request the pipelines whenever we get new folder filters, however we only have this trigger when the folder filters change because we want to supply it as a parameter to the dispatch. It would probably be safer to move this logic elsewhere, either into the init of the app or split into two effects or something to avoid having the network call ever duplicate.

packages/core/hooks/useProcessFilesMenuItems.ts
BrianWhitneyAI: next PR!

packages/core/components/Modal/ComputePipelineModal/ComputePipelineModal.module.css
BrianWhitneyAI: [f8c902c](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/f8c902c94ca0d6a6db1db74f1a3603672c188b88)

packages/core/components/Modal/ComputePipelineModal/ComputePipelineModal.module.css
BrianWhitneyAI: [f8c902c](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/f8c902c94ca0d6a6db1db74f1a3603672c188b88)

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,661 @@
...
+    // Submission state
+    const [errorMessage, setErrorMessage] = React.useState<string | null>(null);
+    const [computeTaskId, setComputeTaskId] = React.useState<string | null>(null);
+    const [dashboardUrl, setDashboardUrl] = React.useState<string | null>(null);
+
+    // Load pipelines on mount; auto-select if a pipeline ID was passed via context menu
+    React.useEffect(() => {
+        let cancelled = false;
+        pipelineService
+            .getPipelines()
```
aswallace, line 85: We'll also have fetched the pipelines already just before opening the modal, does the service cache the result or will it be running a network call every time?

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: [b9601ac](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/b9601ac8178db1e470b863d55a1f6b733423d037)

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: [b9601ac](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/b9601ac8178db1e470b863d55a1f6b733423d037)

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: [b9601ac](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/b9601ac8178db1e470b863d55a1f6b733423d037)

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: yeah loading is loading the modal / ingesting pipeline info. submitting is the post requestr

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: so the way adding a param works is you select if from the drop down then add it with the add button then you can change the values. thats what this is referencing

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: count() is in-memory. It filters this.selections and sums lengths

packages/core/hooks/useComputePipelineMenuItems.ts
BrianWhitneyAI: No dont think so

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: [2f04b89](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/2f04b897dd8720bc5c33444bb78860bedc6fbf7b)

packages/core/hooks/useComputePipelineMenuItems.ts
BrianWhitneyAI: I added caching in [2f04b89](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/2f04b897dd8720bc5c33444bb78860bedc6fbf7b) but we can still move it

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: [021c218](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/021c21828cfd2a593c8868ab7f3e3711b4ba3a13)- Ok I limited it to 10 elements, I assume we will move to virtualization with a UX overhaul

packages/core/hooks/useComputePipelineMenuItems.ts
BrianWhitneyAI: [2f6cf9b](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/2f6cf9b46fcafee62e004b7f01be23618a8f02b7) split into a useEffect that fetches pipelines once (depends only on pipelineService), and a useMemo that rebuilds the menu items whenever pipelines or folderFilters change.

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: [1265bcc](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/1265bcc52814a8930e1d486d107310ad3f408ab5)

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: Im a little less knowledgeable here, agent suggests "Adding it to the dep array would cause the effect to re-run whenever the callback's identity changes, but the intent is for this to only run on mount or when pipelineService/selectedPipelineId change."

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,548 @@
...
+import BaseModal from "../BaseModal";
+import { PrimaryButton, SecondaryButton } from "../../Buttons";
+import BaseComboBox from "../../ComboBox";
+import AnnotationName from "../../../entity/Annotation/AnnotationName";
+import { Pipeline, PipelineParameter } from "../../../entity/ComputePipeline";
+import FileSelection from "../../../entity/FileSelection";
+import { interaction, selection } from "../../../state";
+import PipelineInfoCard from "./PipelineInfoCard";
+import PipelineParameterInput from "./PipelineParameterInput";
+import SelectedFilesExpander from "./SelectedFilesExpander";
```
aswallace: nit: these should go first in relative imports

packages/core/components/Modal/ComputePipelineModal/index.tsx
aswallace: Is the dep just `onSelectPipeline`?

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,548 @@
...
+        },
+        [pipelineService, selectedPipeline]
+    );
+
+    const requiredParams = parameters.filter((p) => p.required && p.type !== FILE_PATHS_PARAM);
+    const optionalParams = parameters.filter((p) => !p.required && p.type !== FILE_PATHS_PARAM);
+    const remainingOptionalParams = optionalParams.filter(
+        (p) => !addedOptionalNames.includes(p.name)
+    );
+    const addedOptionalParams = optionalParams.filter((p) => addedOptionalNames.includes(p.name));
```
aswallace, line 180: would it be possible to add a short comment clarifying the difference between these/how they're getting used?

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,548 @@
...
+        );
+    };
+
+    const title = selectedPipeline
+        ? selectedPipeline.name
+        : phase === "selecting"
+        ? "Select a Pipeline"
+        : "Compute Pipeline";
+
+    const footer = (
```
aswallace, line 512: May be unrelated to this PR, but noticing that the footer sometimes does a transparent overlap thing for narrow screens

Image

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,548 @@
...
+        if (fieldErrors[name]) setFieldErrors((prev) => ({ ...prev, [name]: "" }));
+    };
+
+    const onBlur = (param: PipelineParameter, value: unknown) => {
+        const err = validateParam(param, value);
+        setFieldErrors((prev) => ({ ...prev, [param.name]: err }));
+    };
+
+    const onSubmit = async () => {
+        if (!selectedPipeline || !selectedCluster) return;
```
aswallace, line 221: I'm wondering if the button should be disabled for these cases instead of allowing the user to click and have nothing happen

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -0,0 +1,548 @@
...
+
+import { ModalProps } from "..";
+import BaseModal from "../BaseModal";
+import { PrimaryButton, SecondaryButton } from "../../Buttons";
+import BaseComboBox from "../../ComboBox";
+import AnnotationName from "../../../entity/Annotation/AnnotationName";
+import { Pipeline, PipelineParameter } from "../../../entity/ComputePipeline";
+import FileSelection from "../../../entity/FileSelection";
+import { interaction, selection } from "../../../state";
+import PipelineInfoCard from "./PipelineInfoCard";
```
SeanDuHare: these bottom three should be above ".." import

packages/core/components/Modal/ComputePipelineModal/index.tsx
SeanDuHare: That is true, but the callbacks **should** never change. The exhaustive dep rule exists to help catch cases where dependencies are missed that are important for causing the effect so in general I'm in favor of avoiding removing deps, but... idk the agents are pretty good at catching missing dependencies so I'm alright seeing how this goes

aswallace (approved):

packages/core/components/Modal/ComputePipelineModal/index.tsx
aswallace: oop, duplicated comment with Sean's

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: I think this is for all of our modals, most just dont have alot of content

packages/core/services/PipelineService/index.ts
```diff
@@ -12,13 +12,18 @@ import HttpServiceBase, { ConnectionConfig } from "../HttpServiceBase";
...
+    private pipelinesCache: Promise<Pipeline[]> | null = null;
+
     constructor(config: ConnectionConfig = {}) {
         super(config);
     }
 
     getPipelines(): Promise<Pipeline[]> {
-        // TODO: return this.get(`${this.loadBalancerBaseUrl}/fss2/v4.0/pipelines`).then((r) => r.data);
-        return Promise.resolve(MOCK_PIPELINES);
+        if (!this.pipelinesCache) {
```
SeanDuHare, line 22: When using axios (in PR #800 ) this might automatically cache - might want to check before implementing a second layer of cache

SeanDuHare (approved):

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: [b62fba7](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/b62fba79c927778d86a675bf42fc3deaf6d8907e)

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: [b62fba7](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/b62fba79c927778d86a675bf42fc3deaf6d8907e)

packages/core/services/PipelineService/index.ts
BrianWhitneyAI: We are actually still not using it since we still dont have a discovery endpoint, just a google suggests that axios doesn't cache like this

packages/core/components/Modal/ComputePipelineModal/index.tsx
aswallace: I think the change in [b62fba7](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/b62fba79c927778d86a675bf42fc3deaf6d8907e) might actually be the reverse of our usual import order, usually it's `./` first and then `../` and `../..`

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: I added the cluster selection, pipeline doesnt need to be added since its selected by getting to the modal. you cant really unselect the cluster but I think its a good principle to add for if things changed. [74b6fba](https://github.com/AllenInstitute/biofile-finder/pull/799/commits/74b6fba963012e9eaae345e8aaf42987645ae462)

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: fixed!

PR #804, opened 2026-05-19
Feature/provenance updates

### LINKED ISSUE #650: Provenance: Make focused file more obvious
Lyndsay found during UX testing that the focused file that the graph is generated from is considered not very obvious - we can make it more obvious pretty easily

### LINKED ISSUE #651: Provenance: Make margin top/bottom between nodes bigger
From user feedback, the graph currently keeps the vertical sizing pretty tight - lets make this a bit bigger

### LINKED ISSUE #652: Provenance: Reset view button
User feedback: add global "Reset view" button in prov graph

### LINKED ISSUE #654: Provenance: Remove drag feature
Make nodes not draggable

### LINKED ISSUE #655: Provenance: Make graph interactions more discoverable
From user feedback, some graph interactions are less than obvious. Examples: cursor changes, zoom, pan indicator

### LINKED ISSUE #657: Provenance: Add open/download buttons to metadata panel
Add open/download buttons to metadata panel just like it is outside the provenance graph

aswallace:
## Context
These are the visual interaction updates to the provenance graph. 
Resolves #650, #651, #652, #654, #655

To do: 
- Display provenance button in file details panel and have it redraw the graph when clicked
- Check whether `Resize Observer` error appears in deployed versions of the graph
- Replace all remaining references to "provenance" in UI to "relationship diagram"
- Make url change so that can use browser back buttons

## Changes
- #650: Changes the font and background color for the focused file; changes border color for all other nodes
- #651: Increases the vertical spacing of the graph
- #652: Adds a reset button in the top right that recenters the graph and also reverts any layout changes (e.g., from compact/grid back to tree). This requires using the `ReactFlow` library to surface `fitView`, so needed to wrap the `ReactFlow` component inside of a `ReactFlowProvider` (provides access to the internal state)
- #654: Prevents users from moving nodes
- #655: Adds view controls in the bottom left
- #657: Make the `FileDetailsPanel` match how it appears in the main app. Undoes the graph view-specific customizations we'd previously added

#### Before: Graph
Screenshot 2026-05-18 at 4 50 10 PM

#### After: Graph
Screenshot 2026-05-18 at 4 50 24 PM

### Before vs After: File details panel
Screenshot 2026-05-18 at 4 56 11 PM
Screenshot 2026-05-18 at 4 56 22 PM


packages/core/components/FileDetails/index.tsx
```diff
@@ -200,24 +188,23 @@ export default function FileDetails(props: Props) {
...
-                                    {hasProvenanceSource && (
-                                        <DefaultButton
-                                            onClick={() =>
-                                                dispatch(
-                                                    interaction.actions.setOriginForProvenance(
-                                                        props.fileDetails
-                                                    )
+                            <div className={styles.titleRow}>
+                                <h4>Metadata</h4>
+                                {/* To do: un-hide the provenance source button while in graph mode, should redraw graph */}
```
BrianWhitneyAI, line 193: just double checking this should  still be here

packages/core/styles/global.css
```diff
@@ -55,6 +55,7 @@ body {
     --secondary-dark: #1c1c1c;
     --accent-dark: #3d3d3d;
     --medium-grey: #575859;
+    --med-light-grey: #999;
```
BrianWhitneyAI, line 58: Adding a new color to global seems significant. Do we need to do this?

BrianWhitneyAI (approved):

packages/core/components/FileDetails/index.tsx
aswallace: This is in the follow-up ticket #810

packages/core/styles/global.css
aswallace: the color change was a request from Lyndsay, I'll double check

packages/core/components/NetworkGraph/index.tsx
```diff
@@ -65,26 +63,73 @@ export default function NetworkGraph(props: NetworkGraphProps) {
         );
     }
 
+    // The ReactFlow component can only access state (useReactFlow) if it's the child of a ReactFlowProvider
+    // See https://reactflow.dev/learn/troubleshooting/common-errors#001
+    function ReactFlowComponent() {
```
SeanDuHare: Wondering if this could be cleaner somehow... does it have to be the direct parent? Maybe you could wrap `NetworkGraph` in the provider or move it out to its own component?

packages/core/components/NetworkGraph/index.tsx
```diff
@@ -65,26 +63,73 @@ export default function NetworkGraph(props: NetworkGraphProps) {
...
+        }, [graph, setEdges, setNodes, refreshKey]);
+        const { fitView } = useReactFlow();
+        const onClickReset = () => {
+            graph.resetLayout(); // return to default layout if any
+            fitView(); // reset zoom
+            dispatch(interaction.actions.refreshGraph());
+        };
+        return (
+            <>
+                <DefaultButton
```
SeanDuHare: bummer to have to reach into the style for this - any easy way to re-use a button component?

SeanDuHare (approved):
Minor thoughts - LGTM regardless

packages/core/components/NetworkGraph/index.tsx
aswallace: Switched to using a lightly modified TertiaryButton in [b2f2fec](https://github.com/AllenInstitute/biofile-finder/pull/804/commits/b2f2fec6f9ae1439317d631790e16b155073a36e)

packages/core/components/NetworkGraph/index.tsx
aswallace: Moved the ReactFlowProvider out to RelationshipDiagram in [7256b21](https://github.com/AllenInstitute/biofile-finder/pull/804/commits/7256b21f1197144305b6659ed15088c75da42ec8)

packages/core/components/NetworkGraph/index.tsx
```diff
@@ -47,8 +46,28 @@ const NODE_TYPES = {
  * Component for rendering a graph at the given origin
```
SeanDuHare, line 47: This is awesome! What do you think about one step further so that no user of the component has to really think about it? This way they would import an already wrapped graph, but that part is shoved away nicely (IMO)



```
function NetworkGraph(props: NetworkGraphProps) {
    ...body of function as is...
}

export default function WrappedNetworkGraph(props: NetworkGraphProps) {
    return (
            <ReactFlowProvider>
                <NetworkGraph {...props} />
            </ReactFlowProvider>
     );
 }
```

aswallace: That makes sense to me! Implemented in [dddd5cb](https://github.com/AllenInstitute/biofile-finder/pull/804/commits/dddd5cb02929552a26f783a97ddd6b516e330114), I was split on whether or not to rename the import in RelationshipDiagram but left it as is for now, open to changing it

PR #806, opened 2026-05-20
Add test cases: aggregate 10M+10M, 20M

pgarrison:
## Purpose

Test query performance of large aggregate parquets.

## Changes
* Two new benchmark cases: 10M+10M aggregate, and 20M.
  - This required substantial refactoring of the query benchmarking, including both the config data format and results data format
* TypeScript migration
  - Converted run-local.js, run-regression.js, compare-results.js, and run-benchmark-page.js to .ts with proper type annotations.
  - Added tsx as a devDependency to run TS scripts directly.
  - Updated package.json scripts to use tsx instead of node.
* CI / workflow updates
  - Doubled timeout to 360 minutes to accommodate larger tests (20 iterations takes 5+ hours).
  - Uses `npm run benchmark:*` scripts instead of direct node calls.
* Cleanup uses deleteDataSourceWrapper instead of raw DROP VIEW.

## Testing
* I used these changes thoroughly for the benchmark results presented in #787, but I've had to do some merging and rebasing since then. [This run](https://github.com/AllenInstitute/biofile-finder/actions/runs/26184183563) shows the benchmark results for the latest version.
* I also did some manual tests for `npm run benchmark` and `npm run benchmark:summary` with `--warmups=0`

## Reviewing
Note: the base branch is a temporary development branch, the result of merging PRs #739 and #787, both dependencies of this branch.


.github/workflows/benchmark.yml
```diff
@@ -36,8 +36,8 @@ jobs:
     # hardware variance between VMs would mask the small regressions we actually care about.
     # Running back-to-back on the same VM ensures both measurements share the same hardware
     # baseline, so deltas reflect code differences only.
-    # 180 minutes: fixture download + full task suite (including change_grouping on 10m) × 2 branches
-    timeout-minutes: 180
+    # 360 minutes is the maximum timeout
+    timeout-minutes: 360
```
BrianWhitneyAI, line 40: This is getting pretty long. I wonder if the aggregate should be split from the single sources.

dev-docs/07-query-benchmarking.md
```diff
@@ -25,6 +25,8 @@ mkdir -p packages/web/fixtures
 curl -fL "$BASE/synthetic-100k.parquet" -o packages/web/fixtures/synthetic-100k.parquet
 curl -fL "$BASE/synthetic-1m.parquet"   -o packages/web/fixtures/synthetic-1m.parquet
 curl -fL "$BASE/synthetic-10m.parquet"  -o packages/web/fixtures/synthetic-10m.parquet
+curl -fL "$BASE/synthetic-10m-copy.parquet"  -o packages/web/fixtures/synthetic-10m-copy.parquet
```
BrianWhitneyAI: do we need to use a real copy? could we synthetically rename?

packages/web/scripts/lib/run-benchmark-page.ts
```diff
@@ -6,13 +6,24 @@
  * BenchmarkConfig into the page, and returns the BenchmarkResults.
  */
 
-"use strict";
+declare global {
```
BrianWhitneyAI, line 12: Could you add some comments about this

packages/web/scripts/run-regression.ts
```diff
@@ -77,7 +109,4 @@ async function main() {
     console.log(`[regression] Results written to ${path.relative(process.cwd(), outFile)}`);
 }
 
-main().catch((err) => {
```
BrianWhitneyAI, line 80: remove error?

package.json
```diff
@@ -66,6 +66,7 @@
     "resize-observer": "1.0.x",
     "rimraf": "3.0.x",
     "sinon": "12.x",
+    "tsx": "^4.21.0",
```
BrianWhitneyAI: whats the change around the locks?

BrianWhitneyAI (commented):
I am worried that this aggregate benchmark is too much to put on top of the existing benchmark and should be split out into separate scripts.

package.json
pgarrison: Can you clarify?

packages/web/scripts/run-regression.ts
pgarrison: I removed the error handling so that errors will be logged with full stack traces instead of a single line error message. I did this for debugging because I triggered an error that I needed more context for, and I figured it would simplify future debugging to keep it this way.

packages/web/scripts/lib/run-benchmark-page.ts
pgarrison: Done [dbc1d14](https://github.com/AllenInstitute/biofile-finder/pull/806/commits/dbc1d145278fc80076c9849828b6987b29421b59)

dev-docs/07-query-benchmarking.md
pgarrison: Updated to do a local copy instead of remote. [62082ae](https://github.com/AllenInstitute/biofile-finder/pull/806/commits/62082aefc3fe85fa63e4a6436388ed73a83ebc02)

It's nice to have separate files so that they aren't cached together.

.github/workflows/benchmark.yml
SeanDuHare: As in as a different workflow? I could see the benefit of a "quick benchmark" vs a "full benchmark" or something like that - probably outside the scope of this PR IMO

package.json
SeanDuHare: The versions are handled via the "x" version classifier as opposed to ~ or ^.

Bummer this has to be included in the web deploy

```suggestion
    "tsx": "4.21.x",
```

SeanDuHare (approved):

package.json
pgarrison: It's in the dev dependencies, so it shouldn't be in the deployment? Perhaps I'm misunderstanding.

(Updated the version syntax in [6079147](https://github.com/AllenInstitute/biofile-finder/pull/806/commits/6079147a889cc0f0e9a73a52c24e694036aba658))

pgarrison:
Latest benchmark run: https://github.com/AllenInstitute/biofile-finder/actions/runs/26601221187

.github/workflows/benchmark.yml
pgarrison: I'd like to get #809 merged before making that change. I filed a new issue for it #818

package.json
BrianWhitneyAI: I am just slightly worried about dep changes for code that isnt directly used by the app. This was more a comment about those changes as a whole rather than specifically "tsx"

BrianWhitneyAI: ah seeing its in the dev dependecies in the comment below makes me less worried about this

BrianWhitneyAI (approved):
Thank you for fixing my bad error commit!

pgarrison:
Latest benchmark run: https://github.com/AllenInstitute/biofile-finder/actions/runs/26655081926

PR #808, opened 2026-05-20
New browser instance per benchmark task

pgarrison:
## Purpose

Enable benchmarking cold-start scenarios by using a new browser instance for every trial. This feature is enabled only when the user sets `warmups=0`.

Builds on #739, #806,

## Changes
* The benchmark page (`index.ts`) accepts an optional list of tasks to run, and runs all tasks by default.
* `run-benchmark-page.ts` now splits the logic into `runInBrowser` (for `warmups > 0`) and `runColdStartBenchmarks`, which makes use of `runInBrowser`.
* Fixed a bug where I used `in` instead of `has` for set membership.
* Extracted some logic to `stats.ts` (may need a better name) and `injectFixtures` for DRY and readability.

## Testing
Without this change, running tests with `warmups=0` shows a significant performance decrease for the first trial of each batch. The test below used 4 batches with 15 or 20 iterations each.

Before:
Screenshot 2026-05-28 at 2 36 19 PM

After this change, with `warmups=0`, there are no obvious outliers.

[After](https://github.com/AllenInstitute/biofile-finder/actions/runs/26603676754) (2 batches, 5 iterations each):
Screenshot 2026-05-29 at 10 53 40 AM

### DIFF
--- a/packages/web/benchmark/src/index.ts
+++ b/packages/web/benchmark/src/index.ts
@@ -1,9 +1,7 @@
 import { BENCHMARK_TASKS, createServices } from "./tasks";
 import { BenchmarkConfig, BenchmarkResults, QueryResult, SourceResult } from "./types";
+import { DEFAULT_ITERATIONS, DEFAULT_WARMUP_ROUNDS, buildQueryResult } from "./stats";
 import DatabaseServiceWebWorker from "../../src/services/DatabaseServiceWeb/duckdb-worker.worker";
-
-const DEFAULT_ITERATIONS = 5;
-const DEFAULT_WARMUP_ROUNDS = 1;
 
 // Updates the #status element in the benchmark HTML page and mirrors to console.
 // The page can run headlessly in CI (Playwright), so the console log is the
@@ -12,14 +10,6 @@ function setStatus(msg: string) {
     const el = document.getElementById("status");
     if (el) el.textContent = msg;
     console.log("[benchmark]", msg);
-}
-
-// Nearest-rank percentile over a pre-sorted array. Used to report p50 and p95
-// across timed iterations — p95 surfaces occasional slow outliers (GC pauses,
-// DuckDB cache misses) that the median would hide.
-function percentile(sorted: number[], p: number): number {
-    const idx = Math.ceil((p / 100) * sorted.length) - 1;
-    return sorted[Math.max(0, idx)];
 }
 
 // Fisher-Yates shuffle — randomizes task order each timed iteration so that a
@@ -95,16 +85,7 @@ async function benchmarkSource(
         }
     }
 
-    return tasks.map(({ name }) => {
-        const timings = [...(timingsMap.get(name) ?? [])].sort((a, b) => a - b);
-        return {
-            name,
-            timings,
-            p50: percentile(timings, 50),
-            p95: percentile(timings, 95),
-            p99: percentile(timings, 99),
-        };
-    });
+    return tasks.map(({ name }) => buildQueryResult(name, timingsMap.get(name) ?? []));
 }
 
 async function main() {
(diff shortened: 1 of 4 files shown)


packages/web/benchmark/src/index.ts
```diff
@@ -14,14 +12,6 @@ function setStatus(msg: string) {
...
 }
 
-// Nearest-rank percentile over a pre-sorted array. Used to report p50 and p95
-// across timed iterations — p95 surfaces occasional slow outliers (GC pauses,
-// DuckDB cache misses) that the median would hide.
-function percentile(sorted: number[], p: number): number {
-    const idx = Math.ceil((p / 100) * sorted.length) - 1;
-    return sorted[Math.max(0, idx)];
-}
-
```
pgarrison, line 24: These were moved to `stats.ts` for reuse.

packages/web/scripts/lib/run-benchmark-page.ts
```diff
@@ -254,10 +359,17 @@ export async function runBenchmarkPage({
...
+        );
+        const result = benchmarkResults.results[0];
+        return {
+            initTimeMs: benchmarkResults.initTimeMs,
+            registrationMs: result.registrationMs,
+            queries: result.queries,
+        };
     } finally {
         await browser.close();
-        await new Promise((res) => server.close(res));
```
pgarrison, line 260: This was moved to line 182

packages/web/scripts/lib/run-benchmark-page.ts
```diff
@@ -203,38 +309,7 @@ export async function runBenchmarkPage({
         // The browser reads the file lazily via FileReader (BROWSER_FILEREADER protocol),
         // which is identical to how the real app loads files via the file picker —
         // no HTTP range-request overhead, so DuckDB sort performance matches real-user timing.
-        const loaded = new Set();
-        for (const source of testCases.flat()) {
-            if (source.label in loaded) continue; // Don't add duplicate sources
```
pgarrison, line 208: This whole section was moved verbatim to `injectFixtures`, with one exception: I changed this condition to `loaded.has(source.label)` to fix a bug. (`in` does not check set membership!)

BrianWhitneyAI (approved):
Approved under the assumption the lint issues are cleaned up

packages/web/scripts/lib/run-benchmark-page.ts
```diff
@@ -167,6 +170,108 @@ export async function runBenchmarkPage({
...
+        sourceResults.push({
+            labels: testCase.map((source) => source.label),
+            registrationMs,
+            queries: allTaskNames.map((name) => buildQueryResult(name, timingsMap.get(name) ?? [])),
+        });
+    }
+
+    return {
+        timestamp: new Date().toISOString(),
+        commit: "unknown",
```
SeanDuHare, line 252: Are these two always `unknown`? It looks like the browser sets [them to unknown too](https://github.com/AllenInstitute/biofile-finder/blob/9e1c2a5ecdbde6ca99927a070ed6a8f5a05a2136/packages/web/benchmark/src/index.ts#L188-L189)

SeanDuHare (approved):

packages/web/scripts/lib/run-benchmark-page.ts
pgarrison: They are set in the `run-local.ts` [path](https://github.com/AllenInstitute/biofile-finder/blob/9e1c2a5ecdbde6ca99927a070ed6a8f5a05a2136/packages/web/scripts/run-local.ts#L130). I suspect they could be set here too, but would leave that to future work.

pgarrison:
Waiting on https://github.com/AllenInstitute/biofile-finder/actions/runs/26973097008 to complete successfully after latest adjustments

pgarrison:
> Approved under the assumption the lint issues are cleaned up

Done [4576877](https://github.com/AllenInstitute/biofile-finder/pull/808/commits/4576877042e32f15620e359f0a974a6f7b6bd1da)

PR #814, opened 2026-05-21
bugfix/pipeline-card-info

BrianWhitneyAI:
## Description
Got mixed up about supported files here for the acm pipeline when making the placeholder data. supports zarr not tiff.

### DIFF
--- a/packages/core/services/PipelineService/mockPipelineData.ts
+++ b/packages/core/services/PipelineService/mockPipelineData.ts
@@ -10,7 +10,7 @@ export const MOCK_PIPELINES: Pipeline[] = [
         name: "All Cells Mask Segmentation",
         description:
             "The All Cells Mask workflow generates segmentation masks for the selected files.",
-        restrictions: "Works only for Zarr and CZI files under 100gb",
+        restrictions: "Works only for Zarr and CZI files under 100 GB",
         clusters: ["slurm-prod"],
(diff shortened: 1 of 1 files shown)


packages/core/services/PipelineService/mockPipelineData.ts
```diff
@@ -10,9 +10,9 @@ export const MOCK_PIPELINES: Pipeline[] = [
         name: "All Cells Mask Segmentation",
         description:
             "The All Cells Mask workflow generates segmentation masks for the selected files.",
-        restrictions: "Works only for Tiff and CZI files under 100gb",
+        restrictions: "Works only for Zarr and CZI files under 100gb",
```
aswallace: Small nit I missed before, could we do `100 GB` instead?

aswallace (approved):

packages/core/services/PipelineService/mockPipelineData.ts
BrianWhitneyAI: yeah

BrianWhitneyAI: [47349a7](https://github.com/AllenInstitute/biofile-finder/pull/814/commits/47349a721749f4f27163c6c60d0a0726b21c0155)

SeanDuHare (approved):

PR #819, opened 2026-05-28
Fast benchmark mode

### LINKED ISSUE #818: Add a faster option to the query benchmarks
## Use Case

A developer wants a quick sanity check performance comparison, and they don't want to wait for the full benchmark suite to run.

## Suggest a Solution

Add a boolean option, exposed via workflow dispatch, to disable the 10M+10M and 20M test cases, and also skip the large sort task (this query takes as much time as the others combined).

pgarrison:
## Context
Closes #818 

## Changes
* Give run-regression a `--skip-large` flag to skip 10m+10m, 20m test cases and the large sort task
* Add an option to `benchmark.yml` to expose the flag to the user

## Testing
* See test run here with `iterations=1` and `warmups=0` (using the changes from #808, every task ran in a new browser instance) that skips the expected test cases and task: https://github.com/AllenInstitute/biofile-finder/actions/runs/26606173872

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -22,6 +22,11 @@ on:
         required: false
         type: string
         default: "1"
+      skip_large:
+        description: "Skip 10m+10m, 20m test cases and the large sort task"
+        required: false
+        type: boolean
+        default: false
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


SeanDuHare (approved):
big fan 🚀

aswallace (approved):

BrianWhitneyAI (approved):

pgarrison:
Waiting on https://github.com/AllenInstitute/biofile-finder/actions/runs/26981707916

PR #820, opened 2026-05-28
#817 Don't confuse tables where one name is a prefix of the other

### LINKED ISSUE #817: Rows from wrong data source displayed
## Description

If one parquet data source has a file name that is a prefix of another, queries against the shorter name will show results for both.

## Reproduction

* Add data source `foo2.parquet`
* Add data source `foo.parquet` (separate query: not an aggregate source)
* Observe that rows from `foo2` appear in the `foo` query.

## Root cause
https://github.com/duckdb/duckdb-wasm/issues/2227

## Suggested fix
Instead of `foo` and `foo2` as file handles, use `foo-bff-filehandle` and `foo2-bff-filehandle`. This doesn't completely eliminate the bug, since a user could upload `foo.parquet` and `foo-bff-filehandle.parquet`, but it makes it very unlikely to come up.

## Environment
- [x] Web
   - Browser: Firefox

pgarrison:
## Context
Resolves #817 

## Changes
To work around the underlying bug, register files to duck DB with an appended `-bff-filehandle`. Unfortunately, this means updating every place where the SQL references the file handle name to use the new `fileHandleName` function. It's a little tricky to ensure all of these sites were identified: I missed the ones in `parquetHasBlankValues` at first. I suspect all of the sites that need to change are in the data source preparation phase, because subsequent queries use view/table names instead.

## Testing
 Three manual tests:
 1. Two parquet files `tiny.parquet` and `tiny2.parquet` and verify both appear as distinct data sources with correct row counts.
 2. `parquet_metadata`: Load a single parquet file with a column containing blank/whitespace values and confirm the column filter UI correctly identifies blanks.
 3. Non-parquet: Load a CSV and a JSON to verify they still load and aggregate as normal.


BrianWhitneyAI (approved):

SeanDuHare (approved):

pgarrison:
Whoops, realized I did my testing without pushing these changes: [29713c2](https://github.com/AllenInstitute/biofile-finder/pull/820/commits/29713c297ae191a34c2f972a0207f519eee4e060). @SeanDuHare @BrianWhitneyAI

PR #826, opened 2026-06-09
Feature/provenance query params

### LINKED ISSUE #656: Provenance: Make URL change during transition state
User feedback, when going into prov mode and going from prov mode changing the URL history would help so the back and forward browser buttons this would likely also enable showing specifically the prov graph directly (like the first thing you see when you click the link is the graph)

aswallace:
## Context
Allows the relationship diagram (provenance graph) to render immediately if users are given a url that contains provenance-related query parameters. Closes #656 

## Not done
Currently unable to sync the react router to the browser back button. While clicking that button does add/remove query params, I couldn't get it set up to also affect the UI in a way that wasn't extremely buggy because of the way our queries and other components listen to state changes. This is theoretically possible with a `useLocation()` hook and/or event listeners, but would take more effort, and would affect more than just this ticket.

#### To do
#828 is a follow-up PR for making the graph less dependent on the "File ID" annotation (i.e., use `file.path` for the node IDs instead of `file.id`

## Changes
- Creates a query parameter `p` that, if included in the url, indicates the unique ID of the origin of the graph.
- Moves `getFileByUid` into the `FileService` so that we can use it in the query param logic to convert UIDs into FileDetails
- Some minor changes to when the relationship diagram renders. Loading a graph directly from a url means that we need to wait to make sure that the file defining provenance relationships has been fully processed and that the uid actually refers to a real file

## Testing
Mostly manual testing with [this test file](https://docs.google.com/spreadsheets/d/e/2PACX-1vTesK5EHZZAsqG6YvAjvc6pCiRFgM1gQ-1hQFwkN15sF2g4q3ooXTwR0dzUY3KriCy0LEMr904Ik7uq/pub?gid=978740010&single=true&output=csv) and [provenance definitions](https://docs.google.com/spreadsheets/d/e/2PACX-1vR8qCTDa2L9KYisWqiVf4xtf3Tr56zDT3pyFYU7LbNJq91rhQCg2ABeVESp5MLCPT-AjdqKH0ZAJAmv/pub?gid=1550180462&single=true&output=csv)


packages/core/components/CoreContent/CoreContent.tsx
```diff
@@ -6,22 +6,36 @@ import GlobalActionButtonRow from "../GlobalActionButtonRow";
...
+import SearchParams from "../../entity/SearchParams";
+import { interaction, metadata, selection } from "../../state";
 
 import styles from "./CoreContent.module.css";
 
 /**
  * Core content of the application
  */
 export default function CoreContent() {
+    const searchParams = window.location.search;
```
SeanDuHare, line 18: What causes this variable to update? Does a render of `CoreContent` have to occur for it to happen or does it update automatically and cause `CoreContent` to re-render?

packages/core/entity/SearchParams/index.ts
```diff
@@ -28,7 +28,8 @@ export interface SearchParamsComponents {
     fileView?: FileView;
     sources: Source[];
     sourceMetadata?: Source;
-    prov?: Source;
+    prov?: Source; // file containing provenance relationship info
```
SeanDuHare: Hmm perhaps this should be renamed to be more specific now? `provenanceSource`?

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -137,6 +138,32 @@ export default class DatabaseFileService implements FileService {
         return rows.map((row) => DatabaseFileService.convertDatabaseRowToFileDetail(row, env));
     }
 
+    /**
+     * Get a single file that has a unique ID matching uid
+     */
+    public async getFileByUid(uid: string): Promise<FileDetail | undefined> {
```
SeanDuHare, line 144: This is the same in the Http version too right? If so, perhaps `FileService` should be an abstract class holding this function

aswallace: The difference is that in DatabaseFileService we use the `HIDDEN_UID_ANNOTATION`, which doesn't exist for http/FMS (uses File ID instead)

packages/core/components/CoreContent/CoreContent.tsx
aswallace: It doesn't cause CoreContent to re-render, so this is just a static check on component load

packages/core/entity/SearchParams/index.ts
aswallace: updated in [87c0511](https://github.com/AllenInstitute/biofile-finder/pull/826/commits/87c05115bccf9d52325dd4f802956a1d5c0f6c4d)

packages/core/components/CoreContent/CoreContent.tsx
```diff
@@ -6,22 +6,36 @@ import GlobalActionButtonRow from "../GlobalActionButtonRow";
...
     const requiresDataSourceReload = useSelector(selection.selectors.getRequiresDataSourceReload);
 
     const hasSomethingToQuery = hasQuerySelected || window.location.search;
     const hasNeedToSelectQuery = requiresDataSourceReload || !hasSomethingToQuery;
 
-    if (origin) {
+    // The relationship diagram should only display if:
+    // - the edge definitions have been fully loaded
+    // - the url still contains an ID for the origin
+    // - the ID for the origin can be processed into a file
```
SeanDuHare: Could you elaborate on how `!!origin` represents "the ID for the origin can be processed into a file"?

packages/core/services/FileService/DatabaseFileService/index.ts
SeanDuHare: ohhhhhh okay 👍

SeanDuHare (approved):

packages/core/components/CoreContent/CoreContent.tsx
aswallace: ah yes, so `!!origin` in this case is _not_ the UID from the url, it's the file itself after it has been processed by state logic. The selector here is `getOriginForProvenance`, which returns a `FileDetails` object if `changeProvenanceOriginIdLogic` successfully looked up the ID from the search parameters and found a file (otherwise is undefined).

I'm not sure what the best way to clarify this is in the comments/naming, do you have thoughts?

packages/core/components/DataSourcePrompt/index.tsx
```diff
@@ -77,7 +77,7 @@ export default function DataSourcePrompt(props: Props) {
                     parts: {
                         sources: [dataSource],
                         sourceMetadata: metadataSource,
-                        prov: provenanceSource,
+                        provenanceSource,
```
BrianWhitneyAI: for readability could we do 'provenanceSource: provenanceSource' like metadata?

packages/core/components/FileDetails/index.tsx
```diff
@@ -194,8 +194,8 @@ export default function FileDetails(props: Props) {
                                     <DefaultButton
                                         onClick={() =>
                                             dispatch(
-                                                interaction.actions.setOriginForProvenance(
-                                                    props.fileDetails
+                                                selection.actions.changeProvenanceOriginId(
+                                                    props.fileDetails?.uid
```
BrianWhitneyAI, line 198: uid is guaranteed right?

BrianWhitneyAI (approved):

packages/core/components/FileDetails/index.tsx
aswallace: uid is guaranteed, but fileDetails is not

packages/core/components/DataSourcePrompt/index.tsx
aswallace: changed in [e8c4058](https://github.com/AllenInstitute/biofile-finder/pull/826/commits/e8c4058589513557ae300ae8cfc4390bf26c11fd)

packages/core/components/CoreContent/CoreContent.tsx
aswallace: I updated the comment slightly and am going to move forward with merging, but happy to revisit in a follow-up if needed

PR #828, opened 2026-06-09
Remove dependence on File ID in provenance graph

### LINKED ISSUE #789: Simplify provenance file tracking
Currently the provenance file tracking system relies on the user having a File ID which is less than ideal. Lets investigate if we can find a way around that. Something to note is the user will need to supply something to tell BFF when a column is self-referential (ex. in when a "Child" of a "Well" is intended to be the file that contains that well information) - currently the user inputs "File ID" which is confusing at best

aswallace:
## Context
addresses #789. The provenance graph is currently dependent on `File ID`, which is an internal annotation and not one that we actually require from users. This PR removes that dependence by changing the graph node IDs to use "File Path," since that is a column we require. Also adds a `self` type to indicate when we're referring to the file itself (as opposed to metadata or a different file)

For backwards compatibility, currently also has a fallback check for File ID.

## Testing
The main issue I've been running into is trying to decide how flexible to be in terms of different potential provenance schemas a user might provide. 

- Tested that the original test files still work-- I wasn't sure how concerned we are about backwards compatibility since we don't have people actively using this yet
    - [data file](https://docs.google.com/spreadsheets/d/e/2PACX-1vTesK5EHZZAsqG6YvAjvc6pCiRFgM1gQ-1hQFwkN15sF2g4q3ooXTwR0dzUY3KriCy0LEMr904Ik7uq/pub?gid=978740010&single=true&output=csv)
    - [provenance schema ](https://docs.google.com/spreadsheets/d/e/2PACX-1vR8qCTDa2L9KYisWqiVf4xtf3Tr56zDT3pyFYU7LbNJq91rhQCg2ABeVESp5MLCPT-AjdqKH0ZAJAmv/pub?gid=1550180462&single=true&output=csv)
- Tested new files that use "File Path" instead of "File ID" for the input file, blank when referencing self, and also the new "self" type
    - [data file](https://docs.google.com/spreadsheets/d/e/2PACX-1vSgzDzX5_3RepRdCxx1H-q34jckZcZYiir0pu9rjwVfTl5QyfdqdpaBR_QLAT5E6SwDDF-ucDiKt6eJ/pub?output=csv) 
    - [provenance schema ](https://docs.google.com/spreadsheets/d/e/2PACX-1vTzcU7vJdglvA0d4ypI51bAwkU5DD-4erGQIiUuaFA4pf_3z-j2-wN47-uqu1UWkqSNdLkWxpHvjGZ5/pub?output=csv)


packages/core/entity/Graph/index.ts
```diff
@@ -13,6 +13,10 @@ const METADATA_NODE_HEIGHT = 90;
 const ROW_SPACING = Math.max(FILE_NODE_HEIGHT, METADATA_NODE_HEIGHT) + 25;
 const COLUMN_SPACING = Math.max(FILE_NODE_WIDTH, METADATA_NODE_WIDTH) + 25;
 
+// Using these in the provenance relationship definitions means we're uniquely identifying a file
+const NON_ANNOTATION_FILE_IDENTIFIERS = ["self", "id", "uid", ""]; // These indicate to use the uid of the current file, not an actual annotation
```
SeanDuHare: It feels a little like `id` or `uid` could be used as actual columns that mean something which could cause a bug

packages/core/entity/Graph/index.ts
```diff
@@ -132,7 +136,7 @@ export function getGridPosition(
  */
 function createFileNode(file: FileDetail, isSelected = false): FileNode {
     return {
-        id: file.id, // TODO: This doesn't have to be tied to FMS's ID system, we could be more generic here
+        id: `${file.uid}`, // quotes to stringify
```
SeanDuHare: I think `uid` is always a string

packages/core/entity/Graph/index.ts
```diff
@@ -469,15 +481,22 @@ export default class Graph {
         );
     }
 
+    private static isNodeAFile(node: EdgeNode): boolean {
```
SeanDuHare: (nit) move static to top of class

packages/core/entity/Graph/index.ts
```diff
@@ -13,6 +13,10 @@ const METADATA_NODE_HEIGHT = 90;
 const ROW_SPACING = Math.max(FILE_NODE_HEIGHT, METADATA_NODE_HEIGHT) + 25;
 const COLUMN_SPACING = Math.max(FILE_NODE_WIDTH, METADATA_NODE_WIDTH) + 25;
 
+// Using these in the provenance relationship definitions means we're uniquely identifying a file
+const NON_ANNOTATION_FILE_IDENTIFIERS = ["self", "id", "uid", ""]; // These indicate to use the uid of the current file, not an actual annotation
+const FILE_IDENTIFIERS = [...NON_ANNOTATION_FILE_IDENTIFIERS, "file id", "file path"];
```
SeanDuHare: Why `file id`? - moreover, what the goal with this variable (see other comment where used)

packages/core/entity/Graph/index.ts
```diff
@@ -469,15 +481,22 @@ export default class Graph {
...
         thisNode: FileNode,
-        edgeNode: EdgeNode
+        edgeNode: EdgeNode,
+        identifierType: string // if we need to look up the file, which annotation to use (e.g., File Path, File ID)
     ): Promise<(FileNode | MetadataNode)[]> {
         // The node might just be the current node!
-        const isEdgeNodeThisNode = edgeNode.name === "File ID"; // TODO: Expand this... should there be type of "ID" or "Self" or smthn..?
+        const isEdgeNodeThisNode =
+            edgeNode.type === "self" ||
+            FILE_IDENTIFIERS.includes(edgeNode.name.toLocaleLowerCase());
```
SeanDuHare: Shouldn't type always just be self here if so? What do these other cases cover?

packages/core/entity/Graph/index.ts
```diff
@@ -492,19 +511,25 @@ export default class Graph {
...
         return Promise.all(
-            (annotation.values as string[]).map(async (fileId) => {
+            (annotation.values as string[]).map(async (value) => {
                 // Avoid re-requesting the file when possible
-                const node = this.graph.node(fileId);
-                if (node) return node;
-                const file = await this.fileService.getFileByUid(fileId);
-                if (file) return createFileNode(file);
-                throw new Error(`Unable to find file ${fileId}`);
+                if (NON_ANNOTATION_FILE_IDENTIFIERS.includes(identifierType.toLocaleLowerCase())) {
```
SeanDuHare: Could you add a comment explaining what this case is for? It seems like this is the same as if the type were `self`?

packages/core/entity/Graph/index.ts
```diff
@@ -560,6 +585,39 @@ export default class Graph {
         });
     }
 
+    /**
+     * Get a single file that matches an annotation/value pair
+     */
+    private async getFileByIdentifier(
```
SeanDuHare: I think this language could be simpler:
```
private async getFileBy(column: string, value: (string | number | boolean)[]): Promise<FileDetail | undefined {
```

packages/core/entity/Graph/index.ts
```diff
@@ -492,19 +511,25 @@ export default class Graph {
...
-                const node = this.graph.node(fileId);
-                if (node) return node;
-                const file = await this.fileService.getFileByUid(fileId);
-                if (file) return createFileNode(file);
-                throw new Error(`Unable to find file ${fileId}`);
+                if (NON_ANNOTATION_FILE_IDENTIFIERS.includes(identifierType.toLocaleLowerCase())) {
+                    const node = this.graph.node(value);
+                    if (node) return node;
+                }
+                const file = await this.getFileByIdentifier(identifierType, [value]);
```
SeanDuHare: So this is when the column is something like `Input File ID`?

packages/core/entity/Graph/index.ts
aswallace: ah yes, this was something I meant to include in the PR description -- There are technically two potential solutions here that we need to choose between, and this line is a place where we'd need to remove one of them:
1. We add the type "self" and ignore whatever users put in the (Child if it's the child type, Parent if it's the parent type) column. 
2. The other option is that we don't add a new "self" type , and instead allow some fixed set of labels that we can decide on that would go in the (Child or Parent) column to indicate that we're referring to the same file, e.g., "self", blank, or something else. 
I think the first option is cleaner, but it's also not backwards compatible (if we care about that)

packages/core/entity/Graph/index.ts
aswallace: If we end up going in this direction (my preference is towards using the "self" type) then this is referencing the "File ID" annotation, but lowercase for string matching purposes

packages/core/entity/Graph/index.ts
aswallace: I'm realizing that my logic here is actually incorrect... BUT the problem I'm trying to address is that if we use the file uid as the node id, we need to find a way to look up files for situations like `Input File`, since the user is likely supplying either a `File Path` or a `File ID` and doesn't have access to the internal uid

packages/core/entity/Graph/index.ts
aswallace: moved in [ab474d7](https://github.com/AllenInstitute/biofile-finder/pull/828/commits/ab474d7d8ab5c02ae24ecdedacd82048e5aff872)

packages/core/entity/Graph/index.ts
aswallace: renamed in [1cf4fa1](https://github.com/AllenInstitute/biofile-finder/pull/828/commits/1cf4fa18d17056332e7679e942127707c668a106)

packages/core/entity/Graph/index.ts
aswallace: Settled on "File Path" with a fallback to "File ID" in [1cf4fa1](https://github.com/AllenInstitute/biofile-finder/pull/828/commits/1cf4fa18d17056332e7679e942127707c668a106)

packages/core/entity/Graph/index.ts
```diff
@@ -492,19 +496,26 @@ export default class Graph {
...
-                const file = await this.fileService.getFileByUid(fileId);
-                if (file) return createFileNode(file);
-                throw new Error(`Unable to find file ${fileId}`);
+                try {
+                    const file = await this.getFileBy("File Path", [value]);
+                    if (file) {
+                        return createFileNode(file);
+                    }
+                } catch {
+                    // try looking up by id
```
SeanDuHare:
```suggestion
                    // Backup while moving between provenance versions: Try looking by "File Id", this should be removed in the future
                    
```

packages/core/entity/Graph/index.ts
```diff
@@ -492,19 +496,26 @@ export default class Graph {
...
         }
 
         return Promise.all(
-            (annotation.values as string[]).map(async (fileId) => {
+            (annotation.values as string[]).map(async (value) => {
                 // Avoid re-requesting the file when possible
-                const node = this.graph.node(fileId);
+                const node = this.graph.node(value); // the value should be a file path
                 if (node) return node;
-                const file = await this.fileService.getFileByUid(fileId);
```
SeanDuHare, line 505: is `getFileByUid` removed elsewhere?

SeanDuHare (approved):

packages/core/entity/Graph/index.ts
aswallace: It still gets used in #826 for the query params since we still use the ID in our url instead of the path, but we could potentially remove it in favor of this more general lookup function

BrianWhitneyAI (approved):

PR #830, opened 2026-06-12
Admin/Update node

### LINKED ISSUE #751: Upgrade node

aswallace:
## Context
Update Node from v20 to the current active deployment (v24). 
closes #751 

## Changes
In order to update Node, needed to also update electron from 22 -> 40+.  This also required some minor modifications to the electron renderer since the API has changed slightly in the last 20 versions.

GitHub was also showing this warning in our workflows: 
```Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: AllenCellSoftware/action-electron-builder@fms-file-explorer. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24.```
so I created a new branch with updated node/electron in [the repo we use for building the app](https://github.com/AllenCell/action-electron-builder/tree/biofile-finder).

I also needed to upgrade playwright in order for the benchmark workflow to not time out during the playwright chromium download. It now seems to be running fine, but unfortunately can't fully run the comparison between this branch and main since they now require two different versions of node

## Testing
I made sure the unit and integration tests run and that both web and desktop build & render, but I could use some help in verifying that our various workflows still work as expected (especially the benchmarking workflow).


BrianWhitneyAI (approved):
Its probably ok for the benchmark to break to push this through given the urgency

LuuOW (commented):
Technical audit: code implementation verified for system consistency.

SeanDuHare (approved):

Issue #526, opened 2025-05-14
Lyndsay Test Issue


lynwilhelm:
### Description
- _Type of UX work needed (research, discovery and/or design)_
- _Problem(s) to solve_

***
_Detail what you already know and/or your best guesses. UX will help fill in blanks once assigned._

### Size
_Effort level (xs, sm, m, lg, xl)_

### Scope
_What will we do? What will we NOT do?_

### User needs statement(s)
As a _[type of user]_, I need/want to _[action(s)]_, so that I can _[goal]_.

### Additional notes
_Knowns/unknowns, stakeholders, time/dependency considerations, deliverables_

will-moore:
I would love to be able to specify a way to "Open with" for multiple selected Images.

For example, in our usage from OMERO, every row has an image ID. It would be great if multiple selected Images could be opened in a URL like `www.omero/webclient/?image=1&image=2&image=3` or `www.omero/webclient/images?ID=1&ID=2&ID=3` although you could risk exceeding the typical request string limit of 4096 characters, so something more concise might be nicer like: `www.omero/webclient/?image=1,2,3,4,5`.

SeanDuHare:
Related to #351 - feels time to see if we can slot this into the implementation roadmap for essentially the same case Will alludes to

will-moore:
Is there some documentation on how an author can specify "Open-with" urls?

I think this is relatively easy to do for single selected rows. But harder to do for multiple rows.

In OMERO.webclient we allow users to specify "Open with" - see https://omero.readthedocs.io/en/stable/developers/Web/LinkingFromWebclient.html#open-with

Essentially, it allows users to specify a `url_to_app` and the default behaviour is to open multiple images with image IDs like `url_to_app?image=1&image=2` or if you want something different, you can supply a JavaScript function that gets a list of the selected object IDs and should return a URL. 

From the "figure" point of view see 

https://ome.github.io/omero-figure/new?image=https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr&image=https://livingobjects.ebi.ac.uk/idr/zarr/v0.4/idr0048A/9846151.zarr/0

NB: I propose to remove the need for `/new` in that URL - see https://github.com/ome/omero-figure/pull/688

cc I discussed this with Davide Panzeri @DBP008 at the OME meeting yesterday since he is using BFF for managing all his Lab's images https://github.com/MMV-Lab/ambiom-bff-portals

[...]

PR #831, opened 2026-06-14
Open with OMERO when app is deployed in omero-biofilefinder

will-moore:
## Context
This is a suggestion / question / discussion on how to improve the `Open With` menu when the bff app is deployed
within [omero-biofilefinder](https://github.com/ome/omero-biofilefinder/).
It extends the `Open With > IDR Viewer` logic from #762 with an alternative `Open With > OMERO`.

## Changes
This uses the current URL of the bff app to detect when to show the `OMERO` option. When the app is deployed in omero-bff, the URL will be "https://my-institute-server.org/optional_prefix/biofilefinder/bff/app/".

The csv or parquet file that omero-bff generates may contain links to images in the main webclient, e.g. `https://my-institute-server.org/optional_prefix/webclient/?show=image-123` or Images in the viewer, e.g. `https://my-institute-server.org/optional_prefix/img_detail/123` or ROIs / Shapes in the iviewer e.g. `https://my-institute-server.org/optional_prefix/iviewer?roi=456`. In each case, `Open With > OMERO` works well as a menu text.

## Testing
Testing fully needs omero-biofilefinder.

https://github.com/user-attachments/assets/ec3c7269-95b3-4cd9-bd02-128281dbb3c3


SeanDuHare (approved):
I'd prefer we didn't continue this direction since it requires hardcoding URLs and had the apps hand off what apps they would like individual files to be loaded with - I'm ok rolling this one in at least for now though. My hope is getting the user guide released this summer will help avoid the need for this type of intervention

will-moore:
Thanks for reviewing. I totally understand the concerns about hardcoding etc, hence my "suggestion / question / discussion".
Since it's possible for me to build and bundle any branch/local commit of BFF into omero-biofilefinder, this doesn't *need* to be in the mainline for me to use it. Having it in the mainline would simply save me from repeatedly rebasing this each time I update BFF in omero-biofilefinder.
At the moment it's quite a manual process: I plan to make BFF into a submodule of omero-biofilefinder and setup a build process that would avoid me committing the built assets into the repo directly. But I'm not there yet.

will-moore:
So, I'm happy for this to be closed (or temporarily moved to draft) while we consider if this is the best approach. This PR can simply serve as an FYI that I'm including these changes into omero-bff and a reference for future discussion.

SeanDuHare:
> Thanks for reviewing. I totally understand the concerns about hardcoding etc, hence my "suggestion / question / discussion". Since it's possible for me to build and bundle any branch/local commit of BFF into omero-biofilefinder, this doesn't _need_ to be in the mainline for me to use it. Having it in the mainline would simply save me from repeatedly rebasing this each time I update BFF in omero-biofilefinder. At the moment it's quite a manual process: I plan to make BFF into a submodule of omero-biofilefinder and setup a build process that would avoid me committing the built assets into the repo directly. But I'm not there yet.

I think its fine to integrate it - my note is really just a notice to any other reviewers that we should consider not doing this going forward.

aswallace (approved):

will-moore:
Thanks for the reviews...

Please just hold off on merging while I give this another round of consideration along with https://github.com/ome/omero-biofilefinder/pull/17.
I just want to think if there are any tweaks/improvements etc I can make, to avoid having to come back to BFF with more changes. Thanks...

will-moore:
OK, just to summarize the current behaviour:

The file extension is defined as any letters following the last `.` *after* the last `/` in the `File Path`.

IF the file extension is empty (e.g. no `.` after the last `/`) *and* the BFF app is hosted at a URL with pathname ending in `/biofilefinder/bff/app/` then we show "Open with > OMERO".

Potential issues:
 - the app host URL changes in `omero-biofilefinder`: I don't see this being a problem. The URL shouldn't need to change - all the parameters we need to pass the app are in the query string rather than the URL.
 - the `File Path` *does* have an extension: E.g. if the `File Path` was `https://my-institute-server.org/optional_prefix/img_detail/123.0` or ended in `/image.ext` then extension wouldn't be "". I also don't see this as a problem: all the URLs we use to objects in OMERO use object IDs rather than names etc, so I'm pretty sure the "extension" will be a empty string. 
 

As an aside, I realise that some Zarr URLs don't always end in `.zarr`. E.g. if your image is nested under a `bioformats2raw` layout such as https://livingobjects.ebi.ac.uk/idr/zarr/v0.4/idr0079A/idr0079_images.zarr/0 then this won't be detected as a `.zarr` extension. For the right-panel zarr-based thumbnail generation, I think this allows `.zarr` anywhere in the File Path?


So, this PR is looking good, but I realise I meant to try the `Column_Description.csv` as described at https://github.com/AllenInstitute/biofile-finder/issues/526#issuecomment-4357109954 to see if that provides better options and to test how it plays with this PR. So let me give that a try...

will-moore:
Having tested the `Column_Description.csv` for specifying Open with behaviour in BFF (https://github.com/ome/omero-biofilefinder/pull/17), this seems to be working quite well, so I think the approach in this PR won't be needed currently. Thanks.

PR #832, opened 2026-06-15
Feature/home page

lynwilhelm:
## Redesigned home page

Replaces the previous placeholder home page with a structured marketing landing page.

### What's in this PR
- **Hero** with tagline and CTA linking to the live example dataset, new image
- **Intro** band with the publication link and quick-start CTAs
- **Why BioFile Finder** — 4 value-prop cards with graphics
- **How does BioFile Finder work** — 2 explanatory cards + flow diagram
- **What would you like to do next** — 3 action cards routing to use cases, user guide, and datasets
- **Engage with us** — links to forum, GitHub, and email

### Architecture
Content (copy, links, card data) lives entirely in `content.ts` so future copy edits are a one-file change with no markup risk. `InfoCard` and `Section` are shared primitives that all grids and bands compose from.

### Non-obvious decisions
- **SVG graphics** — `react-svg-loader` compiles SVGs to React components, not URL strings. `InfoCard` detects the type at render time and renders either `` or the component. `custom.d.ts` was updated to reflect this accurately.
- **Scroll reset** — fixed a pre-existing bug where the app's scroll container (an inner div) wasn't resetting on navigation since React Router only resets `window.scrollY`.
- **Static content** — no loading, empty, or error states; this page has no data dependencies.

### Tradeoffs
- Back-button navigation always scrolls to top rather than restoring position. Acceptable for now; native scroll restoration would require more investment.

### Testing
- 718 tests passing, 0 failing across core and web suites
- Desktop suite has a pre-existing `getPath` error unrelated to this PR
- Verified at mobile (375px), tablet (750px, 999px), and desktop (1200px+) widths
- Checked all CTAs route correctly
- _Add screenshots before merging_

### Known follow-ups
- In-app tutorial trigger on entry from nav bar (separate issue)


BrianWhitneyAI:
should this be closed?

packages/core/components/Buttons/BaseButton.tsx
```diff
@@ -8,12 +8,22 @@ import Tooltip from "../Tooltip";
 import styles from "./BaseButton.module.css";
 
 interface Props {
+    // Accessible name when no tooltip is wanted. `title` also sets the
```
SeanDuHare, line 11: This comment can be removed (common dev info / verbose)

packages/core/components/Buttons/BaseButton.tsx
```diff
@@ -34,17 +44,23 @@ export default function BaseButton(props: Props) {
         directionalHint: props.menuDirection,
     });
 
+    const iconAfter = props.iconPosition === "after";
```
SeanDuHare, line 47: (nit) boolean language
```suggestion
    const isIconAfter = props.iconPosition === "after";
```

packages/core/components/Buttons/TertiaryButton.tsx
```diff
@@ -17,7 +18,9 @@ interface Props {
     onClick?: () => void;
     invertColor?: boolean;
     text?: string;
-    title: string;
+    // Shows a tooltip and sets the accessible name. Use `ariaLabel` instead to
```
SeanDuHare, line 21: IMO this comment should move to BaseButton since the prop definitions seem to be gathering in there (or just removed)

packages/web/src/components/Header/Menu.tsx
```diff
@@ -21,6 +22,11 @@ export default function Menu() {
         items: helpMenuOptions,
         directionalHint: DirectionalHint.bottomAutoEdge,
     });
+    // Submenu for the mobile menu's "Launch app" item, styled like the Help submenu.
+    const launchMenu = useButtonMenu({
+        items: LAUNCH_APP_MENU_ITEMS,
+        directionalHint: DirectionalHint.bottomAutoEdge,
```
SeanDuHare, line 28: Bottom left edge might be better since its in the top right corner of the app always but this is probably fine

packages/web/src/components/Home/sections/Hero.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * call-to-action that launches the app pre-loaded with an example dataset.
+ */
+export default function Hero() {
+    return (
+        <section className={styles.hero} aria-label={`${APPLICATION_NAME} introduction`}>
+            <div className={styles.heroContent}>
+                <h1 className={styles.heroTitle}>{APPLICATION_NAME}</h1>
+                <p className={styles.heroTagline}>Find, explore, and share data—faster.</p>
+                {/*
+                    Navigate via onClick (rather than wrapping in an <a>) so the
```
SeanDuHare, line 20: This is kind of a strange workaround but idk that it really matters in practice so I guess its fine 🤷‍♂️

packages/web/src/components/Home/sections/EngageWithUs.tsx
```diff
@@ -0,0 +1,45 @@
...
+    return (
+        <Section
+            id="engage"
+            title="Engage with us"
+            alt
+            bandClassName={styles.bandFooterClearance}
+            className={styles.centeredText}
+        >
+            <div className={styles.engageLinks}>
+                {ENGAGE_LINKS.map((link) => {
```
SeanDuHare, line 23: Do all of these links need to be defined elsewhere rather than here?

packages/web/src/components/Home/sections/Intro.tsx
```diff
@@ -0,0 +1,52 @@
+import classNames from "classnames";
+import * as React from "react";
+import { Link } from "react-router-dom";
+
+import { SecondaryButton } from "../../../../../core/components/Buttons";
+import { APPLICATION_NAME } from "../../../constants";
+import SiteLogo from "../../../../assets/site-logo.png";
+import Section from "../Section";
+import { LINKS } from "../content";
```
SeanDuHare, line 9: (nit) The imports in several of these files are ordered in reverse (the closest relative imports should be on top and the furthest relative imports on bottom)

packages/web/src/components/Home/content.ts
```diff
@@ -0,0 +1,156 @@
...
+
+/** Route/URL targets used across the page. */
+export const LINKS = {
+    app: "/app",
+    datasets: "/datasets",
+    tryNow: EXAMPLE_DATASET_URL,
+    userGuide: "/user-guide",
+    userGuideSetupOverview: "/user-guide/getting-started/setup-overview",
+    userGuideUseCases: "/user-guide/real-world-use-cases/use-cases",
+    // External resources (open in a new tab).
```
SeanDuHare, line 22: Do these need to be treated differently? As in, should the internal facing links use the react-router link while these external ones use <a />?

packages/web/src/components/Home/content.ts
```diff
@@ -0,0 +1,156 @@
...
+    {
+        accent: "Integrates with existing tools",
+        heading: "",
+        body:
+            "BioFile Finder serves as a lightweight entry point, connecting metadata exploration with the applications you already use.",
+    },
+];
+
+/** Flow diagram rendered at the bottom of the How It Works section. */
+export const HOW_GRAPHIC = HowBFFWorksGraphic;
```
SeanDuHare, line 115: (nit) A little odd to export a constant this way rather than just exporting the `HowBFFWorksGraphic` import directly

packages/web/src/components/Home/Home.module.css
SeanDuHare, line 1: Combining all of the CSS for all of the different components into this file makes this a bit hard to read (rather than a component per file). Do they share a lot of the same classes?

packages/web/src/components/Home/InfoCard.module.css
```diff
@@ -0,0 +1,54 @@
...
+.heading {
+    /* Lighter than the global h3 weight to match the marketing layout. */
+    color: var(--secondary-text-color);
+    font-size: 20px;
+    font-weight: 400;
+    line-height: 1.4;
+}
+
+/* Emphasized lead-in phrase; weight and color match the heading. */
+.accent {
```
SeanDuHare, line 40: empty class

packages/web/src/components/Home/content.ts
```diff
@@ -0,0 +1,156 @@
...
+
+// ---------------------------------------------------------------------------
+// What Would You Like to Do Next?
+// ---------------------------------------------------------------------------
+
+export interface ActionCard {
+    title: string;
+    body: string;
+    cta: { text: string; href: string; external?: boolean };
```
SeanDuHare, line 124: `external` seems to be unused (or at least I'm having trouble finding where it is used)

packages/web/custom.d.ts
SeanDuHare, line 1: There are SVG components elsewhere in the app - make sure you poke around to see if they still work after changing the import style

packages/web/src/constants/index.ts
```diff
@@ -1 +1,7 @@
 export const APPLICATION_NAME = "BioFile Finder";
+
+// Deep link into the deployed app pre-loaded with a public example dataset
+// (Drug Perturbation Dataset). Shared by the home page hero and the
+// "Launch app" dropdown so there is a single source of truth.
+export const EXAMPLE_DATASET_URL =
```
SeanDuHare, line 6: I think this can move to the `content` file you have

packages/web/src/components/Layout/index.tsx
```diff
@@ -14,12 +14,25 @@ export default function Layout() {
...
+    }, [currentPath]);
+
     return (
         <div className={styles.root}>
             <Header />
             <div
+                ref={scrollRef}
                 className={classNames(isApp ? styles.fillScreen : styles.scrollable, {
-                    [styles.footerClearance]: !isApp && !isUserGuide,
+                    [styles.footerClearance]: !isApp && !isUserGuide && !isHome,
```
SeanDuHare, line 35: Is the clearance only present for the dataset page now?

SeanDuHare (commented):
Image

This section looks a bit funky on mobile

packages/core/components/Buttons/BaseButton.tsx
```diff
@@ -8,12 +8,22 @@ import Tooltip from "../Tooltip";
...
 
 interface Props {
+    // Accessible name when no tooltip is wanted. `title` also sets the
+    // accessible name but additionally shows a tooltip; use `ariaLabel` to
+    // name the button (e.g. an icon-only button) without a tooltip.
+    ariaLabel?: string;
     className?: string;
     disabled?: boolean;
     iconName?: string;
+    // Which side of the text the icon sits on. Defaults to "before" (the icon
```
BrianWhitneyAI, line 18: also overly verbose. This is what dan was talking about in todays ai lunch and learn

packages/core/components/Buttons/BaseButton.tsx
```diff
@@ -8,12 +8,22 @@ import Tooltip from "../Tooltip";
...
+    // leads the text), preserving existing call sites.
+    iconPosition?: "before" | "after";
     id?: string;
     isSelected?: boolean;
     menuDirection?: DirectionalHint;
+    // Optional Fluent icon name for the trailing menu chevron. When omitted the
```
BrianWhitneyAI, line 24: also overly verbose. This is what dan was talking about in todays ai lunch and learn

packages/core/components/Buttons/BaseButton.tsx
```diff
@@ -67,10 +83,15 @@ export default function BaseButton(props: Props) {
...
                 })}
                 data-testid={`base-button-${props.id}`}
-                ariaLabel={props.title}
+                ariaLabel={props.ariaLabel ?? props.title}
                 disabled={props.disabled}
                 id={props.id}
-                menuIconProps={{ className: styles.hidden }}
+                menuIconProps={
+                    props.menuIconName
+                        ? // Match the weight/styling of the button's leading/trailing icons.
```
BrianWhitneyAI, line 91: not a useful comment

packages/web/src/components/Header/Menu.tsx
```diff
@@ -21,6 +22,11 @@ export default function Menu() {
         items: helpMenuOptions,
         directionalHint: DirectionalHint.bottomAutoEdge,
     });
+    // Submenu for the mobile menu's "Launch app" item, styled like the Help submenu.
```
BrianWhitneyAI, line 25: usually for comments try to to say " done like X" unless you reference the class/code directly

packages/web/src/components/Home/sections/Hero.tsx
```diff
@@ -0,0 +1,40 @@
...
+                    button's tooltip host wraps the button directly. The
+                    inline-block wrapper shrinks the (block) tooltip host to the
+                    button's width so the tooltip anchors to the button instead
+                    of centering on the full-width hero band. `title` drives the
+                    tooltip.
```
BrianWhitneyAI, line 25: verbosity, 'what does this need to tell a dev"

packages/web/src/components/Home/sections/Hero.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { APPLICATION_NAME } from "../../../constants";
+import { LINKS } from "../content";
+
+import styles from "../Home.module.css";
+
+/**
+ * Full-bleed hero banner with the page's single H1, tagline, and primary
+ * call-to-action that launches the app pre-loaded with an example dataset.
+ */
+export default function Hero() {
```
BrianWhitneyAI, line 13: What is Hero? (besides a banner)

packages/web/src/constants/index.ts
BrianWhitneyAI, line 6: Would this ever have use on stg or local host? maybe we dont hardcode PROD

packages/web/src/components/LaunchAppMenu/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Defined once here and shared by the desktop dropdown (below), the home page
+ * "How do I get started?" section, and the Header's mobile menu submenu so the
+ * entry points stay in exactly one place.
+ */
+export const LAUNCH_APP_MENU_ITEMS: IContextualMenuItem[] = [
+    {
+        key: "your-data",
+        text: "Start with your data",
+        href: "/app",
```
BrianWhitneyAI, line 20: "Start with your data" now points at internal route /app with target: "_blank" instead of the previous
  in-SPA <Link to="app"> — costs an extra click and forces a full reload for a route the SPA owns. JSDoc at line
   13 is also stale

packages/web/src/components/Home/PlaceholderImage.tsx
```diff
@@ -0,0 +1,30 @@
+import classNames from "classnames";
```
BrianWhitneyAI, line 1: Unused?

lynwilhelm:
There is a strange thing where as soon as the modify columns flyout appears, things feel jumpy. I can see the search icon goes away so "search" words moves to the left, automatically putting focus on the field. Then the menu has a moment where it animates in sliding left to right for a split second. Maybe it's because stuff is moving in opposite directions it feels off. This is more apparent here on hover than on click (when you click on group of filter). Maybe we could just remove that sliding animation on the flyout globally?

packages/web/src/components/Home/content.ts
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * A card whose heading leads with an emphasized (accent-colored) phrase
+ * followed by the rest of the heading, then a description.
+ * Used by the "Why" and "How it works" grids.
+ */
+export interface AccentCard {
+    /** Emphasized lead-in phrase, rendered in the accent color. */
+    accent: string;
+    /** Remainder of the heading, rendered inline after the accent phrase. */
+    heading: string;
```
aswallace, line 38: This currently is not true (unless there are plans to add styling to the accent class in the future), so this separation may not be necessary

packages/web/src/components/Home/content.ts
```diff
@@ -0,0 +1,156 @@
...
+        heading: "and reproducibility",
+        body:
+            "Share persistent, interactive dataset views so collaborators can open the same filtered, organized state and see the exact view you see.",
+        image: { src: PeopleShareGraphic, alt: "People sharing a dataset icon", icon: true },
+    },
+    {
+        accent: "Dynamically organize",
+        heading: "and curate datasets",
+        body:
+            "Filter, group, and restructure collections using hierarchical views—without modifying underlying metadata or breaking existing reference.",
```
aswallace: very minor thing that is totally fine if intentional but did catch my attention in scanning for AI flags-
Both this and the "reduce time" one use the exact same LLM-style sentence pattern: "X, y, and z a thing--without doing this other thing."

aswallace (commented):
The new page looks great! A couple minor comments, otherwise seconding Sean & Brian's comments.

packages/core/components/Buttons/BaseButton.tsx
lynwilhelm: Removed. 4a3ea95b or 1a658557

packages/core/components/Buttons/BaseButton.tsx
lynwilhelm: 1a658557

packages/core/components/Buttons/BaseButton.tsx
lynwilhelm: 1a658557

packages/core/components/Buttons/BaseButton.tsx
lynwilhelm: Thanks, updated to your suggestion. 1560ed537

packages/core/components/Buttons/BaseButton.tsx
lynwilhelm: b26b0858

packages/core/components/Buttons/TertiaryButton.tsx
lynwilhelm: Removed. 8be088d9

packages/web/src/components/Header/Menu.tsx
lynwilhelm: Added class name to comment. Got it, thx! cb33fe36

packages/web/src/components/Header/Menu.tsx
lynwilhelm: I think it's ok where it is:)

packages/web/src/components/Home/sections/EngageWithUs.tsx
lynwilhelm: A it makes more sense to just have links local in one spot since they are only used once. Updated to reflect that. 721e6a02

packages/web/src/components/Home/sections/Hero.tsx
lynwilhelm: A hero is typically on a homepage and may be more complex than just a simple banner on an interior page, that's often just a graphic that is not very tall. I take your point though.

packages/web/src/components/Home/sections/Hero.tsx
lynwilhelm: Left as-is, but honestly I don't understand the alternative well.

packages/web/src/components/Home/sections/Hero.tsx
lynwilhelm: Removed the comment

packages/web/src/components/Home/sections/Intro.tsx
lynwilhelm: Resolved 0a880b57

packages/web/src/components/Home/content.ts
lynwilhelm: Seems to be a pretty normal and reasonable approach to me

React Router Link is better for in-app routes because it:
- preserves SPA navigation
- avoids full page reloads
- keeps route state/history behavior consistent

Anchors are better for:
- external URLs
- new-tab destinations
- mailto / docs / publication links
- anything not handled by the app router

packages/web/src/components/Home/content.ts
lynwilhelm: I don't fully understand what you mean but I tried to understand and Claude is telling me that we agree and to make no change.

packages/web/src/components/Home/content.ts
lynwilhelm: ok, I removed the unnecessary alias and kept the asset export direct 8b2bc1c6

packages/web/src/components/Home/content.ts
lynwilhelm: Cool. Removed the external flag and derived the link type from the href instead.

packages/web/src/components/Home/Home.module.css
lynwilhelm: I separated the hero-specific styles into its own module so the shared home stylesheet is less crowded and more readable. 07cc6dee

packages/web/src/components/Home/PlaceholderImage.tsx
lynwilhelm: Thx, removed! 9b61a14e

packages/web/src/components/LaunchAppMenu/index.tsx
lynwilhelm: Changed so that "Start with your data" now uses target: "_self" but the second option "Explore example dataset" opens in a new tab (for reason that it's more of a start experience flow and users may likely want to explore quickly initially but hen return to the homepage). d9d96b9a

packages/web/src/components/Layout/index.tsx
lynwilhelm: Cool, I simplified the footer clearance logic. 964e6507

packages/web/src/constants/index.ts
lynwilhelm: Done

lynwilhelm: Cool, hardcode removed

packages/web/custom.d.ts
lynwilhelm: Checked and did not see anything!

packages/web/src/components/Home/InfoCard.module.css
lynwilhelm: Removed the empty .accent class and kept the accent text rendering directly in the heading. Commit: e3674a18

packages/web/src/components/Home/content.ts
SeanDuHare, line 1: I feel like some of these could be moved closer to where they are used since they don't seem to be re-used

packages/web/src/components/Home/sections/Intro.tsx
SeanDuHare: The css import for a few of these files is now within this import block - it is usually as its own separate 4th block since it is rather special

SeanDuHare (approved):

BrianWhitneyAI (approved):

packages/web/src/components/Home/content.ts
aswallace: I just meant that the comments here and the fact that `accent` and `heading` are split up for each of these cards implies that the `accent` lead-in phrase (e.g., "Reduce time" in the screenshot) would be the accent color, aka, aqua. But there isn't actually any special styling for the `accent` class in `InfoCard.module.css` or `InfoCard.tsx`, so it's the same grey across the whole heading. This means it's not really necessary to split up each of these cards like 
```
        accent: "Reduce time",
        heading: "to find the right files",
```
But this is a little nitpicky and it also makes sense to leave it if you think we'll eventually want to style those accent phrases differently.

image

aswallace (approved):

packages/web/src/components/Home/content.ts
lynwilhelm: Oh, I totally get it now! Yes, I originally had them styled differently and then removed it. This is residual code and I just removed it. Thanks!

lynwilhelm: Simplified the card data to use a single heading field and removed the redundant accent field. The visible headings are unchanged. Commit: fd1a192d

PR #851, opened 2026-06-18
Nested Metadata

SeanDuHare:
##  Description

This PR extends BioFile Finder to query, display, filter, sort, group, and export files whose metadata is stored in DuckDB `STRUCT[]` columns (e.g. `Well STRUCT(Dose STRUCT(Unit VARCHAR, Value DOUBLE)[])[]`). Sub-field annotations like `"Well.Dose.Unit"` are surfaced alongside flat annotations throughout the app — in the file list, file detail panel, query sidebar (filter/sort/group), annotation picker, and manifest download.

### Schema detection (`packages/core/services/DatabaseService/`)

`parseStructFields(dataType)` recursively walks a DuckDB `STRUCT(...)[]` type string, producing flat dotted-name field entries with `intermediateIsArray` flags. Used to generate sub-field `Annotation` objects with correct `path` and `pathIsArray` at data-source load time.

### Query sidebar UI (`packages/core/components/QueryPart/`)

`QueryPartRow` / `useSizeSpecificPrefix` — breadcrumb display for nested paths in Group/Sort/Filter rows. Truncation priority: full path → `… / Parent / Leaf` (fitting whole trailing parents) → leaf-only. Uses a `ResizeObserver` deferred via `requestAnimationFrame` to avoid the "ResizeObserver loop" browser warning. Row `maxWidth` now accounts for indentation depth so the × button never overflows.

### File detail panel (`packages/core/components/FileDetails/MetadataList/`)

Complete replacement of `FileAnnotationList`/`FileAnnotationRow` with a new `MetadataList` component tree (`index`, `Row`, `Section`, `Value`, `ContentLengthToggle`, `useDisplayText`). Nested parent annotations show an expandable `Section` with child rows for each sub-field entry.

### Components
1. **`packages/core/components/QueryPart/useSizeSpecificPrefix.ts`** + **`QueryPartRow.tsx`**
   Breadcrumb measurement (probe spans, `ResizeObserver` + `rAF`, `available = rowTitle.clientWidth - iconWidth`). The hidden-probe pattern is subtle.

2. **`packages/core/components/FileDetails/MetadataList/`**
   New component tree — particularly `Row.tsx` (nested vs flat rendering), `useDisplayText.ts` (async `formatPathForHost`, S3 URL rewriting, local path), and `Section.tsx` (expand/collapse).

### ...state management / data wrangling
1. **`packages/core/entity/SQLBuilder/index.ts`**
   `buildNestedAccessExpression` — the core recursive expression builder every nested SQL path goes through. Review the `varIdx`/`flattenCount` recursion and `flatten()` wrapping logic.

2. **`packages/core/entity/FileFilter/index.ts`**
   `toSQLWhereString` nested branch, `toCorrelatedSQLWhereString`, `buildElementCondition`, and SQL injection sanitization in `parseRangeBounds`. The correlated filter logic (OR within sub-field group, AND across groups, fallback for double-array nesting) is the most complex new behavior.

3. **`packages/core/services/DatabaseService/index.ts`**
   `parseStructFields` recursive STRUCT type parsing and the annotation generation loop (~line 1280) that assembles `pathIsArray = [rootIsArray, ...field.intermediateIsArray]`.

## Follow-up tasks

Several TODOs were added throughout to avoid changing even more - examples include avoiding using "." concatenation as a key for annotation name paths.

## Testing
I've been doing a lot of manual testing throughout and testing performance of various SQL routes that I'm happy to chat about, an easy fairly comparable to what we will likely do in FMS is [this one](https://staging-biofile-finder-datasets.s3.us-west-2.amazonaws.com/simple+nested+demo.parquet) or this one for the [Drug dataset](https://staging-biofile-finder-datasets.s3.us-west-2.amazonaws.com/Drug%2BPerturbation%2BDataset.parquet)

If you'd like other files to test with let me know, I have some with millions of rows for example.


packages/core/state/metadata/logics.ts
```diff
@@ -127,29 +126,54 @@ const receiveAnnotationsLogic = createLogic({
...
+        // columns once available. This intentionally does not block the dispatches above so the
+        // file list renders immediately at the default width.
+        const widthByAnnotation = await annotationService.fetchOptimalWidthForAnnotations(
+            annotations
+        );
+        const resizedColumns = columns.map((column) => {
+            const width = widthByAnnotation.get(column.name);
+            return width === undefined ? column : { ...column, width };
+        });
+        dispatch(selection.actions.setColumns(resizedColumns));
```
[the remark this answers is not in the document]
SeanDuHare: Adjust in [ab196f](https://github.com/AllenInstitute/biofile-finder/pull/851/commits/6ab196fd1443e163ef9871c54694c455222315ff) by avoiding the two setColumns calls altogether. Because the fetch for the auto calculated widths is time-boxed it doesn't seem important to really try to save time by having the two calls since it causes the UI to re-render majorly

packages/core/entity/SearchParams/test/searchparams.test.ts
```diff
@@ -100,15 +106,16 @@ describe("SearchParams", () => {
...
-             * group=Cell+Line&filter={"name":"file_name","value":"testname.csv","type":"default"}
-             * &filter={"name":"file_path","value":"","type":"fuzzy"}
-             * &filter={"name":"Gene","value":"","type":"exclude"}
-             * &filter={"name":"Cell+Line","value":"","type":"include"}
+            /** URL decodes to (filters/sort now encode a `path` array rather than `name`):
+             * group=Cell+Line&group=Well.Dose.Solution.Name
+             * &filter={"path":["file_name"],"value":"testname.csv","type":"default"}
+             * &filter={"path":["file_path"],"value":"","type":"fuzzy"}
+             * &filter={"path":["Gene"],"value":"","type":"exclude"}
+             * &filter={"path":["Cell+Line"],"value":"","type":"include"}
```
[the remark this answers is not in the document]
SeanDuHare: Fixed in [ab196f](https://github.com/AllenInstitute/biofile-finder/pull/851/commits/6ab196fd1443e163ef9871c54694c455222315ff)

packages/core/entity/SearchParams/test/searchparams.test.ts
```diff
@@ -53,22 +53,28 @@ describe("SearchParams", () => {
...
             );
-            /** URL decodes to:
+            /** URL decodes to (filters/sort now encode a `path` array rather than `name`):
              * group=Cell+Line&group=Donor+Plasmid&group=Lifting?
-             * &filter={"name":"Cas9","value":"spCas9","type":"default"}
-             * &filter={"name":"Donor+Plasmid","value":"ACTB-mEGFP","type":"default"}
+             * &filter={"path":["Cas9"],"value":"spCas9","type":"default"}
+             * &filter={"path":["Donor+Plasmid"],"value":"ACTB-mEGFP","type":"default"}
              * &openFolder=["AICS-0"]&openFolder=["AICS-0","ACTB-mEGFP"]
```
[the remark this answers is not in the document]
SeanDuHare: Fixed in [ab196f](https://github.com/AllenInstitute/biofile-finder/pull/851/commits/6ab196fd1443e163ef9871c54694c455222315ff)

packages/core/state/metadata/logics.ts
```diff
@@ -77,39 +77,38 @@ const receiveAnnotationsLogic = createLogic({
...
         if (!columnsThatStillExist.length) {
             // Remove filename annotations from columns before re-adding it at the front,
-            columns = columns.filter((column) => column.name !== AnnotationName.FILE_NAME);
+            columns = columns.filter((column) => column.name !== "File Name");
 
             // Add "File Name" back to the front of the columns array
             columns.unshift({
-                name: AnnotationName.FILE_NAME,
-                width: widthByAnnotation.get(AnnotationName.FILE_NAME) ?? DEFAULT_COLUMN_WIDTH,
+                name: "File Name",
```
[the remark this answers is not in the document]
SeanDuHare: This was intentional, but I can see it being confusing so I changed the columns to use .name consistently see [ab196f](https://github.com/AllenInstitute/biofile-finder/pull/851/commits/6ab196fd1443e163ef9871c54694c455222315ff)

packages/core/components/FileDetails/MetadataList/Row.tsx
```diff
@@ -0,0 +1,193 @@
...
+        dispatch(
+            interaction.actions.processError(
+                `<Row />-${props.name}`,
+                `Unexpected internal error. Unable to find column metadata for field "${path.join(
+                    " : "
+                )}". Omitting this field from display.`
+            )
+        );
+        return null;
+    }
```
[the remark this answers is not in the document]
SeanDuHare: Fixed in [ab196f](https://github.com/AllenInstitute/biofile-finder/pull/851/commits/6ab196fd1443e163ef9871c54694c455222315ff)

packages/core/entity/FileDetail/index.ts
```diff
@@ -153,44 +230,65 @@ export default class FileDetail {
...
+        // 2. For each subsequent path segment, drill into matching sub-fields
+        const rootValue = this.metadata.get(path[0]);
+        if (!rootValue) return undefined;
 
-    public getFirstAnnotationValue(annotationName: string): string | number | boolean | undefined {
-        return this.getAnnotation(annotationName)?.values[0];
+        const values = FileDetail.extractNestedValues(rootValue, path.slice(1));
+        return isObject(values) && !isNil(values)
+            ? uniq(values as PrimitiveMetadataValue[])
+            : (values as NestedMetadataValue[] | undefined);
```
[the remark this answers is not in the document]
SeanDuHare: Fixed in [ab196f](https://github.com/AllenInstitute/biofile-finder/pull/851/commits/6ab196fd1443e163ef9871c54694c455222315ff)

packages/core/components/AnnotationFilterForm/test/AnnotationFilterForm.test.tsx
```diff
@@ -242,21 +242,21 @@ describe("<AnnotationFilterForm />", () => {
                 </Provider>
             );
             await waitFor(
-                () => expect(getByTestId(`${LISTROW_TESTID_PREFIX}False`)).to.not.be.undefined
+                () => expect(getByTestId(`${LISTROW_TESTID_PREFIX}false`)).to.not.be.undefined
```
BrianWhitneyAI, line 245: what does going to lowercase do here?

packages/core/components/AnnotationPicker/index.tsx
```diff
@@ -36,6 +36,11 @@ interface Props {
 /**
  * Form for selecting which annotations to use in some exterior context like
  * downloading a manifest.
+ *
+ * Nested sub-field annotations (e.g. "Well.Gene", "Well.Dose.Unit") are listed under their
+ * top-level parent, with each leaf showing its full ancestry as breadcrumbs (e.g.
```
BrianWhitneyAI, line 41: this analogy is hard for me to understand

packages/core/components/AnnotationPicker/index.tsx
```diff
@@ -67,22 +68,46 @@ export default function AnnotationPicker(props: Props) {
             data: annotation,
             value: annotation.name,
             description: annotation.description,
-            displayValue: annotation.displayName,
+            displayValue: annotation.displayName.split(".").slice(-1)[0],
```
BrianWhitneyAI, line 71: so this is taking the last value of the annotation in X.Y.Z so "Z" for "Well.Dose.Unit" does that make Unit the display name?

packages/core/components/AnnotationPicker/index.tsx
```diff
@@ -67,22 +68,46 @@ export default function AnnotationPicker(props: Props) {
...
             recent: recentAnnotationNames.includes(annotation.name) && !selected,
             loading: props.disableUnavailableAnnotations && areAvailableAnnotationLoading,
+            breadcrumbs: annotation.path.length > 1 ? annotation.path.slice(0, -1) : undefined,
```
BrianWhitneyAI, line 74: this is really confusing to me. Breadcrumbs is boolean or a substring or undefined?

packages/core/components/AnnotationPicker/index.tsx
```diff
@@ -67,22 +68,46 @@ export default function AnnotationPicker(props: Props) {
...
+    const nonUniqueItems = annotations
+        // Annotations must be selectedable and not have nested children
+        .filter((a) => isSelectable(a) && !a.isParent)
+        // Sort recent annotations to the top then sort by alphabetically
+        .sort((a, b) => {
+            // Check if annotation is more or less recent than the other
+            const aIsRecent = recentAnnotationNames.includes(a.name);
+            const bIsRecent = recentAnnotationNames.includes(b.name);
+            const isUsedMoreRecent = aIsRecent && !bIsRecent;
+            const isUsedLessRecent = bIsRecent && !aIsRecent;
```
BrianWhitneyAI: nit "Most" and "Least"?

packages/core/components/FileDetails/MetadataList/test/MetadataList.test.tsx
```diff
@@ -4,34 +4,34 @@ import { expect } from "chai";
...
 
-describe("<FileAnnotationList />", () => {
+describe("<MetadataList />", () => {
     describe("file path representation", () => {
         const annotations = [
             ...TOP_LEVEL_FILE_ANNOTATIONS,
             new Annotation({
-                annotationDisplayName: "Cache Eviction Date",
-                annotationName: AnnotationName.CACHE_EVICTION_DATE,
+                annotationDisplayName: AnnotationName.CACHE_EVICTION_DATE,
```
BrianWhitneyAI, line 21: Are these formatted correctly?

packages/core/components/FileDetails/MetadataList/test/Section.test.tsx
```diff
@@ -0,0 +1,100 @@
+import { render, fireEvent } from "@testing-library/react";
+import { expect } from "chai";
+import * as React from "react";
+
+import Section from "../Section";
+import { MetadataValue } from "../../../../services/FileService";
+import sinon from "sinon";
+
+const ChildRow = ({ name, value }: { name: string; value: MetadataValue }) => (
```
BrianWhitneyAI, line 11: this could use a comment

packages/core/state/metadata/logics.ts
```diff
@@ -286,14 +291,16 @@ const storeNewAnnotationLogic = createLogic({
...
+                ([_type, id]) => id === annotation.annotationTypeId
+            )?.[0] as AnnotationType) || AnnotationType.STRING;
         const newMmsAnnotation = new Annotation({
+            type,
             annotationName: annotation.name,
             annotationDisplayName: annotation.name,
             annotationId: annotation.annotationId,
             description: annotation.description,
-            type: type as AnnotationType,
+            pathIsArray: [false],
```
BrianWhitneyAI, line 303: this is strange formatting to me

packages/core/state/metadata/logics.ts
```diff
@@ -127,29 +89,72 @@ const receiveAnnotationsLogic = createLogic({
...
+                const { annotation: annotationB, currentColumnIndex: indexB } = b;
+                // Move columns that were already selected for display to the front, in their existing order
+                if (indexA !== -1 && indexB !== -1) {
+                    return indexA - indexB;
+                }
+                if (indexA !== -1) return -1;
+                if (indexB !== -1) return 1;
+
+                // Move "File Name" to the front then sort alphabetically
+                if (annotationA.displayName === "File Name") return -1;
```
BrianWhitneyAI, line 138: could we use a real ref to file names display name here?

packages/core/state/interaction/logics.ts
```diff
@@ -876,10 +872,8 @@ const copyFilesLogic = createLogic({
                     status === "DOWNLOAD_STARTED"
                 ) {
                     const file = fileDetails.find((f) => f.id === fileId);
-                    const isAlreadyCached = file?.annotations.some(
-                        ({ name, values }) =>
-                            name === "Should Be in Local Cache" && values[0] === true
-                    );
+                    const isAlreadyCached =
+                        file?.getFirstAnnotationValue(AnnotationName.SHOULD_BE_IN_LOCAL) === true;
```
BrianWhitneyAI, line 880: couldnt we use the actual value rather than asserting equal to true

packages/core/entity/SQLBuilder/index.ts
```diff
@@ -21,12 +23,280 @@ export default class SQLBuilder {
...
+    }
+
+    /**
+     * Build a WHERE-clause expression that tests whether a DuckDB JSON array expression
+     * contains a specific value.
+     *
+     * `json_extract("Well"::JSON, '$[*].Gene')` returns a JSON array like `["Chroma4","Chroma5"]`.
+     * This helper generates a `json_contains` call to check membership.
+     */
+    public static jsonArrayContains(
```
BrianWhitneyAI: Is this unused?

BrianWhitneyAI (commented):
This is alot of code to review, even if a good portion is just moving stuff around theres over 3000 lines of new stuff, I can flag small logical issues I see from scanning the diff but I am having trouble understanding the changes holistically. Some demo videos of the effects of this change might be super useful

SeanDuHare:
> This is alot of code to review, even if a good portion is just moving stuff around theres over 3000 lines of new stuff, I can flag small logical issues I see from scanning the diff but I am having trouble understanding the changes holistically. Some demo videos of the effects of this change might be super useful

I hear ya! I struggled paring this down further without introducing bugs or opportunity for bugs - happy to make any videos that'll help.

Are you thinking videos of the UI in action or videos of how the code corresponds to the UI (like me showing which code does what in the UI)?

BrianWhitneyAI:
> Are you thinking videos of the UI in action or videos of how the code corresponds to the UI (like me showing which code does what in the UI)?


yeah maybe showcasing changes in the query sidebar and file details separately and tieing them to which file does what. Additionally maybe vids of what happens if something isnt there

packages/core/components/AnnotationFilterForm/test/AnnotationFilterForm.test.tsx
SeanDuHare: This is because of this [line change](https://github.com/AllenInstitute/biofile-finder/pull/851/changes/BASE..45f6ea5cf736d51ab55f4e00b2c0262f1febef31#diff-2c7e78d1e257284069181cbd156c4ad576811fcb6fd4b136dd488dc93649a021R67) which was a) incorrect to begin with and b) now that the displayValue vs value props are more complex given the nesting nature must use value to avoid key collision

packages/core/components/AnnotationPicker/index.tsx
SeanDuHare: How about this?
Nested sub-fields (for example, "Well.Gene" and "Well.Dose.Unit") are organized under their top-level parent. Each leaf field displays its full hierarchy as breadcrumbs (for example, for the value "Well / Dose / Unit", "Unit" is the leaf and "Well / Dose" are the two parent segments). Fields are grouped by each ancestry segment and ordered so that related sub-fields remain adjacent (ex "Well / Dose / Unit" and "Well / Vendor / Location" should be grouped together under "Well").

The point of this is to explain that some of the complexity under this is because the picker is ordering them in a specific way - although the comment doesn't necessarily have to be component level.

packages/core/components/AnnotationPicker/index.tsx
SeanDuHare: Yes! This is annoying and complex, but to avoid re-implementing how annotations get built and handled across the app I had to fit name and displayName into here like so

packages/core/components/AnnotationPicker/index.tsx
SeanDuHare: This is a ternary operator saying that if `annotation.path` (which is an array of strings)  has a length > 1 then the breadcrumbs are the `annotation.path` excluding the leaf (so for `["Well", "Dose", "Unit"]` it would be `["Well", "Dose]`). However if `annotation.path` <= 1 then breadcrumbs is undefined since we don't need breadcrumbs when the path represents just one element because the only value is the leaf.

It would always be either a `string` or `undefined` - never a substring though - the `slice` operation is taking place on an array.

packages/core/components/FileDetails/MetadataList/test/MetadataList.test.tsx
SeanDuHare: Yes! I found this while going through the constants to make sure I wasn't breaking anything comparing displayName and name and found we just had hard-coded a duplicate value.

packages/core/state/metadata/logics.ts
SeanDuHare: The `[false]`? It is technically unnecessary since we only need this variable to track whether segments within a nested structure are arrays and _not_ when there is just a single top level element which will always be the case here. The `[false]` represents that the annotations under this code path are never nested arrays.

packages/core/state/metadata/logics.ts
SeanDuHare: I don't think we have a ref to that - this "File Name" is defined several places across the app unfortunately - I have a follow-up in mind that will remove `displayName` entirely in favor of just using `name`. The reason we have both is an artifact of a feature long since removed.

packages/core/state/interaction/logics.ts
SeanDuHare: I'm not sure what you mean. The `getFirstAnnotationValue` will grab the first element in the `values` property of the given annotation within the file. The reason this line item was included in the changeset is because I removed the `annotations` property from the `FileDetail` entity in favor of `metadata` and this new method `getFirstAnnotationValue` because reaching into an `annotations` array when some of the annotations could be nested would silently yield a misleading value

packages/core/components/FileDetails/MetadataList/test/Section.test.tsx
SeanDuHare: Added in 1dbe9529

packages/core/components/AnnotationPicker/index.tsx
SeanDuHare: Adjusted in 7bad833f

packages/core/entity/SQLBuilder/index.ts
SeanDuHare: oof yes! Thank you, removed in 99fd883c

packages/core/components/AnnotationFilterForm/index.tsx
SeanDuHare, line 1: This file is changed because I refactored how `formatter` is pulled out of `annotation` & because of some **unnecessary** type casting to `AnnotationType` (it is already implicitly that type)

packages/core/components/AnnotationFilterForm/useAnnotationValues.ts
SeanDuHare, line 1: An import was moved - it was moved to avoid circular dependency

packages/core/components/AnnotationPicker/index.tsx
SeanDuHare, line 1: This component was modified to support have nested annotations:
1) It adjusts the order to support ordering within segments
2) It adds additional data necessary for nested annotations like breadcrumbs
3) It prevents select all from selecting parents of nested metadata (JSON blobs)

packages/core/components/DirectoryTree/useDirectoryHierarchy.tsx
SeanDuHare, line 1: Changed to use `getAnnotationNameToAnnotationMap` to avoid bug that would've happened if `.name` was compared using `===`. Also removed an unnecessary explicit type cast.

packages/core/components/AnnotationPicker/index.tsx
BrianWhitneyAI: I may be getting confused just due to this example but would the unit be a reasonable display name? Usually I would think of unit as auxiliary to a annotation

packages/core/components/AnnotationPicker/index.tsx
BrianWhitneyAI: Ok I think I got it! Thanks for the update

packages/core/components/AnnotationPicker/index.tsx
SeanDuHare: The unit property on Annotation right? Nope! We use that when creating a value to display for the annotation (value part of a key/value pair of metadata). 

For example, given a metadata key "Volume" and value "30" with a unit of "mL". The `annotation.displayName` would be "Volume" and then elsewhere we would render the value as "30 mL".

packages/core/components/FileDetails/MetadataList/ContentLengthToggle.module.css
SeanDuHare, line 1: Styling for the paired component - specifically just styles a simple icon to have matching color and hover behavior as elsewhere in the app

packages/core/components/FileDetails/MetadataList/ContentLengthToggle.tsx
SeanDuHare, line 1: Re-useable component used in Section and Row for displaying a clickable icon representing collapse/expand

Image

packages/core/components/FileDetails/MetadataList/index.tsx
SeanDuHare, line 1: Core component of the file details pane (the right side of the app) - renders each metadata key/value pair by breaking the top level of keys into sections based on their nested parent (if any). The actual logic for rendering the section and row is in the sibling files Row.tsx and Section.tsx

packages/core/components/FileDetails/MetadataList/Row.tsx
SeanDuHare, line 1: Renders a row in the metadata list. Will recursively render nested sections of metadata if the `value` provided is a nested object. `Section` then displays the actual label and collapse button for the row.

Image

packages/core/components/FileDetails/MetadataList/Section.tsx
SeanDuHare, line 1: Displays a label and icon for a nested Row.tsx - `SectionEntries` will render arrays of sections if applicable. (for example if an annotation like "Well" is an array of nested objects `SectionEntries` will split those into separate collapsible sections
Image

packages/core/components/FileDetails/MetadataList/SectionEntries.tsx
SeanDuHare, line 1: Renders the entires within an array of nested objects as separate collapsible items
Image

packages/core/components/FileDetails/MetadataList/Value.tsx
SeanDuHare, line 1: This is all just copy/pasted out of what was `<Row />` - behavior is still the same. This renders the value part of the key/values in the metadata list

packages/core/components/FileDetails/index.tsx
SeanDuHare, line 1: This is intended to:
*  remove an extra unnecessary layer of `div` nesting that was causing a visual bug
* Move content into separate component, MetadataList
* Move "View relationship diagram" to top of display per Thao's design

packages/core/components/FileList/Header.tsx
SeanDuHare, line 1: One intended change, track duplicate column leaf names (ex. For columns "Channel.Color" and "Solution.Color" the two leaves would be "Color" which would make the headers appear duplicate when in fact they represent different data) - when present will show breadcrumbs

packages/core/components/FileList/LazilyRenderedRow.tsx
SeanDuHare, line 1: `annotationNameToAnnotationMap` was an object leading to bad type tracking. Now that it is a Map you must use the `get` function

packages/core/components/FileRow/Cell.tsx
SeanDuHare, line 1: Previously the tooltips were causing the content to not truncate on overflow properly - this was made more obvious by nested indentations

packages/core/components/ListPicker/index.tsx
SeanDuHare, line 1: Adds breadcrumbs to fuzzy search considerations so when searching it will include parents in the results. For example, if I search by "Image" when there are three columns 'Color", "Image.Acquisition Date", and "Channel.Imaged By" we want the result to be "Image.Acquisition Date" and "Channel.Imaged By" - previously it would just show "Channel.Imaged By".

packages/core/components/ListPicker/ListRow.tsx
SeanDuHare, line 1: Renders breadcrumbs in tooltip information and display (when applicable) and uses `.value` instead of `.displayValue` to avoid possible collisions

Image

packages/core/components/Modal/CopyFileManifest/index.tsx
SeanDuHare, line 1: Uses `getFirstAnnotationValue` to avoid potential search miss due to nesting when just searching on `.annotations`

packages/core/components/Modal/EditMetadata/useAnnotationValues.ts
SeanDuHare, line 1: Adjusted this hook to avoid including parents of nested values (AKA objects) in annotation value reports

packages/core/components/QueryPart/QueryFilter.tsx
SeanDuHare, line 1: Includes annotation path breadcrumbs in label when applicable

(Edit: specifically just the tooltip label, not the actual label per Thao's design)

Image

packages/core/components/QueryPart/QueryGroup.tsx
SeanDuHare, line 1: Includes annotation path breadcrumbs in label when applicable

Image

packages/core/components/QueryPart/QueryPartRow.tsx
SeanDuHare, line 1: Responsible for rendering the actual visual change `QueryGroup` and `QuerySort` take advantage of to display path information for annotations in the label and tooltip

packages/core/components/QueryPart/QuerySort.tsx
SeanDuHare, line 1: Renders path information in tooltip and label

Image

packages/core/components/QueryPart/useSizeSpecificPrefix.ts
SeanDuHare, line 1: We need some way of conditionally rendering a prefix for a column name (the prefix is the ancestry of the column ex "Well" for "Well.Name"). We can't rely on text or width based overflow because we want the cut off to be precisely on whether an entire segment of the path can fit.

For example, for "Well.Dose.Unit" if "Well.Dose.Unit" all fit, display "Well : Dose : Unit". If just "Dose.Unit" fit, display "Dose : Unit", if just "Unit" fits, display "Unit". Lastly, if only partial text fits then truncate and use ellipsis.

Image

packages/core/components/QuerySidebar/Query.tsx
SeanDuHare, line 1: Simplfy / avoid bugs with nested metadata using name mapping rather than explicit search

packages/web/src/components/OpenSourceDatasets/DatasetTable.tsx
SeanDuHare, line 1: Found a loose equality check - seemed unintentional so I fixed it - moreover likely would cause a bug when "annotationName" is constructed on the fly from the `path` property so probably best this way

packages/web/src/components/OpenSourceDatasets/useDatasetDetails.ts
SeanDuHare, line 1: property name changed (in meaning and name)

packages/web/src/entity/PublicDataset/index.ts
SeanDuHare, line 1: Implements useful getAnnotation methods and adjusts contructor to take in a Map (like the mirror of this entity does in `/core`)

packages/web/src/services/DatabaseServiceWeb/duckdb-worker.worker.ts
```diff
@@ -382,16 +381,10 @@ export default class DatabaseServiceWebWorker extends DatabaseService {
...
-                (row) =>
-                    new Annotation({
-                        annotationName: row["column_name"],
-                        annotationDisplayName: row["column_name"],
-                        description: annotationNameToDescriptionMap[row["column_name"]] || "",
-                        type:
-                            (annotationNameToTypeMap[row["column_name"]] as AnnotationType) ||
-                            DatabaseServiceWebWorker.columnTypeToAnnotationType(row["data_type"]),
-                    })
+            const annotations = DatabaseServiceWebWorker.buildAnnotationsFromRows(
```
SeanDuHare, line 387: The mapping here got more complex and was already duplicated so I moved it into this shared static method

packages/core/state/selection/selectors.ts
SeanDuHare, line 1: Simplified this selector to use the `getAnnotationNameToAnnotationMap` selector - the same behavior is expected

packages/core/state/selection/logics.ts
SeanDuHare, line 1: Three intended changes:
* Fix buggy lack of typing which was giving me a headache
* Pass full `annotation` along to fetch width (necessary since columns can be nested now)
* Fix bug where the type wasn't passed into the contructor of the FileFilter

packages/core/state/metadata/selectors.ts
SeanDuHare, line 1: Remove unused selector and switch selector from object to map (map provides better typing in TS)

packages/core/state/metadata/logics.ts
SeanDuHare, line 1: The way these columns were being filtered out and combined was confusing and about to be even more confusing so I tried to make it simpler almost entirely by just chaining everything in the `columns` initialization and taking advantage of sorting. Also, this had a bug where it was causing the file list to flash (re-render) because the columns array in the state was a different reference because this logic is called twice. I added a TODO to fix that bug, but added a guard against it here where it checks if the columns array is deeply equal before dispatching a change to state.

packages/core/state/interaction/logics.ts
SeanDuHare, line 1: Use `getAnnotation` and `getFirstAnnotationValue` instead of searching directly on the `annotations` array. This would eventually silently(?) cause a bug because the array would contain nested values.

packages/core/services/AnnotationService/index.ts
SeanDuHare, line 1: Moved this export to avoid circular dependency

packages/core/services/AnnotationService/HttpAnnotationService/index.ts
SeanDuHare, line 1: Changes necessary to de-couple the interfaces this needs due to FES from interfaces more useful elsewhere in the app

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -106,58 +107,78 @@ export default class DatabaseAnnotationService implements AnnotationService {
...
 
-        return this.fetchFilteredValuesForAnnotation(hierarchy[path.length], filtersByAnnotation);
+        return this.fetchFilteredValuesForAnnotation(hierarchy[path.length], [
+            ...filters,
+            ...hierarchyAsFilters,
+        ]);
     }
 
     // Given a particular annotation in the hierarchy list, apply filters to the files in that category
     private async fetchFilteredValuesForAnnotation(
```
SeanDuHare, line 134: This is changed in two ways:
1) Has logic added to support nested columns
2) Has logic removed from it into shared `FileFilter.toListOfWhereClauses` to make more DRY (which is especially useful with complex nested filtering)

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -75,7 +76,7 @@ export default class DatabaseAnnotationService implements AnnotationService {
      */
     public async fetchAnnotationDetails(name: string): Promise<AnnotationDetails> {
```
SeanDuHare, line 73: This behavior is the same - I changed this because while reading to see if it needed updating for nesting I was confused why it was overly complex

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -191,35 +227,40 @@ export default class DatabaseAnnotationService implements AnnotationService {
             return this.databaseService.query(sql).promise;
         });
         const results = (await Promise.all(queries)) as QueryResult[][];
-        return results.filter((result) => result.length > 0).map((result) => result[0].column_name);
+        const availablePhysicalColumns = new Set(
```
SeanDuHare: This was added so that children of nested columns weren't always excluded from available annotation list. It doesn't work perfectly since it doesn't consider if each individual child of a nested column would actually provide a value though

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -191,35 +227,40 @@ export default class DatabaseAnnotationService implements AnnotationService {
...
-            // which is fine and we will just cancel it and fall back to default column widths in that case
-            const timeoutPromise = new Promise<never>((_, reject) =>
-                setTimeout(() => {
-                    fetchQuery.cancel?.();
-                    reject(new Error("Timeout fetching annotation values"));
-                }, 1000)
-            );
-            const annotationToLength = await Promise.race([fetchQuery.promise, timeoutPromise]);
-            for (const [annotation, length] of Object.entries(annotationToLength)) {
+            const annotationNameToLength = await this.fetchLengthiestValues(annotations);
```
SeanDuHare, line 300: Moved the timeout and race into `fetchLengthiestValues`

packages/core/services/AnnotationService/DatabaseAnnotationService/index.ts
```diff
@@ -253,39 +289,60 @@ export default class DatabaseAnnotationService implements AnnotationService {
      * Fetch the length of the longest value for each annotation, which can be used to compute optimal column widths in the UI.
      * This is a bit of a hack, but it allows us to avoid fetching all values for an annotation just to compute column widths.
      */
-    private fetchLengthiestValues(
-        annotationNames: string[]
-    ): CancellablePromise<{ [annotation: string]: number }> {
-        if (!this.dataSourceNames.length || annotationNames.length === 0) {
-            return { promise: Promise.resolve({}) };
+    private async fetchLengthiestValues(annotations: Annotation[]): Promise<Map<string, number>> {
```
SeanDuHare, line 333: It takes longer to do this for some nested columns (from sub-second to several seconds) so a timeout was brought in here to first try to get the actual "lengthiest values" from the entire set of data and if not possible will try to get the "lengthiest values" from a small subset of the data and if still not possible will just give up

packages/core/entity/resolvePathIsArray.ts
SeanDuHare, line 1: See docstring at top of file

packages/core/entity/SQLBuilder/index.ts
SeanDuHare, line 1: Added several new SQL expressions necessary for querying nested data. They could remain closer to where they are used, but it felt nicer to read/understand IMO to have them abstracted out

packages/core/entity/Graph/index.ts
SeanDuHare, line 1: Changes are due to `file.getAnnotation` not providing an FMSFileAnnotation interface but rather just the values that were within (which was necessary to make returning nested data more readable)

packages/core/entity/FileSort/index.ts
SeanDuHare, line 1: Modified to support paths to columns (ex. Well.Dose.Unit)

packages/core/entity/FileSet/index.ts
SeanDuHare, line 1: Changed to take advantage of SQL building in FileFilter and FileSort and to remove unnecessary splitting of filters by type

packages/core/entity/SearchParams/index.ts
SeanDuHare, line 1: Modified to support `.path` being found in parameters

packages/core/services/FileService/HttpFileService/index.ts
SeanDuHare, line 1: Modified to let the interfaces within the rest of the app drift from what is required in FES (such as is the case with nesting)

packages/core/services/FileService/index.ts
SeanDuHare, line 1: Support nesting in type interfaces

packages/core/hooks/useOpenWithMenuItems/index.tsx
SeanDuHare, line 1: annotationNameToAnnotationMap is a Map now so how this interacted with it had to change slightly - same intended behavior

packages/core/entity/FileDetail/index.ts
SeanDuHare, line 1: Modified to support nesting and added some helper methods to help prevent issues fetching nested values off this entity

packages/core/entity/AnnotationFormatter/index.ts
SeanDuHare, line 1: Add NESTED type and make typing more useful / narrow

packages/core/entity/Annotation/index.ts
SeanDuHare, line 1: Add nesting support through various helper methods

packages/core/entity/FileFilter/index.ts
SeanDuHare, line 1: Major additions to this entity to support querying on nested filter columns

packages/core/components/FileDetails/index.tsx
```diff
@@ -101,17 +101,33 @@ export default function FileDetails(props: Props) {
...
+                            <div className={styles.leftAlign}>
+                                <Pagination className={styles.pagination} />
+                            </div>
+                            {/* spacing component */}
+                            <div className={styles.gutter}></div>
+                            <div className={styles.buttonRow}>
+                                <div>
+                                    {hasProvenanceSource && (
+                                        <DefaultButton
+                                            className={styles.simpleButton}
```
aswallace, line 117: styling does not seem to be getting applied, I think there's an extra `button` tag in the css

Image

packages/core/components/FileList/Header.module.css
```diff
@@ -55,6 +55,17 @@
   vertical-align: middle;
 }
 
+.header-title-prefix {
+  opacity: 0.55;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  display: inline-block;
+  max-width: 50%;
```
aswallace, line 66: (may be intentional but) this appears to cause it to truncate even when there's space
Image

packages/core/components/ListPicker/ListRow.tsx
```diff
@@ -44,8 +45,15 @@ export default function ListRow(props: Props) {
         return null;
     }
 
+    const breadcrumbs = item.breadcrumbs ? `${item.breadcrumbs.join(" : ")} : ` : undefined;
+    let tooltip: string | undefined;
+    if (item.breadcrumbs && item.description)
+        tooltip = `${breadcrumbs}${item.displayValue}: ${item.description}`;
```
aswallace: minor thought: I wonder if we should use `-` instead of `:` between the display value and description now that `:` is the breadcrumbs divider

packages/core/services/FileService/index.ts
```diff
@@ -1,18 +1,30 @@
...
 import FileSet from "../../entity/FileSet";
+import FileSort from "../../entity/FileSort";
 import { JSONReadyRange } from "../../entity/NumericRange";
 
+export type PrimitiveMetadataValue = string | number | boolean;
+/**
+ * A value within a nested annotation entry. Can be a primitive, a nested object,
+ * or an array of nested entries — supporting arrays-of-objects at any depth.
+ */
+// export type NestedAnnotationValue = FmsFileAnnotationValue | NestedAnnotation | NestedAnnotation[];
```
aswallace: Is this an accidental comment or part of future changes?

aswallace (commented):
Still working on understanding the logic changes, but wanted to make sure to submit a few of the UI side comments

packages/core/services/FileService/index.ts
SeanDuHare: Accidental, removed in 1312d707

packages/core/components/ListPicker/ListRow.tsx
SeanDuHare: oh good point - changed in 893b6bc1

packages/core/components/FileList/Header.module.css
SeanDuHare: Yep! I didn't think this **hopefully** niche case of duplicate column names would happen frequently enough to justify measuring the text width with a hook. It is a bit funky though, so lmk if you think it is worth changing :)

packages/core/components/FileDetails/index.tsx
SeanDuHare: Fixed in a8f24937
Screenshot 2026-06-23 at 11 18 49 AM

packages/core/components/FileList/Header.module.css
aswallace: That makes sense to me!

packages/core/components/QueryPart/QueryPartRow.module.css
```diff
@@ -61,26 +66,71 @@
...
+    white-space: nowrap;
+    height: 100%;
+}
+
+/* Nested leaf last-resort truncation happens from the START ("…umn"), never the end. */
+.title-leaf-nested {
+    direction: rtl;
+}
+
+/* Off-screen container holding width-measurement probes; never affects layout or a11y. */
```
aswallace, line 124: comment only: I wasn't familiar with a11y as a shortening of accessibility :O

packages/core/components/QueryPart/useSizeSpecificPrefix.ts
```diff
@@ -0,0 +1,101 @@
...
+            for (let kept = parents.length; kept >= 0; kept--) {
+                const probe = probeElements.current.get(kept);
+                // +1 tolerance: offsetWidth rounds up.
+                if (probe && probe.offsetWidth + leafWidth <= available + 1) {
+                    setPrefix(buildPrefix(parents, kept));
+                    return;
+                }
+            }
+            // if not even "… : " + leaf fits -> show leaf only
+            setPrefix(null);
```
aswallace, line 81: For QueryFilter only, the breadcrumbs never render even when there's space, possibly from a width or max-width constraint, but I couldn't nail down exactly what was causing it

Image

packages/core/hooks/useOpenWithMenuItems/index.tsx
BrianWhitneyAI: Object.keys() called on a Map returns [] use `Array.from(annotationNameToAnnotationMap.keys()).sort()`

packages/core/entity/SQLBuilder/index.ts
```diff
@@ -21,12 +23,248 @@ export default class SQLBuilder {
...
      */
     public static regexMatchValueInList(
-        column: string,
+        expr: string,
         value: string | boolean | number | null
     ): string {
         // Escape special characters for regex
         const escapedValue = `${value}`.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, "\\$&");
-        return `REGEXP_MATCHES(CAST("${column}" AS VARCHAR), '(,\\s*${escapedValue}\\s*,)|(^\\s*${escapedValue}\\s*,)|(,\\s*${escapedValue}\\s*$)|(^\\s*${escapedValue}\\s*$)') = true`;
+        return `REGEXP_MATCHES(CAST(${expr} AS VARCHAR), '(,\\s*${escapedValue}\\s*,)|(^\\s*${escapedValue}\\s*,)|(,\\s*${escapedValue}\\s*$)|(^\\s*${escapedValue}\\s*$)') = true`;
```
BrianWhitneyAI, line 31: This could use the `.replaceAll("'", "''");` from the others

packages/core/components/FileDetails/MetadataList/ContentLengthToggle.tsx
BrianWhitneyAI: no image but makes sense

packages/core/components/FileList/Header.tsx
BrianWhitneyAI: How common is this with our current schema?

packages/core/components/ListPicker/index.tsx
BrianWhitneyAI: I dont think its a problem but this is the first time I have seen nested annotations with spaces. Anything weird around that?

packages/core/components/Modal/CopyFileManifest/index.tsx
BrianWhitneyAI: Is the boolean logic maintained here? moving to annotation value and equal to false or undefined seems like it could have some hidden implications, would None or "" come up?  I guess Im more worried about going from .some to a direct assertion

packages/core/components/QueryPart/useSizeSpecificPrefix.ts
BrianWhitneyAI: the ... : is weird visually to me

BrianWhitneyAI (commented):
Still parsing through. the comment descriptions on files are super useful

packages/core/entity/SQLBuilder/index.ts
SeanDuHare: The `expr` this receives should be quoted so shouldn't be necessary to escape any characters (ex. "O'Brian" or "Imaging Date" should process properly in SQL) - the values do need to be escaped (as opposed to wrapped in quotes) because they are matching via regex.

packages/core/components/FileDetails/MetadataList/ContentLengthToggle.tsx
SeanDuHare: The image attached is just zoomed in to show the specific element

packages/core/components/FileList/Header.tsx
SeanDuHare: There isn't any specific current schema - it'll depend on the dataset - it shouldn't be too common but is likely to happen occasionally

packages/core/components/ListPicker/index.tsx
SeanDuHare: Nah - you left another comment about escaping characters but in general SQL behaves really well with nearly any character in a column name so long as you wrap it in quotes

packages/core/components/Modal/CopyFileManifest/index.tsx
SeanDuHare: `Should Be in Local Cache` should only ever contain a `boolean[]`or be `undefined` - and in practice should only ever have a single element in the `boolean[]`. Given that this mirrors the same behavior as before.

The first `some` usage (on what was line 124) looked for any annotation with that key and a first value that is _strictly_ `true` - this is intended to be the same in `getFirstAnnotationValue`

The second `some` usage (on what was line 132) look for either:
* The annotation to not exist
* The annotation to exist and have a first element of `false`
This is intended to be the same. `getFirstAnnotationValue` will return `undefined` if the annotation does not exist and will return `false` if the first element is `false` in the array so by checking for those two things we mirror the same behavior.

packages/core/components/QueryPart/useSizeSpecificPrefix.ts
SeanDuHare: This came from Thao - unsure what the alternatives she had thought of were if any. Do you have any alternative ideas? I do think this is a convention for breadcrumbs though.

packages/core/hooks/useOpenWithMenuItems/index.tsx
SeanDuHare: Fixed in 28c36cee

packages/core/components/AnnotationPicker/index.tsx
aswallace: I think I get what you mean, Brian. in this example Sean just gave but reconfigured as a nested struct, we'd have "Volume:Value = 30" and "Volume:Unit = ml" which means that in the app we'd see a display name that's just "Value" and one that's just "Unit", and I agree that seeing Unit on its own might be a little weird

packages/core/entity/FileFilter/index.ts
```diff
@@ -35,124 +24,258 @@ export enum FilterType {
...
+                                resolvePathIsArray(f.name, f.path.length, pathIsArrayByName)
+                            )
+                        )
+                        .join(" OR ")
+                );
+            }
+            if (correlatable.length > 0) {
+                const inner = FileFilter.buildNestedConditions(
+                    correlatable,
+                    "__e0",
```
aswallace: Could this get a short comment or become a variable? I don't think I'd know this was a lambda var outside of the context of the doc string on `buildNestedConditions`, so it made me think it needed to be this value to match `__e0` in `toSimpleWhereClause`

packages/core/entity/FileFilter/index.ts
```diff
@@ -35,124 +24,258 @@ export enum FilterType {
...
+                return `("${this.path[0]}" IS NULL OR len(${listExpr}) = 0)`;
+            if (this.type === FilterType.FUZZY)
+                return SQLBuilder.listRegexContains(listExpr, this.value);
+            return SQLBuilder.listContains(listExpr, this.value);
+        }
+
+        // Otherwise, simplest case, match by a scalar condition on the column (or an IS NULL / IS NOT NULL check for ANY/EXCLUDE)
+        const columnExpr = isNested
+            ? SQLBuilder.buildNestedAccessExpression(this.path, pathIsArray)
+            : `"${this.path[0]}"`;
```
aswallace: I may be reading this wrong, but is there a reason this needs to go down here and not at the beginning of the method? It seems like the middle part of this ternary expression is duplicated by `targetColumnExpr` and `listExpr`  in the previous conditionals, which makes things seem more complicated than they actually are

packages/core/entity/SQLBuilder/index.ts
```diff
@@ -21,12 +23,248 @@ export default class SQLBuilder {
...
+        return `EXTRACT(epoch FROM ${expr})::BIGINT * 1000 = ${ms}`;
+    }
+
+    /**
+     * Build an ORDER BY expression that sorts a nested list by its minimum (ASC) or
+     * maximum (DESC) element: `list_sort(listExpr)[1 | -1] order`.
+     * Used by FileSort to sort files by a nested STRUCT[] sub-field.
+     */
+    public static listSortOrderBy(listExpr: string, order: string): string {
+        const idx = order === "ASC" ? 1 : -1;
```
aswallace, line 102: This feels like probably a silly question on my part, but why is the min index 1 rather than 0 in this case? (edit: jk, I looked it up and I see that DuckDB uses 1-indexing for lists)

packages/core/entity/SQLBuilder/index.ts
```diff
@@ -21,12 +23,248 @@ export default class SQLBuilder {
...
+        pathIsArray: boolean[],
+        leafTransform = (leafExpr: string) => leafExpr
+    ): string {
+        // Single-segment paths (no nesting) are just a column reference, no transforms needed.
+        if (path.length === 1) {
+            return leafTransform(`"${path[0]}"`);
+        }
+
+        // Keep lambda variable names deterministic so generated SQL strings remain stable in tests.
+        const lambdaVars = ["x", "y", "z", "w", "v", "u", "t", "s"];
```
aswallace, line 179: is this an arbitrarily chosen number of lambda var names? It seems reasonable to assume there won't be more nesting than this, just making sure I'm following

SeanDuHare: Yep!

packages/core/entity/SQLBuilder/index.ts
SeanDuHare: I'll add a comment for this! Will probably come up in the future

packages/core/entity/FileFilter/index.ts
SeanDuHare: True! Fixed in 59517498

packages/core/entity/FileFilter/index.ts
SeanDuHare: Yep, let me know if this didn't help ([d92633dc](https://github.com/AllenInstitute/biofile-finder/pull/851/commits/d92633dccbeec71bf8b67342aaa5fa96be03e5bf))

packages/core/components/AnnotationPicker/index.tsx
SeanDuHare: Oh I also totally misread your question! I was thinking about the _property_ `unit` rather than the example literally just right there "Well.Dose.Unit". Yes! You're absolutely right that "Unit" would be what is displayed and also that yes it might be weird! It is just an example though and I don't think this changeset should include anything to add more complex support for adding units _however_ we should definitely look to get that on the roadmap.

If useful I could change the example?

packages/core/components/QueryPart/useSizeSpecificPrefix.ts
SeanDuHare: Oh this is intentional & is based on Thao's design. I do see how it is very unintuitive though based on how the props are handed off in QueryFilter. I tried to make this feel more explicit in 93252b05 but lmk your thoughts

aswallace: Ah, I see!

packages/core/entity/FileFilter/index.ts
```diff
@@ -131,13 +131,13 @@ export default class FileFilter {
      */
     public toSimpleWhereClause(pathIsArray: boolean[]): string {
         const isNested = this.path.length > 1;
+        const columnExpr = SQLBuilder.buildNestedAccessExpression(this.path, pathIsArray);
```
aswallace, line 137: I think [5951749](https://github.com/AllenInstitute/biofile-finder/pull/851/commits/595174984aa7dfb1e953aad726eebfba28999ac5) accidentally got rid of the non `isNested` ternary case that uses `"${this.path[0]}"` instead of `buildNestedAccessExpression` for the simple cases

SeanDuHare: Yep! In a different recent commit the static build method got the ability to handle single element paths in the same way the ternary did so I removed it

aswallace: Ahhh never mind then!

packages/core/components/FileDetails/MetadataList/Value.module.css
```diff
@@ -0,0 +1,20 @@
...
+    min-height: 30px;
+    max-height: 90px;
+    white-space: normal;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    display: -webkit-box !important;
+    -webkit-line-clamp: 4 !important; /* number of lines to show */
+            line-clamp: 4; 
+    -webkit-box-orient: vertical !important;
+    word-wrap: break-word !important;
```
aswallace, line 37: Appears not to be breaking correctly for truncated values 

Image

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -1,23 +1,134 @@
...
+ *     "Solution": [
+ *       {
+ *         "Dose": "97",
+ *         "Unit": "mL"
+ *       }
+ *     ]
+ *   }
+ * ]
+ */
+function unwrapNestedMetadata(values: UnwrappedMetadataValue | undefined | null): MetadataValue {
```
aswallace: nit: type `UnwrappedMetadataValue` already includes `undefined | null`

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -33,34 +144,144 @@ export default class DatabaseFileService implements FileService {
...
+
+    /**
+     * Build the SELECT clause columns for a manifest query.
+     *
+     * Selecting a nested sub-field column (dotted name like "Well.Dose.Unit") directly would
+     * cause a DuckDB Binder Error since no such column exists at the top level.
+     *
+     * For CSV, nested sub-fields are extracted and joined into a readable comma-separated
+     * VARCHAR cell per row (e.g. "uM, mM"), which is the most human-readable form for a
+     * flat text format.
```
aswallace: Is this a to do? @ subfields extracted/joined for csv

packages/core/services/FileService/DatabaseFileService/test/DatabaseFileService.test.ts
```diff
@@ -390,5 +292,252 @@ describe("DatabaseFileService", () => {
...
+            const databaseFileService = new DatabaseFileService({
+                dataSourceNames: ["parquet_source"],
+                databaseService: mockDbService,
+                downloadService: new FileDownloadServiceNoop(),
+            });
+            const selections = [
+                { indexRanges: [{ start: 0, end: 1 }], filters: [], sort: undefined },
+            ];
+
+            // Only "Well.Color" is selected — sibling "Well.Name" must NOT appear.
```
aswallace: I think I'm slightly confused with this test, is `Well.Name` defined somewhere?

aswallace (approved):
Have now reviewed all files! Only other comment is the manifest saving issue I messaged about separately

packages/core/components/FileDetails/MetadataList/Value.module.css
SeanDuHare: Fixed in 874204de

packages/core/services/FileService/DatabaseFileService/index.ts
SeanDuHare: Fixed in 89bf876a8a78d008e7beb2694cc515ff20fa389d

BrianWhitneyAI (approved):
Nice work! thanks for breaking down the changes

packages/core/services/FileService/DatabaseFileService/test/DatabaseFileService.test.ts
SeanDuHare: This was a wonky way of trying to make sure another column wasn't selected - but since I moved the selection sql to its own function it became kind of useless. I removed this assertion and also refactored out the SQL tests to be even more separate from the download() function tests

packages/core/services/FileService/DatabaseFileService/index.ts
SeanDuHare: This was a TODO! It was also very much broken anyway! This function had to be majorly refactored to better handle intermediate and leaf nodes of nested data to be arrays (or single elements) - previously it was assuming one or the other

PR #858, opened 2026-06-30
Feature: Zarr Conversion

BrianWhitneyAI:
## Summary
Adds in `zarr-conversion` to the pipelines option. In order to accomodate some of the new params, we have to update the pipeline-param display and typing logic.

Try it out on https://staging.bff.allencell.org/  select a CZI or ND2, -> process files -> conversion ->submit


packages/core/services/PipelineService/mockPipelineData.ts
```diff
@@ -38,4 +47,35 @@ export const MOCK_PARAMETERS: Record<string, PipelineParameter[]> = {
             validation: { min: 0 },
         },
     ],
+    "zarr-conversion": [
```
BrianWhitneyAI, line 50: mock data to be dropped when discovery is live

packages/core/components/Modal/ComputePipelineModal/index.tsx
```diff
@@ -474,9 +483,10 @@ export default function ComputePipelineModal({ onDismiss }: ModalProps) {
                                                                 text: p.label,
                                                             })
                                                         )}
-                                                        selectedKey={
-                                                            optionalSelectorKey ?? undefined
-                                                        }
+                                                        // null (not undefined) so the dropdown
```
SeanDuHare: A bit verbose of a comment IMO - the type is defined elsewhere for this variable so perhaps it would belong there

packages/core/components/Modal/ComputePipelineModal/PipelineParameterInput.tsx
```diff
@@ -20,22 +21,62 @@ export default function PipelineParameterInput({
...
     onBlur,
 }: PipelineParameterInputProps) {
+    const renderControl = () => {
+        switch (param.type) {
+            case "select":
+                return (
+                    <BaseComboBox
+                        label=""
+                        selectedKey={
+                            value !== null && value !== undefined && value !== ""
```
SeanDuHare: FWIW I feel like using `isNil` and `isEmpty` from lodash are handy for making things like this more readable - just an opinion though
```suggestion
                            (!isNil(value) && !isEmpty(value)) ? String(value) : null
```

packages/core/components/Modal/ComputePipelineModal/PipelineParameterInput.tsx
```diff
@@ -20,22 +21,62 @@ export default function PipelineParameterInput({
...
+                                : "Select an option"
+                        }
+                        onChange={(option) => onChange(option ? option.key : "")}
+                    />
+                );
+            case "string":
+                return (
+                    <TextField
+                        type="text"
+                        value={value !== null && value !== undefined ? String(value) : ""}
```
SeanDuHare: Same comment here for `isNil` usage but (again) just an opinion

packages/core/components/Modal/ComputePipelineModal/PipelineParameterInput.tsx
```diff
@@ -20,22 +21,62 @@ export default function PipelineParameterInput({
...
+                        onChange={(_, v) => onChange(v ?? "")}
+                        onBlur={onBlur}
+                        placeholder={param.default !== null ? String(param.default) : ""}
+                        borderless
+                        className={styles.textField}
+                    />
+                );
+            case "number":
```
SeanDuHare, line 54: Does this mean the `default` would also catch the `file_paths` type?

packages/core/services/PipelineService/index.ts
```diff
@@ -35,11 +35,22 @@ export default class PipelineService extends HttpServiceBase {
...
+    }
+
     async submitComputeTask(request: ComputeTaskRequest): Promise<ComputeTaskResponse> {
         const url = `${this.loadBalancerBaseUrl}/fss2/v4.0/compute/${request.pipeline}`;
 
         const { file_paths, ...rest } = request.parameters;
-        const body: Record<string, unknown> = { files: file_paths };
+        const files = Array.isArray(file_paths)
+            ? file_paths.map((p) => PipelineService.encodeFilePath(p))
+            : file_paths;
```
SeanDuHare: Why do they not need to be encoded when not in an array?

packages/core/services/PipelineService/index.ts
```diff
@@ -35,11 +35,22 @@ export default class PipelineService extends HttpServiceBase {
...
+        const body: Record<string, unknown> = { files };
```
SeanDuHare: Can the type narrow down from `unknown`? I assume the values that get added below must be of some constrained typing like `string | number | boolean` etc

packages/core/services/PipelineService/mockPipelineData.ts
SeanDuHare: Is there an anticipated date for that?

packages/core/components/ComboBox/index.tsx
```diff
@@ -6,7 +6,7 @@ import styles from "./ComboBox.module.css";
 
 interface Props {
     className?: string;
-    selectedKey?: string;
+    selectedKey?: string | null;
```
SeanDuHare, line 9: Should `undefined` be removed in favor of `null`?

BrianWhitneyAI:
>> Is there an anticipated date for that?

3 of the 4 parts of it are out for review now

packages/core/components/ComboBox/index.tsx
BrianWhitneyAI: null is the clear value so I think having both undefined and null here is ok/

aswallace: we also still need `undefined` because it's currently an optional param, so if it were removed `ExistingAnnotationPathway` and `MetadataDetails` would need to be refactored slightly

aswallace (approved):
+1 on Sean's comments, otherwise lgtm

packages/core/components/Modal/ComputePipelineModal/PipelineParameterInput.tsx
BrianWhitneyAI: No file paths are already filtered out

packages/core/services/PipelineService/index.ts
BrianWhitneyAI: moved away from the unknown typing in [0d86855](https://github.com/AllenInstitute/biofile-finder/pull/858/commits/0d868556c1dd99ede3fae3fa24bfcb69eea6db50)

packages/core/components/Modal/ComputePipelineModal/PipelineParameterInput.tsx
BrianWhitneyAI: [0b0811e](https://github.com/AllenInstitute/biofile-finder/pull/858/commits/0b0811ee66911ee87e3ddcb9596276c513b8a2fe)

packages/core/components/Modal/ComputePipelineModal/PipelineParameterInput.tsx
BrianWhitneyAI: [0b0811e](https://github.com/AllenInstitute/biofile-finder/pull/858/commits/0b0811ee66911ee87e3ddcb9596276c513b8a2fe)

packages/core/components/Modal/ComputePipelineModal/index.tsx
BrianWhitneyAI: [0b0811e](https://github.com/AllenInstitute/biofile-finder/pull/858/commits/0b0811ee66911ee87e3ddcb9596276c513b8a2fe)

packages/core/components/Modal/ComputePipelineModal/PipelineParameterInput.tsx
SeanDuHare: Yes, I think an explicit catch of file_paths here would be good to avoid relying on an upstream type guard

SeanDuHare (approved):

packages/core/services/PipelineService/index.ts
SeanDuHare: @BrianWhitneyAI ?

PR #859, opened 2026-06-30
Auto detect webpage-like files as needing browser viewer

SeanDuHare:
## Context
Some users want to have their rows represent webpages or like for [IDR](https://bff.allencell.org/app?c=File+Name%3A9%2CAge%3A15%2CChannels%3A45%2CClinical+Information%3A19%2Cdata_source%3A11%2Cdescription%3A11&source=%7B%22uri%22%3A%22https%3A%2F%2Fidr.openmicroscopy.org%2Fsearchengine%2Fapi%2Fv1%2Fresources%2Fcontainer_data%2F%3Fcontainer_name%3Didr0026-weigelin-immunotherapy%2FexperimentA%26container_type%3Dproject%26file_type%3Dparquet%22%2C%22type%22%3A%22parquet%22%2C%22name%22%3A%22idr0026-weigelin-immunotherapy%2FexperimentA.parquet%22%7D) want to supply the file in the context of their own viewer. 

## Changes
The Column Descriptions feature (where you tell BFF which columns represent links) handles this more explicitly. However, it seems like a nice to have to have BFF auto-detect when a linked "File Path" is a webpage - when it is able to do so it will supply "Browser (<hostname>)" as the "Open with" option. For example, if we remove the IDR specialized handling it would render "Browser (idr.openmicroscopy.org)" as the option.

## Testing
This feature only works if BFF is able to reach out to the web page via a HEAD (ideal) or GET request which requires CORS to be enabled for BFF on the webpage.

### DIFF
--- a/packages/core/hooks/useOpenWithMenuItems/index.tsx
+++ b/packages/core/hooks/useOpenWithMenuItems/index.tsx
@@ -265,12 +265,19 @@ function getSupportedApps(
             // If the content of the file linked appears to be a webpage
             // then offer Browser as the only option
             if (fileContentType === "webpage") {
-                const hostname = new URL(fileDetails.path).hostname;
+                let hostname: string | undefined;
+                try {
+                    hostname = new URL(fileDetails.path).hostname;
+                } catch (_e) {
+                    // Somehow, not a valid URL; skip formatting hostname
+                }
                 return [
                     {
                         ...apps.browser,
-                        text: `Browser (${hostname})`,
-                        title: `Open ${hostname} in the current browser in a new tab`,
+                        text: hostname ? `Browser (${hostname})` : "Browser",
+                        title: hostname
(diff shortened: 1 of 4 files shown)


packages/core/hooks/useOpenWithMenuItems/index.tsx
```diff
@@ -261,6 +262,19 @@ function getSupportedApps(
...
             }
 
+            // If the content of the file linked appears to be a webpage
+            // then offer Browser as the only option
+            if (fileContentType === "webpage") {
+                const hostname = new URL(fileDetails.path).hostname;
+                return [
+                    {
+                        ...apps.browser,
+                        text: `Browser (${hostname})`,
```
SeanDuHare: I wonder if just `text: hostname` would be better UX?

PR #862, opened 2026-07-02
Omit unexpected keys in FmsFile from detail

### LINKED ISSUE #861: Bug: Issue with selecting any file opens error
## Description
When selecting a file in BFF on fms data the following is shown 'Unexpected internal error. Unable to find column metadata for field "uploadedBy". Omitting this field from display.' even if the file in question has a 'uploadedBy' annotation.

## Reproduction
Open BFF and select a file

SeanDuHare:
## Context
The below error appears when loading the FMS data source. This is due to a duplicate, unnecessary key "uploadedBy" at the top level of the file document in the FMS MongoDB.
Screenshot 2026-07-02 at 3 20 10 PM

## Changes
This adds a check to have any unexpected keys found on the object given to the `FileDetail` to get omitted.

### DIFF
--- a/packages/core/entity/FileDetail/index.ts
+++ b/packages/core/entity/FileDetail/index.ts
@@ -90,6 +90,16 @@ export interface FmsFile {
     uploaded?: string;
     thumbnail?: string;
 }
+// Keys of fields that are expected to be present in the FMS file detail response. Any other keys will be ignored.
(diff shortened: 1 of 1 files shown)


BrianWhitneyAI (approved):
Should this just be dropping uploadedBy from the fmsfile since we already have the annotation?

aswallace (approved):

PR #867, opened 2026-07-07
Feature/allow adding optional files alone

### LINKED ISSUE #742: Allow adding metadata/provenance files alone
## Use Case

The data source prompt currently only allows users to upload metadata/provenance descriptor files when they're also uploading a regular data source at the same time. Users should be able to add them singly to existing queries (e.g., that already have a data source)

## Suggest a Solution

Enable the `LOAD` button on the data source prompt when users are adding descriptor files to an existing query, as long as that query already has data sources we still have access to/don't require reloading.

## Additional Details

Relies on #732

aswallace:
## Context
Addresses two issues 
- #742: Users could only add the optional descriptor files (provenance info, metadata descriptors) at the same time they initially add their data source, which meant there was no way to add it later on 
- Provenance wasn't loading properly when coming from a local file

## Changes
- Adds an option in the query's data source menu that allows users to add just a provenance or metadata file
- Moves the main body of the data source prompt into a template/helper to avoid duplicating code & styling

#### For provenance: 
- Switches to using the file name as the table name instead of `"source_provenance"`, which is in line with how we already handle metadata descriptor files
- Includes processProvenance as part of the logic for loading a new query 

## Testing
- Tested adding provenance & metadata files after the fact
- Made sure both local and web sources work for provenance & metadata

## Screenshots
New menu options in the data source section for queries
image

Modal for loading just a metadata descriptor source using the simple version of the data source modal 
image


packages/core/components/DataSourcePrompt/DataSourcePromptTemplate.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+import styles from "./DataSourcePrompt.module.css";
+import { DefaultButton, Icon } from "@fluentui/react";
+
+const ADDITIONAL_COLUMN_DETAILS = [
+    'If a "Thumbnail" column is present, it should contain a web URL to a thumbnail image for the file. ',
+    'If a "File Name" column is present, it should be the file\'s name, which will replace the "File Name" created by default from the path. ',
+    'If a "File Size" column is present, it should contain the size of the file in bytes, providing feedback to the user during downloads. ',
+    'If an "Uploaded" column is present, it should contain the date the file was uploaded to the storage and be formatted as YYYY-MM-DD HH:MM:SS.Z where Z is a timezone offset. ',
+];
```
aswallace, line 16: This is copied from the original DataSourcePrompt component without changes

packages/core/components/DataSourcePrompt/DataSourcePromptTemplate.tsx
```diff
@@ -0,0 +1,173 @@
...
+                    <PrimaryButton
+                        className={classNames(styles.loadButton)}
+                        disabled={props.submitDisabled}
+                        text="LOAD"
+                        onClick={props.onSubmit}
+                    />
+                </div>
+            </div>
+
+            {!!props?.showAdvancedOptions && (
```
aswallace, line 63: Everything from here through to the end of this file is copied from the original DataSourcePrompt component without changes

packages/core/services/DatabaseService/index.ts
```diff
@@ -575,23 +574,30 @@ export default abstract class DatabaseService {
...
+                await this.prepareDataSourceWrapper(
+                    {
+                        ...sourceProvenance,
+                        name: sourceProvenance.name,
+                    },
+                    true
+                );
+            }
+        }
+        // If the source doesn't have a uri, we should instead try to use the cached table
```
aswallace, line 708: Changed to be consistent with how we handle deleting/replacing the metadata descriptor file

packages/core/state/selection/selectors.ts
```diff
@@ -39,9 +43,11 @@ const getPlatformDependentServices = (state: State) => state.interaction.platfor
 // COMPOSED SELECTORS
 export const hasQuerySelected = createSelector([getSelectedQuery], (query): boolean => !!query);
 
+// If there are no edge definitions, i.e., it's an invalid provenance file,
+// the UI should treat it as if there is no provenance file
```
aswallace, line 47: Was running into an issue where if the provenance file didn't load correctly (e.g., we fail to define edges/nodes based on the rules in the file, or the file is not accessible), the "View relationship diagram" would still show up in the FileDetailsPanel but cause weird UI bugs when clicked since no provenance graph could be created

packages/core/components/DataSourcePrompt/DataSourcePromptTemplate.tsx
```diff
@@ -0,0 +1,173 @@
+import classNames from "classnames";
+import * as React from "react";
+
+import FilePrompt from "./FilePrompt";
+import { PrimaryButton } from "../Buttons";
+import { Source } from "../../entity/SearchParams";
+
+import styles from "./DataSourcePrompt.module.css";
+import { DefaultButton, Icon } from "@fluentui/react";
```
SeanDuHare: (nit) import order

packages/core/state/selection/selectors.ts
SeanDuHare: Feels like the database handler for ingesting provenance files should toss up an error if there aren't any edges defined rather than silently fail. 
> file is not accessible

When does this happen? On reload?

packages/core/components/DataSourcePrompt/index.tsx
```diff
@@ -16,45 +16,119 @@ interface Props {
...
+    onDismiss: () => void;
+}
 
 export enum DataSourceType {
     default = 0,
     metadata = 1,
     provenance = 2,
 }
 
+export const DataSourceTypeToString = (sourceType: DataSourceType) => {
```
SeanDuHare: cased like an enum or type rather than a function

SeanDuHare: should this just move into DataSource/index.tsx? It looks like it is only used there

packages/core/services/DatabaseService/index.ts
SeanDuHare: Is this a possible case for provenance or descriptor files? Only the currently selected provenance/descriptor file is cached I thought - is it possible for them to re-select the same file that is already selected?

packages/core/state/selection/selectors.ts
aswallace: > When does this happen? On reload?

Yeah, so far I've just had it happen on reload

You can see the issue in action [here](https://bff.allencell.org/app?c=File+Name%3A10%2CColor%3A7%2CFile+ID%3A7%2CFile+Path%3A10%2CInput+File%3A10%2CInput+File+Model%3A37&source=%7B%22name%22%3A%22pub+%287%2F7%2F2026+11%3A47%3A58+AM%29%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22https%3A%2F%2Fdocs.google.com%2Fspreadsheets%2Fd%2Fe%2F2PACX-1vTesK5EHZZAsqG6YvAjvc6pCiRFgM1gQ-1hQFwkN15sF2g4q3ooXTwR0dzUY3KriCy0LEMr904Ik7uq%2Fpub%3Fgid%3D978740010%26single%3Dtrue%26output%3Dcsv%22%7D&prov=%7B%22name%22%3A%22Demo+Provenance+-+prov+%282%29%22%2C%22type%22%3A%22csv%22%7D) since I used a url for the data source and a local file for provenance. If you click on any file and then click "View relationship diagram" in the file details panel, it causes weird UI issues

packages/core/components/DataSourcePrompt/index.tsx
aswallace: fixed both in [0ba78f5](https://github.com/AllenInstitute/biofile-finder/pull/867/commits/0ba78f52d07be1181de4d269899f57a6cf1549ed)

BrianWhitneyAI (approved):

packages/core/state/selection/selectors.ts
SeanDuHare: I think this would be better if it was more visible. Is this something we could catch while processing provenance and appears to the user as a warning level status message (unsure if error is right but maybe it is?)

aswallace: In the case I linked there it technically does, we display an error letting the user know that the provenance file needs to be reloaded, and this selector here just prevents the "View provenance diagram" button from rendering in the file details panel. I agree though that in general we should also warn the user if we _were_ able to load the file and there are no edge definitions (and also open to alternatives re whether we should show the "View provenance diagram" button when there have been errors, maybe it should be present but disabled?)

SeanDuHare (approved):

packages/core/services/DatabaseService/index.ts
aswallace: The problem is when switching between queries that use local files. Since we delete/lose reference to the files & definitions in state when we switch, so we need to use the reference to the cached table instead.
We may need a separate ticket to clean this up, I've run into it a few times now. I think the readme backend PR will help since part of the issue is the search params encode/decode step, but we may also want to revisit the fact that we're deleting sources when we change queries

packages/core/state/selection/selectors.ts
aswallace: I'm resolving this comment for now since I think it's been resolved by #872

PR #869, opened 2026-07-08
Feature/add prov to aics fms

SeanDuHare:
## Context
Provenance is available for use in the application, however for internal users their query defaults to AICS FMS which won't include provenance at the moment.

## Changes
This adds a provenance source to the default AICS FMS query. While doing so I encountered 3 other things:
1) The origin would be out of sight after the graph loads for large graphs, so I added a zoom to `NetworkGraph/index.tsx`
2) "File Path" is not a good column to query AICS FMS by, so this changes the column to search to "File ID" when querying AICS FMS
3) Fixed a bug that happens on large graphs when the node searching ends early. If an edge exists without both nodes existing already the graph would generate a placeholder node that would cause the `graph.node(<node id>)` to bubble up an error (`Graph/index.ts`)

## Testing
[Tested using this provenance source file](https://staging-biofile-finder-datasets.s3.us-west-2.amazonaws.com/FMS.provenance.csv)

## Example
Screenshot 2026-07-08 at 10 21 43 AM

### DIFF
--- a/packages/core/components/NetworkGraph/index.tsx
+++ b/packages/core/components/NetworkGraph/index.tsx
@@ -7,8 +7,8 @@ import {
     useEdgesState,
     Controls,
     useReactFlow,
-    useNodesInitialized,
     ReactFlowProvider,
+    FitViewOptions,
 } from "@xyflow/react";
 import "@xyflow/react/dist/style.css";
 import classNames from "classnames";
@@ -55,8 +55,7 @@ function NetworkGraph(props: NetworkGraphProps) {
     const provenanceSource = useSelector(selection.selectors.getSelectedSourceProvenance);
     const [edges, setEdges, onEdgesChange] = useEdgesState<Edge<AnnotationEdge>>([]);
     const [nodes, setNodes, onNodesChange] = useNodesState<FileNodeType | MetadataNodeType>([]);
-    const { fitView, setCenter, getZoom } = useReactFlow();
-    const nodesInitialized = useNodesInitialized();
+    const { fitView, getZoom } = useReactFlow();
 
     // Unfortunately we have to have some notion of state at a high level for control from the components
     // and at the dagre level for when the user does a drag action causing this duplication of efforts
@@ -65,22 +64,14 @@ function NetworkGraph(props: NetworkGraphProps) {
         setNodes(graph.nodes);
     }, [graph, setEdges, setNodes, refreshKey]);
 
-    // Once the graph has finished building, pan so the origin (selected) node
-    // is centered in the viewport. The origin is the sole node flagged as
-    // selected (see Graph.originate -> createFileNode(origin, true)).
-    const lastCenteredRefreshKeyRef = React.useRef<string | undefined>(undefined);
-    React.useEffect(() => {
-        if (isLoading || !nodesInitialized) return;
-        if (lastCenteredRefreshKeyRef.current === refreshKey) return;
-
-        const originNode = nodes.find((node) => node.data.isSelected);
-        if (!originNode) return;
-
-        lastCenteredRefreshKeyRef.current = refreshKey;
-        const centerX = originNode.position.x + (originNode.width ?? 0) / 2;
-        const centerY = originNode.position.y + (originNode.height ?? 0) / 2;
-        setCenter(centerX, centerY, { zoom: getZoom(), duration: 800 });
-    }, [isLoading, nodesInitialized, nodes, refreshKey, setCenter, getZoom]);
+    const originNodeId = React.useMemo(() => nodes.find((node) => node.data.isSelected)?.id, [
+        nodes,
+    ]);
+    const fitViewOptions: FitViewOptions<FileNodeType | MetadataNodeType> = React.useMemo(() => {
+        if (!originNodeId) return {};
+        // Center the origin node in the viewport and set the zoom level to fit the graph
+        return { nodes: [{ id: originNodeId }], duration: 800, maxZoom: getZoom() };
+    }, [originNodeId, getZoom]);
 
     // The option to open this graph shouldn't even appear when a
     // source isn't available so this shouldn't ever happen
@@ -98,7 +89,7 @@ function NetworkGraph(props: NetworkGraphProps) {
 
     const onClickReset = () => {
         graph.resetLayout(); // return to default layout if any
-        fitView(); // reset zoom
+        fitView(fitViewOptions); // reset zoom and center on origin
         dispatch(interaction.actions.refreshGraph());
     };
 
@@ -112,6 +103,8 @@ function NetworkGraph(props: NetworkGraphProps) {
                 onClick={onClickReset}
             />
             <ReactFlow
+                fitView
+                fitViewOptions={fitViewOptions}
                 onlyRenderVisibleElements
                 className={styles.graph}
                 edgesFocusable={false}
--- a/packages/core/components/NetworkGraph/index.tsx
+++ b/packages/core/components/NetworkGraph/index.tsx
@@ -66,16 +66,19 @@ function NetworkGraph(props: NetworkGraphProps) {
     // Once the graph has finished building, pan so the origin (selected) node
     // is centered in the viewport. The origin is the sole node flagged as
     // selected (see Graph.originate -> createFileNode(origin, true)).
+    const lastCenteredRefreshKeyRef = React.useRef<number | undefined>(undefined);
     React.useEffect(() => {
         if (isLoading) return;
+        if (lastCenteredRefreshKeyRef.current === refreshKey) return;
 
         const originNode = nodes.find((node) => node.data.isSelected);
         if (!originNode) return;
 
+        lastCenteredRefreshKeyRef.current = refreshKey;
         const centerX = originNode.position.x + (originNode.width ?? 0) / 2;
         const centerY = originNode.position.y + (originNode.height ?? 0) / 2;
         setCenter(centerX, centerY, { zoom: getZoom(), duration: 800 });
-    }, [isLoading, nodes, setCenter, getZoom]);
+    }, [isLoading, nodes, refreshKey, setCenter, getZoom]);
 
     // The option to open this graph shouldn't even appear when a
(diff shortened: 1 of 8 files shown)


packages/core/entity/Graph/index.ts
```diff
@@ -510,16 +526,8 @@ export default class Graph {
                 // Avoid re-requesting the file when possible
                 const node = this.graph.node(value); // the value should be a file path
                 if (node) return node;
-                try {
-                    const file = await this.getFileBy("File Path", [value]);
-                    if (file) {
-                        return createFileNode(file);
-                    }
-                } catch {
-                    // Backup while moving between provenance versions: Try looking by "File ID", this should be removed in the future
```
SeanDuHare, line 519: @aswallace lmk if you think we should still have a fallback

packages/core/components/NetworkGraph/index.tsx
```diff
@@ -63,6 +63,23 @@ function NetworkGraph(props: NetworkGraphProps) {
...
+        if (isLoading) return;
+        if (lastCenteredRefreshKeyRef.current === refreshKey) return;
+
+        const originNode = nodes.find((node) => node.data.isSelected);
+        if (!originNode) return;
+
+        lastCenteredRefreshKeyRef.current = refreshKey;
+        const centerX = originNode.position.x + (originNode.width ?? 0) / 2;
+        const centerY = originNode.position.y + (originNode.height ?? 0) / 2;
+        setCenter(centerX, centerY, { zoom: getZoom(), duration: 800 });
```
BrianWhitneyAI: Whats duration here? 800?

BrianWhitneyAI (approved):

packages/core/components/NetworkGraph/index.tsx
SeanDuHare: Time to pan, its in milliseconds. Just makes it animate a little rather than jump suddenly.

packages/core/components/NetworkGraph/index.tsx
```diff
@@ -63,6 +65,23 @@ function NetworkGraph(props: NetworkGraphProps) {
...
+        if (isLoading || !nodesInitialized) return;
+        if (lastCenteredRefreshKeyRef.current === refreshKey) return;
+
+        const originNode = nodes.find((node) => node.data.isSelected);
+        if (!originNode) return;
+
+        lastCenteredRefreshKeyRef.current = refreshKey;
+        const centerX = originNode.position.x + (originNode.width ?? 0) / 2;
+        const centerY = originNode.position.y + (originNode.height ?? 0) / 2;
+        setCenter(centerX, centerY, { zoom: getZoom(), duration: 800 });
```
aswallace: This is really nice on initial node and when resetting the view! but it's slightly inconvenient when working elsewhere in the graph, e.g., if changing the organization structure

https://github.com/user-attachments/assets/18429c1c-9009-4bf7-aca0-ec1a40cf2367

packages/core/entity/Graph/index.ts
aswallace: I think this is fine with me! I don't think we have enough people using this right now for it to be an issue. I only sent the old template to Chantelle, so we can check in with her directly

packages/core/services/FileService/DatabaseFileService/index.ts
```diff
@@ -143,6 +143,7 @@ export default class DatabaseFileService implements FileService {
     private readonly databaseService: DatabaseService;
     private readonly downloadService: FileDownloadService;
     private readonly dataSourceNames: string[];
+    public readonly provenanceIdColumn = "File Path";
```
aswallace: Do we have the "File Path" string defined as a constant anywhere? just wondering since it's our one required column name

packages/core/services/FileService/HttpFileService/index.ts
```diff
@@ -90,6 +90,7 @@ export default class HttpFileService extends HttpServiceBase implements FileServ
     private static readonly CSV_ENDPOINT_VERSION = "2.0";
     public static readonly BASE_CSV_DOWNLOAD_URL = `file-explorer-service/${HttpFileService.CSV_ENDPOINT_VERSION}/files/selection/manifest`;
     private readonly downloadService: FileDownloadService;
+    public readonly provenanceIdColumn = "file_id";
```
aswallace: Making sure I'm on the same page-- For "Input File Path", we just render an entity instead of looking up the file path?

aswallace (commented):
Overall lgtm! A couple bugs(?) that I wasn't sure where to comment about: 

I'm still seeing some floating files, e.g., if you scroll to the right for [this query](http://localhost:9001/app?c=file_name%3A30%2CKind%3A15%2CType%3A20%2Cfile_size%3A10%2C520+or+570%3F%3A15%2CAbsence+Of+Migrating+Cells+Coming+From+Colony+Out+Of+FOV+At+Time+Of+Migration%3A15&filter=%7B%22name%22%3A%22uploaded%22%2C%22value%22%3A%22RANGE%282025-07-09T07%3A00%3A00.000Z%2C2026-07-10T06%3A59%3A59.693Z%29%22%2C%22type%22%3A%22default%22%7D&source=%7B%22name%22%3A%22AICS+FMS%22%7D&prov=%7B%22name%22%3A%22AICS%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22https%3A%2F%2Fbiofile-finder-datasets.s3.us-west-2.amazonaws.com%2FFMS.provenance.csv%22%7D&p=21201a9ccb774a5ab030bea6d37a9b37&sort=%7B%22annotationName%22%3A%22uploaded%22%2C%22order%22%3A%22DESC%22%7D)

Image

I'm also running into an `annotationName.split is not a function` error for opening queries related to certain edge connection types, I think because it's expecting "is derived from" to have a "pointer" type

https://github.com/user-attachments/assets/22e4cb43-ec3d-478f-800a-9144c714f6e8

packages/core/components/NetworkGraph/index.tsx
SeanDuHare: Oh good point! I'll make sure it just functions on the initial load (maybe the reset too?).

packages/core/services/FileService/DatabaseFileService/index.ts
SeanDuHare: Not really - we do have an enum of PreDefinedColumns but the usage of that is a bit different. Once we can stop using the FMS names for things (file_id for example) it'll be a good opportunity to clean a lot of the usage around names up

We did have a constant for "file_id" that I swapped in

packages/core/services/FileService/HttpFileService/index.ts
SeanDuHare: Yes, we're talking in Slack about this causing issues for Brian's work potentially & it does seem like we could add a way for each edge to define how to look up their file's ID column

SeanDuHare: Rolling with this for now (https://github.com/AllenInstitute/biofile-finder/pull/869/commits/7c5251060950cea400652fc6ef2874ce90ffea32) but it does feel increasingly like the edges should define their lookup column - lmk if anyone has thoughts or ideas

packages/core/components/NetworkGraph/index.tsx
SeanDuHare: Addressed in this https://github.com/AllenInstitute/biofile-finder/pull/869/commits/31b78bfcf7debc649580fc457da41dff4b905ee6 - I looked harder into the fitView options and found that I had missed that you can have it focus the view on specific nodes by using the `nodes` property in the `FitViewOptions` - now it just tries to find the origin and passes that to the fitView which in my testing makes it so:
* Inits cause a centering
* Re-renders like via the re-org of nodes does NOT cause a centering
* Resetting the nodes causes a centering

packages/core/services/FileService/HttpFileService/index.ts
aswallace: I agree, I think we can still default to something (file id or path) but it does feel like it's starting to be more complicated to have to work around it then to just allow that

SeanDuHare:
> Overall lgtm! A couple bugs(?) that I wasn't sure where to comment about:
> 
> I'm still seeing some floating files, e.g., if you scroll to the right for [this query](http://localhost:9001/app?c=file_name%3A30%2CKind%3A15%2CType%3A20%2Cfile_size%3A10%2C520+or+570%3F%3A15%2CAbsence+Of+Migrating+Cells+Coming+From+Colony+Out+Of+FOV+At+Time+Of+Migration%3A15&filter=%7B%22name%22%3A%22uploaded%22%2C%22value%22%3A%22RANGE%282025-07-09T07%3A00%3A00.000Z%2C2026-07-10T06%3A59%3A59.693Z%29%22%2C%22type%22%3A%22default%22%7D&source=%7B%22name%22%3A%22AICS+FMS%22%7D&prov=%7B%22name%22%3A%22AICS%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22https%3A%2F%2Fbiofile-finder-datasets.s3.us-west-2.amazonaws.com%2FFMS.provenance.csv%22%7D&p=21201a9ccb774a5ab030bea6d37a9b37&sort=%7B%22annotationName%22%3A%22uploaded%22%2C%22order%22%3A%22DESC%22%7D)
> 
> Image
> I'm also running into an `annotationName.split is not a function` error for opening queries related to certain edge connection types, I think because it's expecting "is derived from" to have a "pointer" type
> 
>  Screen.Recording.2026-07-09.at.9.43.35.AM.mov

For the floating files - I think those are actually connected its just a jumbled mess :/ when I inspect them they appear to have children - if you point me to a specific file to use as the origin i can triple check this

For the `annotationName.split is not a function` error, this has to do with parent and child being nil or an empty string - when that happens it kind of messes up the query that this creates. It doesn't seem like my PR caused this so much as it exposes it since my provenance file has this use case. I have a fix, but its a bit large so I'm splitting it out into its own PR that I'll target this branch with

aswallace (approved):

PR #872, opened 2026-07-10
Warn user of bad provenance input

SeanDuHare:
## Context
Anya noticed a bug [here](https://github.com/AllenInstitute/biofile-finder/pull/869#pullrequestreview-4664642723) where the feature to open a query from a edge in the provenance graph would generate an error. This was/is an existing bug in the system where because "self" type edges point are self-referential there is no need for a child/parent for that edge causing the value for the edge to be misrepresented when a query is attempted to be spun off of it.

## Changes
This fixes the aforementioned bug by checking to see if the parent/child are an actual value & adds a validation/warning system for checking the provenance data (it actually found a bug in the provenance file I was planning to use).

## Testing
Tried creating various kinds of provenance issues and loading them in

### Example error message (truncated and full)
Screenshot 2026-07-13 at 10 11 18 AM

Screenshot 2026-07-13 at 10 11 22 AM


packages/core/components/NetworkGraph/Edges/DefaultEdge.tsx
```diff
@@ -61,10 +62,15 @@ const DefaultEdge: FC<EdgeProps<Edge<AnnotationEdge>>> = ({
                         provOriginId: undefined,
                     });
                 } else {
+                    const filters = [data.parent, data.child]
```
SeanDuHare, line 65: This is the main thing that fixes the bug - but these being empty is a result of some of the formatting work that now happens that prevents empty strings / nil values

packages/core/services/DatabaseService/index.ts
```diff
@@ -422,6 +447,51 @@ export default abstract class DatabaseService {
             : str;
     }
 
+    // Checks to see if the row is compatible with ProvenanceRow interface and returns a string describing
+    // the issue if not compatible
+    private static checkProvenanceRowForIssues(row: Record<string, unknown>): string | undefined {
```
SeanDuHare: I feel like all provenance stuff should move into its own class/file and same with the description metadata stuff - feels like bloat here - seems like too much to include in this change

packages/core/state/selection/logics.ts
```diff
@@ -781,18 +781,30 @@ const changeProvenanceSourceLogic = createLogic({
 
         try {
             if (selectedSourceProvenance) {
-                const edgeDefinitions = await databaseService.processProvenance(
+                const { edgeDefinitions, warnings } = await databaseService.processProvenance(
                     selectedSourceProvenance
                 );
-                dispatch(metadata.actions.receiveEdgeDefinitions(edgeDefinitions));
-                // provenance definitions may finish loading after we've already processed url query args.
-                // If we do have a graph origin, this ensures the graph actually starts rendering
```
SeanDuHare, line 789: @aswallace do you remember what this comment is about? I can't tell if it is just a heads up that the graph will start rendering with or without the edge definitions and that is a neutral, positive, or negative effect.

I went with assuming it was negative or neutral and removed it in favor of only dispatching the origin and edge definitions once the provenance has processed _and_ we actually have provenance to show. I think this would happen with the issue you experienced where inaccessible provenance files caused the "View relationship diagram" to still appear.

packages/core/services/DatabaseService/index.ts
```diff
@@ -1187,69 +1257,62 @@ export default abstract class DatabaseService {
...
             // and/or the columns may not exist
             const errMsg = typeof err === "string" ? err : err instanceof Error ? err.message : "";
             if (errMsg.includes("does not exist") || errMsg.includes("not found in FROM clause")) {
-                return [];
+                return { edgeDefinitions: [], warnings: [] };
             }
             throw err;
+        } finally {
+            // The definitions will already be in the state memory, no need to keep this in the database
+            await this.deleteSourceProvenance();
```
SeanDuHare, line 1349: Thought this was kinda nice! I couldn't find anywhere else that we used the provenance file so it felt unnecessary to keep in the DuckDB memory - if I'm missing something though lmk

packages/core/state/selection/logics.ts
```diff
@@ -781,18 +781,28 @@ const changeProvenanceSourceLogic = createLogic({
...
+                if (warnings.length > 0) {
+                    const fullWarning = warnings.join("\n");
+                    dispatch(
+                        interaction.actions.processWarning(
+                            "processProvenanceWarning",
+                            fullWarning.substring(0, 150),
+                            fullWarning
+                        )
+                    );
+                }
```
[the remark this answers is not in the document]
SeanDuHare: I did the <br /> in a commit already

packages/core/state/selection/logics.ts
```diff
@@ -781,18 +781,28 @@ const changeProvenanceSourceLogic = createLogic({
...
+                // Single success case: processed provenance source into edge definitions
+                if (edgeDefinitions.length > 0) {
+                    dispatch(metadata.actions.receiveEdgeDefinitions(edgeDefinitions));
+                    // provenance definitions may finish loading after we've already processed url query args.
+                    // If we do have a graph origin, this ensures the graph actually starts rendering
+                    dispatch(changeProvenanceOriginId(origin) as AnyAction);
+                }
```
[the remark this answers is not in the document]
SeanDuHare: This isn't true, the origin and the edge definitions are reset anytime the action that causes this logic is triggered just within the reducers now

packages/core/state/selection/logics.ts
aswallace: I * think * this was from when we added provenance to the query arguments so that when you're passed a url directly to a provenance graph it will load immediately. At the time, the logic for processing the search params was finishing before the db logic finished processing the provenance schema from the url. It seems like that might no longer be an issue with this change though

packages/core/services/DatabaseService/index.ts
```diff
@@ -15,6 +15,26 @@ export interface CancellablePromise<T> {
...
 }
 
+interface ProvenanceRow {
+    child: string;
+    childtype: EdgeNodeType;
+    parent: string;
+    parenttype: EdgeNodeType;
+    relationship: string;
+    relationshiptype?: RelationshipType;
+}
```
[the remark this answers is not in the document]
SeanDuHare: Untrue

packages/core/services/DatabaseService/index.ts
```diff
@@ -422,6 +447,54 @@ export default abstract class DatabaseService {
...
+        if (
+            !isNil(row.relationshiptype) &&
+            !isEmpty(row.relationshiptype) &&
+            !VALID_RELATIONSHIP_TYPES.includes(row.relationshiptype as RelationshipType)
+        ) {
+            return (
+                `Skipping provenance row with invalid "relationshiptype" value "${row.relationshiptype}". ` +
+                `Expected one of: ${VALID_RELATIONSHIP_TYPES.join(", ")} or none (default).`
+            );
+        }
```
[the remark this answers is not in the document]
SeanDuHare: I'd rather just format out any non-strings, done in most recent commit

packages/core/state/selection/logics.ts
SeanDuHare: I reverted any attempts trying to touch any of the logical ordering of the actions in this PR - seemed unnecessary to try to do any cleanup rn

BrianWhitneyAI (approved):

aswallace (approved):

PR #894, opened 2026-07-29
Fix downloads of AICS FMS files stored on VAST

### LINKED ISSUE #821: Bug: Vast file ID encoded in download file name
## Description
When downloading a file with a vast path the pathing is encoded into the downloaded file name. For example downloading  `3500008658_20260518_10Xoverviews-02(6).czi` downloads as `b34/ff8/aef/363/746/90a/c48/11a/ff3/6d9/85/3500008658_20260518_10Xoverviews-02(6).czi`

### Reproduction 
Download the file here https://bff.allencell.org/app?c=file_name%3A0.4%2CKind%3A0.2%2CType%3A0.25%2Cfile_size%3A0.15&filter=%7B%22name%22%3A%22uploaded%22%2C%22value%22%3A%22RANGE%282025-05-29T07%3A00%3A00.000Z%2C2026-05-30T06%3A59%3A59.255Z%29%22%2C%22type%22%3A%22default%22%7D&filter=%7B%22name%22%3A%22file_name%22%2C%22value%22%3A%223500008658_20260518_10Xoverviews-02%286%29.czi%22%2C%22type%22%3A%22default%22%7D&source=%7B%22name%22%3A%22AICS+FMS%22%7D&sort=%7B%22annotationName%22%3A%22uploaded%22%2C%22order%22%3A%22DESC%22%7D

### LINKED ISSUE #878: Unable to download AICS local files from desktop app
## Description

Users are unable to download files stored on Vast (e.g., file id `7d7b7cd850b641199b7433d8f48fe480`) from the desktop app. The download appears to succeed, but only the placeholder file actually appears in their downloads folder.
I suspect this is related to #821 

Mentioned by Duncan in [this Slack thread](https://allencellscience.slack.com/archives/C021CEY96LD/p1784241837303919?thread_ts=1784139408.474289&cid=C021CEY96LD) as well as [here](https://allencellscience.slack.com/archives/C021CEY96LD/p1784246320504899?thread_ts=1784137830.673879&cid=C021CEY96LD) 

## Expected Behavior

The full file should download, the same way it does on web.

## Reproduction

On the desktop app, using AICS FMS as the source, filter for File ID `7d7b7cd850b641199b7433d8f48fe480`, and attempt to download the file.

## Environment
- [ ] Web
- [x] Desktop
   - OS: Not yet verified 
   - Version: 9.0.0

BrianWhitneyAI:
## Context

Fixes #878
Fixes #821

**#878 (desktop)** — On desktop the download path was taken from the `File Path (Local VAST)` annotation. That path is a POSIX path, not a URL, so it fell through to a branch that wraps string data in a Blob — saving the *path text* as the file's contents. Users got a small placeholder instead of the file, and the download reported success.

**#821 (web)** — We ran `encodeURIComponent` over the whole S3 key, percent-encoding the `/` separators so the browser named the file after the last URL segment — the entire key. `3500008658_20260518_10Xoverviews-02(6).czi` downloaded as `b34/ff8/aef/.../3500008658_20260518_10Xoverviews-02(6).czi`.

## Testing

- Updated the existing `formatAsHttpResource` expectations, which asserted the `%2F` behavior, and added a case covering characters that must still be escaped inside a segment (`path with spaces/a#b?c.txt`).
- Added a desktop test that a file of unknown size downloads in full rather than as 1 byte.

Manual — verified downloading a VAST-stored FMS file on both desktop and web.


packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -60,6 +61,16 @@ export default class FileDownloadServiceElectron extends FileDownloadService {
                 : this.downloadCloudDirectory(fileInfo, downloadRequestId, onProgress, destination);
         }
 
+        if (!fileInfo.data) {
+            if (!fileInfo.path.startsWith("http")) {
```
aswallace: will this cause issues for s3: files?

BrianWhitneyAI: oh for protocol files. yeah probably will investigate

BrianWhitneyAI: [a26e873](https://github.com/AllenInstitute/biofile-finder/pull/894/commits/a26e873cf24e345a996ae8794a4dd16056f96f30)

BrianWhitneyAI: oh weird I think the force push messed up that commit hash. anyways it should be fixed

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -60,6 +61,28 @@ export default class FileDownloadServiceElectron extends FileDownloadService {
                 : this.downloadCloudDirectory(fileInfo, downloadRequestId, onProgress, destination);
         }
 
+        if (!fileInfo.data) {
```
SeanDuHare: I think this duplicates or at least overwrites some of the logic below since this catches all `fileInfo.path` use cases

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -60,6 +61,28 @@ export default class FileDownloadServiceElectron extends FileDownloadService {
...
+            let downloadPath = fileInfo.path;
+            // S3 protocol URLs can't be requested directly, format as an https resource
+            if (downloadPath.startsWith("s3")) {
+                downloadPath =
+                    (await this.s3StorageService.formatAsHttpResource(downloadPath)) ||
+                    downloadPath;
+            }
+            if (!downloadPath.startsWith("http")) {
+                throw new DownloadFailure(
```
SeanDuHare: The electron version has access to the user's local file system so it should be able to reach a local file path if it has access to whatever file share the file is in - I think rather than fixing how the local download copy happens (which was below) this just restricts it to the cloud version (which is also fine, but I don't think we need to block local copies)

BrianWhitneyAI: Ok yeah I should have added more context on the decision here. I was thinking about two cases. 

1) FMS file with local vast path 
2) csv dataset pointing to local files. 

A few years ago we had a seperation for local files because we would route different download URLs depending on if the file was on VAST, I think thats where the local download actully came from. Now all our data is in cloud and its actully faster to download from there then copy the local file from the mount. So for this it makes sense to just use the cloud path. 

For csv local file we would really just be copying the file into their downloads which is fine but we have to open the save as file system window. I think this is not useful and we should just drop it. If they have a path locally, anything they want to do with the download they can already do

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: There was some weird path formatting before, I think this is cleaner but lmk if you have an alternate

packages/desktop/src/services/FileDownloadServiceElectron.ts
SeanDuHare: >For csv local file we would really just be copying the file into their downloads which is fine but we have to open the save as file system window. I think this is not useful and we should just drop it. If they have a path locally, anything they want to do with the download they can already do

We've been requested to have this feature so I don't think this is true. In the past its been mainly because:
a) they won't have access to the VPN/network during the work they want to do on the file
b) the connection is slower over VPN/network via Vast than on your hard drive

(files on Vast don't all exist in FMS)

packages/desktop/src/services/FileDownloadServiceElectron.ts
SeanDuHare: I think this is cleaner but it is still a duplicate of the work below - they should be de-duplicated to let the logic flow better - considering moving this into the if/else chain below

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: Ok I will add back in the local copy

BrianWhitneyAI: [f3c62e5](https://github.com/AllenInstitute/biofile-finder/pull/894/commits/f3c62e5de9d40793b632b38fbe5b5d4279546ca1)

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: [f3c62e5](https://github.com/AllenInstitute/biofile-finder/pull/894/commits/f3c62e5de9d40793b632b38fbe5b5d4279546ca1)

packages/desktop/src/services/FileDownloadServiceElectron.ts
```diff
@@ -83,17 +82,22 @@ export default class FileDownloadServiceElectron extends FileDownloadService {
...
+            return this.copyFile(fileInfo, downloadRequestId, onProgress, destination);
         }
 
+        throw new DownloadFailure(
+            `Unable to download ${fileInfo.name}. Only files with a remote location or a reachable local path can be downloaded; this file is at ${fileInfo.path}.`,
+            downloadRequestId
+        );
+    }
+
+    private saveDataToDisk(fileInfo: FileInfo, data: Uint8Array | Blob): DownloadResult {
```
SeanDuHare: This is a bit non-descriptive. Maybe `downloadBlobLikeFile` would be more in line with other functions in this

SeanDuHare (approved):

aswallace (approved):

packages/desktop/src/services/FileDownloadServiceElectron.ts
BrianWhitneyAI: [5977c7b](https://github.com/AllenInstitute/biofile-finder/pull/894/commits/5977c7b11682eb57d1165ca3fee6538c8b888a54)

PR #896, opened 2026-07-29
Allow users to select which columns are displayed

### LINKED ISSUE #865: Allow users to filter/select which columns are displayed in the file list
### Description
Feedback from Antoine Borensztejn on the new column-display behavior:

With the recent switch to a drag-and-drop column model, **all annotation columns are shown by default** and the only way to organize them is to drag or right-click a single column to "Move to start"/"Move to end" ([`Header.tsx`](packages/core/components/FileList/Header.tsx)). Users report this is frustrating when there are many columns:

- Most of the time users only need to see a few columns, but must reorganize by dragging across a long list.
- To view a specific annotation (e.g. local file path), users have to scroll through every column to find it, then move it to the front.
- The horizontal scrollbar becomes tiny/nearly invisible with many columns, and mouse users must shift-scroll or use that scrollbar.
- The **column search** functionality from the old system is missed.

BrianWhitneyAI:
Closes #865

## Description

Adds **"Modify columns"** to the header context menu, opening the `AnnotationPicker` to Select All / Clear All, recent annotations.

Showing every annotation remains the default.

## How the selection persists

`receiveAnnotationsLogic` previously rebuilt the column list from all annotations on every `RECEIVE_ANNOTATIONS`, which would silently undo a user's selection on refresh or source load. The fix:

- New `hasUserSelectedColumns` flag in selection state. When true, `receiveAnnotationsLogic` only prunes columns the data source no longer has.
- New `SELECT_COLUMNS` action + `selectColumnsLogic`
- The flag is encoded in the URL as `cs=true` and persisted on desktop. The 6-column URL cap is skipped when columns are user-selected, since that list is authoritative rather than regenerated on load. Im less sure about this we could probably do something dynamic here.

## Empty-header edge case

Clearing every column previously left a 0px header with nothing to right-click. The header now stays visible as an empty bar and a plain click opens the picker.


packages/core/state/selection/logics.ts
```diff
@@ -507,6 +550,11 @@ const decodeSearchParamsLogics = createLogic({
             dispatch(changeDataSources(sources));
             dispatch(setAnnotationHierarchy(hierarchy));
             columns && dispatch(setColumns(columns));
+            dispatch(
```
SeanDuHare: Do we need to track this as a variable in the state? Not sure I'm picking up the benefit of knowing if the user has selected columns vs just displaying whatever is in the URL / state at the time

packages/core/state/selection/logics.ts
```diff
@@ -482,6 +485,45 @@ const resizeColumnLogic = createLogic({
...
+        const currentColumns = selectionSelectors.getColumns(deps.getState());
+        const nameToAnnotationMap = metadata.selectors.getAnnotationNameToAnnotationMap(
+            deps.getState()
+        );
+        const annotationService = interaction.selectors.getAnnotationService(deps.getState());
+
+        const retainedColumns = currentColumns.filter((column) =>
+            selectedAnnotationNames.includes(column.name)
+        );
+        const addedAnnotationNames = selectedAnnotationNames.filter(
```
SeanDuHare, line 501: Annotations get silently filtered out here due to some weird state error - a warning notification could be useful

packages/core/state/selection/test/logics.test.ts
```diff
@@ -1319,6 +1321,69 @@ describe("Selection logics", () => {
...
     });
 
+    describe("selectColumns", () => {
+        const annotations = annotationsJson.map((annotation) => new Annotation(annotation));
+        const optimalWidth = 275;
+
+        beforeEach(() => {
+            sinon.stub(interaction.selectors, "getAnnotationService").returns(({
+                fetchOptimalWidthForAnnotations: (annotationsToSize: Annotation[]) =>
+                    Promise.resolve(new Map(annotationsToSize.map((a) => [a.name, optimalWidth]))),
```
SeanDuHare: This width is always 275 - having this vary per annotation would make for a more robust test

packages/core/components/FileList/Header.tsx
```diff
@@ -61,9 +62,32 @@ function Header(
         dispatch(selection.actions.sortColumn(columnName));
     };
 
+    const columnPickerMenuItems: ContextMenuItem[] = [
+        {
+            key: "available-annotations",
```
SeanDuHare: (nit) not super descriptive/accurate - perhaps just literally "column-picker"

packages/core/components/FileList/Header.tsx
```diff
@@ -88,6 +112,16 @@ function Header(
         dispatch(interaction.actions.showContextMenu(items, evt.nativeEvent));
     };
 
+    const onHeaderContextMenu = (evt: React.MouseEvent) => {
+        evt.preventDefault();
+        dispatch(interaction.actions.showContextMenu([modifyColumnsMenuItem], evt.nativeEvent));
+    };
+
+    // No columns means no header to right-click; a plain click on the empty header opens the picker.
```
SeanDuHare, line 106: This seems like a funky experience - either it should be impossible to have 0 columns selected (what we used to do) or the display for no columns should be explicit with a "No columns selected" display or similar

BrianWhitneyAI: I think that we need it since we have (select all/deselect all). IMO it wouldnt make sense to hit deselect all and go down to say just file name or something. We could add a 'No columns selected" dialogue in the main window like we do for no dataset selected

packages/core/components/FileList/Header.tsx
BrianWhitneyAI: [e054423](https://github.com/AllenInstitute/biofile-finder/pull/896/commits/e05442385776c578168f3857cd5408f670035291)

packages/core/state/selection/test/logics.test.ts
BrianWhitneyAI: [e054423](https://github.com/AllenInstitute/biofile-finder/pull/896/commits/e05442385776c578168f3857cd5408f670035291)

packages/core/state/selection/logics.ts
BrianWhitneyAI: [e054423](https://github.com/AllenInstitute/biofile-finder/pull/896/commits/e05442385776c578168f3857cd5408f670035291)

packages/core/state/selection/logics.ts
BrianWhitneyAI: This is for which side wins when receiveAnnotationsLogic runs. We rebuild the column list from all annotations on every RECEIVE_ANNOTATIONS, so without a signal that the current list is the user's choice rather than the generated default, a refresh or source load silently regenerates it and wipes the selection.

packages/core/state/selection/test/logics.test.ts
```diff
@@ -1319,6 +1321,87 @@ describe("Selection logics", () => {
...
+
+        beforeEach(() => {
+            sinon.stub(interaction.selectors, "getAnnotationService").returns(({
+                fetchOptimalWidthForAnnotations: (annotationsToSize: Annotation[]) =>
+                    Promise.resolve(
+                        new Map(
+                            annotationsToSize.map((a) => [a.name, optimalWidthByName.get(a.name)])
+                        )
+                    ),
+            } as unknown) as HttpAnnotationService);
```
aswallace: wondering if this `as unknown` is necessary or if it can be tightened/removed

packages/core/state/metadata/logics.ts
```diff
@@ -110,14 +114,36 @@ const receiveAnnotationsLogic = createLogic({
...
+            if (!isEqual(validUserSelectedColumns, currentColumns)) {
+                dispatch(selection.actions.setColumns(validUserSelectedColumns));
+            }
+            done();
+            return;
+        }
+
+        // Fall back to all annotations.
+        if (hasUserSelectedColumns) {
+            dispatch(selection.actions.setHasUserSelectedColumns(false) as AnyAction);
```
aswallace, line 146: can we do something similar if a user selects all? (e.g., validUserSelectedColumns = displayableAnnotationNames) 
Currently there doesn't seem to be a way to ever remove the `cs=true` flag after the user starts interacting with the menu, which means that we end up encoding all the column names in the search params even when we don't need to

packages/core/state/metadata/logics.ts
```diff
@@ -110,14 +114,36 @@ const receiveAnnotationsLogic = createLogic({
...
+        }
+
         // This request should be unable to take longer than 2 seconds
         const widthByAnnotation = await annotationService.fetchOptimalWidthForAnnotations(
             annotations
         );
 
-        const columns: Column[] = annotations
-            // Exclude parents of nested fields since they don't have their own values to display
```
aswallace, line 119: might be helpful to keep this comment for reference in the new place that we're using this line of code

packages/core/components/FileList/Header.tsx
```diff
@@ -61,9 +62,32 @@ function Header(
...
+        items: columnPickerMenuItems,
+    };
+
     const onHeaderColumnClick = (evt: React.MouseEvent, columnName: string) => {
         evt.preventDefault();
+        evt.stopPropagation();
         const items: ContextMenuItem[] = [
+            modifyColumnsMenuItem,
+            {
+                key: "modify-columns-divider",
```
aswallace: What does this divider do? I'm not sure if I'm seeing it in the menu

aswallace (approved):
My only major comment is that selecting all should revert "hasUserSelectedColumns" to false since this otherwise would have us encoding all the column names in the search params unnecessarily

packages/core/components/FileList/Header.tsx
SeanDuHare: Yep just needs something to prevent causing a weird experience

packages/core/state/selection/logics.ts
SeanDuHare: Rather than tracking this could we modify the behavior of the RECEIVE_ANNOTATIONS to avoid dispatching setColumns each time? As in, if the columns currently in the state still exist within the list of possible annotations found just keep them otherwise continue with wiping away the columns

packages/core/components/FileList/Header.tsx
BrianWhitneyAI: Lyndsay is working on some designs around the column movement and settings. This pr is already pretty long functionality wise so I think ill make a PR off of this branch with those changes and include the missing message.

packages/core/components/FileList/Header.tsx
BrianWhitneyAI: oh I didnt notice it stopped rendering. the idea was a split between modify and movement options. Im going to drop it for now then add the updated UX version in a followup

packages/core/state/selection/test/logics.test.ts
BrianWhitneyAI: It is not

packages/core/components/FileList/Header.tsx
BrianWhitneyAI: [565c1ff](https://github.com/AllenInstitute/biofile-finder/pull/896/commits/565c1ff42dd45788046843d466d9debf80db62ef)

packages/core/state/metadata/logics.ts
BrianWhitneyAI: [565c1ff](https://github.com/AllenInstitute/biofile-finder/pull/896/commits/565c1ff42dd45788046843d466d9debf80db62ef)

packages/core/state/metadata/logics.ts
BrianWhitneyAI: [565c1ff](https://github.com/AllenInstitute/biofile-finder/pull/896/commits/565c1ff42dd45788046843d466d9debf80db62ef)

packages/core/state/selection/test/logics.test.ts
BrianWhitneyAI: [565c1ff](https://github.com/AllenInstitute/biofile-finder/pull/896/commits/565c1ff42dd45788046843d466d9debf80db62ef)

packages/core/state/selection/logics.ts
BrianWhitneyAI: This accounts for "user picked a subset" and "the data source just gained a new annotation" look the same when comparing columns to annotations but the first should keep the columns as-is while the second should add the new column, so we need the flag to tell them apart. RECEIVE_ANNOTATIONS does already keep valid columns instead of wiping them

SeanDuHare (approved):

PR #902, opened 2026-08-04
Add discoverable compute pipeline modal with live JSS integration

### LINKED ISSUE #719: Pipeline Discovery
## Description 

As we add more pipelines we should shift away from specific development (manually adding pipelines) and allow for pipeline discovery. This would mean receiving a list active pipelines and metadata about their parameters then making them visible from the run pipelines dropdown.

BrianWhitneyAI:
## Summary

Resolves #719. Replaces hardcoded mock pipeline data with live discovery from the JSS `/1.0/compute/pipelines` API

## Notes

Tested against a dev deploy but jss pipeline isnt live yet.

## Test plan

- [x] Verify pipeline list loads from JSS in dev environment
- [x] Verify pipeline parameters render correctly from JSON Schema
- [x] Verify compute task submission succeeds end-to-end
- [x] Confirm modal handles JSS errors gracefully (network down, empty pipeline list)


SeanDuHare (approved):

aswallace (approved):

packages/core/constants/index.ts
```diff
@@ -115,5 +115,12 @@ export enum TemporaryFileServiceBaseUrl {
     TEST = "http://dev-aics-dtp-001.corp.alleninstitute.org:8080",
 }
 
+export enum JSSBaseUrl {
+    LOCALHOST = "http://localhost:9061",
+    STAGING = "https://stg-aics-api.corp.alleninstitute.org",
+    PRODUCTION = "https://aics-api.corp.alleninstitute.org",
+    TEST = "http://test-aics-api.corp.alleninstitute.org",
```
saeliddp, line 122: is this a real URL? I've never seen that before

packages/core/services/PipelineService/index.ts
```diff
@@ -3,36 +3,109 @@ import {
...
 
+interface JssPipelineSummary {
+    id: string;
+    name: string;
+    description: string;
+}
+
+interface JssPipeline extends JssPipelineSummary {
+    clusters: string[];
+    destinations?: string[];
```
saeliddp, line 18: destinations will always be defined

packages/core/services/PipelineService/index.ts
```diff
@@ -3,36 +3,109 @@ import {
...
+        const type = inferParameterType(prop);
+        return {
+            name,
+            label: (prop.title as string) ?? name,
+            description: (prop.description as string) ?? "",
+            type,
+            required: requiredFields.has(name),
+            default: (prop.default as number | string | null) ?? null,
+            options: type === "select" ? (prop.enum as unknown[]).map(String) : undefined,
+            validation: {
```
saeliddp, line 45: Note that this logic covers a small subset of the full JSON schema spec (e.g. validations: https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-01). On the other hand, JSS allows any valid JSON schema for parameter definition.

I'm not _too_ concerned about it at this point, but if we find ourselves wanting to use more JSON schema types/validations in the future, it might be worthwhile to investigate preexisting libraries that could handle some of this for us.

packages/core/services/PipelineService/index.ts
```diff
@@ -3,36 +3,109 @@ import {
...
+        acceptedExtensions: jss.allowedFileExtensions ?? [],
+        maxFileSizeBytes: jss.maxFileSizeBytes ?? null,
+        parameterSchema: jss.parameterSchema,
+    };
+}
+
 /**
  * Service responsible for fetching available compute pipelines and submitting
- * compute tasks via the FSS HTTP API.
+ * compute tasks via the JSS and FSS HTTP APIs.
```
saeliddp: 'and FSS'? Pipelines should only go through JSS after this change.

saeliddp (commented):
I'm not seeing anything related to pipeline submission via JSS--assuming you're waiting for a JSS deploy to staging before adding that?

packages/core/constants/index.ts
BrianWhitneyAI: No I missed that thanks for catching. I plan to test this against staging before merging so this branch hasnt been tested yet. Will update

BrianWhitneyAI: context, this was the jss dashboard url base

saeliddp: I think `aics-api...` and `aics...` are both valid to talk to JSS. Not sure if there is a preference between the two, but following the patterns in this file seems good anyways.

I was actually specifically asking about the `test-aics-api...` URL. I've never seen that used before, and it appears to be non-functional (?)

BrianWhitneyAI: Oh its supposed to be non functional. Test in this case is for non-real representations. string tracking etc

PR #903, opened 2026-08-04
Validate user against Active Directory before pipeline submission

### LINKED ISSUE #801: Verify AD user for pipeline submission
If the user does need to exist in AD - it is worth checking if there is something we could ping (anywhere) to verify that the user entered exists and adding a validation effect here

_Originally posted by @SeanDuHare in https://github.com/AllenInstitute/biofile-finder/pull/793#pullrequestreview-4300208402_

BrianWhitneyAI:
## Summary

Resolves #801 Calls MMS `/metadata-management-service/1.0/user?userName=<id>` on userId when user navigates away from user field

Need jss to release to try this out

## Test plan

- [x] Enter a valid AD username and blur — no error shown, submission proceeds
- [x] Enter an invalid username and blur — "User not found in Active Directory." error appears
- [ ] Enter a valid username and submit without blurring — validation fires at submit time
- [ ] Disconnect network and blur — "Unable to validate user. Please try again." error appears
- [ ] Changing the username after a successful validation resets validated state

### DIFF
--- a/packages/core/services/PipelineService/index.ts
+++ b/packages/core/services/PipelineService/index.ts
@@ -139,4 +139,17 @@ export default class PipelineService extends HttpServiceBase {
             dashboardUrl: response.data.dashboardUrl ?? "",
         };
     }
+
+    async validateUser(userName: string): Promise<boolean> {
+        const url = `${
+            this.metadataManagementServiceBaseURl
+        }/metadata-management-service/1.0/user?userName=${encodeURIComponent(userName)}`;
+        try {
+            await this.httpClient.get(url);
+            return true;
+        } catch (err: any) {
+            if (err?.response?.status === 404) return false;
+            throw err;
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


packages/core/services/PipelineService/index.ts
```diff
@@ -139,4 +139,17 @@ export default class PipelineService extends HttpServiceBase {
...
+    async validateUser(userName: string): Promise<boolean> {
+        const url = `${
+            this.metadataManagementServiceBaseURl
+        }/metadata-management-service/1.0/user?userName=${encodeURIComponent(userName)}`;
+        try {
+            await this.httpClient.get(url);
+            return true;
+        } catch (err: any) {
+            if (err?.response?.status === 404) return false;
+            throw err;
```
SeanDuHare, line 158: Do we want this validation failing on the server to block a pipeline submission? It doesn't seem critical to validate the user, rather a nice check to help fail fast - in the event MMS is down we could let them fail later down the line

SeanDuHare (approved):

packages/core/services/PipelineService/index.ts
BrianWhitneyAI: I think for tracking we want to. We have already had issues with storage where we have had to reach out via the submission field. This would ensure we could at least find the person.

aswallace (approved):

PR #908, opened 2026-08-19
Fix multi-file parquet row identity, and expose S3 URL parsing

SeanDuHare:
Precursor work to supporting `deltalake` / sharded parquets

### 1. Fix `hidden_bff_uid` collisions across multi-file parquet scans

`hidden_bff_uid` was selected as parquet_scan's `file_row_number`, which restarts
at 0 in every file. Any view spanning more than one parquet gave row 0 of file A and row 0 of
file B the same id.

Also fixes `convertDatabaseRowToFileDetail` rejected a falsy
uid, so row 0 raised "Missing auto-generated unique ID".

### 2. Expose S3 URL parsing on `S3StorageService`

Makes `parseUrl` and `ParsedUrl` public so callers outside the class can turn an
s3: URL into an addressable https one. No behavior change, this will be used in a follow-up.

### DIFF
--- a/packages/core/entity/SQLBuilder/index.ts
+++ b/packages/core/entity/SQLBuilder/index.ts
@@ -4,6 +4,18 @@ import { AnnotationType } from "../AnnotationFormatter";
 import { HIDDEN_UID_ANNOTATION } from "../../constants";
 
 /**
+ * Escape special characters for regex
+ */
+function toEscapedSqlString(str: string | number | boolean | null): string {
+    return (
+        `${str}`
+            .replace(/[-[\]{}()*+?.,\\^$|#\s]/g, "\\$&")
+            // Escape single-quotes for SQL string literal (ex. O'Reilly -> O''Reilly)
+            .replaceAll("'", "''")
+    );
+}
+
+/**
  * A simple SQL query builder.
  */
 export default class SQLBuilder {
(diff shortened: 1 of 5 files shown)


SeanDuHare:
@copilot re-review

packages/core/entity/SQLBuilder/index.ts
```diff
@@ -3,6 +3,18 @@ import { castArray } from "lodash";
 import { AnnotationType } from "../AnnotationFormatter";
 import { HIDDEN_UID_ANNOTATION } from "../../constants";
 
+/**
+ * Escape special characters for regex
+ */
+function toEscapedSqlString(str: string | number | boolean | null): string {
```
SeanDuHare, line 9: Body of function taken from `regexContains` and brought here to be re-used for `regexMatchValueInList`

packages/core/entity/SQLBuilder/index.ts
```diff
@@ -3,6 +3,18 @@ import { castArray } from "lodash";
...
+    return (
+        `${str}`
+            .replace(/[-[\]{}()*+?.,\\^$|#\s]/g, "\\$&")
+            // Escape single-quotes for SQL string literal (ex. O'Reilly -> O''Reilly)
```
BrianWhitneyAI, line 13: Is that example actually how that works? ' is the escape char?

BrianWhitneyAI (approved):

packages/core/entity/SQLBuilder/index.ts
aswallace: it is in duckdb, but just for escaping single quotes

aswallace (approved):

PR #909, opened 2026-08-20
Split LoadBalancerBaseUrl into FileStorageServiceBaseUrl and LabKeyBaseUrl

BrianWhitneyAI:
## Summary

`LoadBalancerBaseUrl` was one config value serving three different backends that happen to share the same gateway host: FSS2 file caching, FSS2 pipeline compute, and user-facing LabKey Plate UI links. This PR splits it into two enums so each concern can be configured independently values. Then in the pipelines PR we will implement a seperate JSS url base so that each base has its own purpose.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #910, opened 2026-08-20
Add developer modal for per-service environment overrides

BrianWhitneyAI:
## Summary

Stacked on #909.

Adds a developer tool for pointing individual services at a different environment (LOCALHOST / STAGING / PRODUCTION / TEST) — e.g., FSS2 at staging while everything else stays on production — so devs can better utilize our staging environments.

## How it works

- **Ctrl+Shift+E** toggles a "Service environments (dev)" modal. The shortcut is gated on `isAicsEmployee` 
- The modal has one dropdown per service (FES, MMS, FSS2, LabKey, Dataset Bucket, Vol-E), an "All services" shortcut dropdown, and a reset button. 
- Overrides are session-only; a reload returns to a known state.


packages/core/state/interaction/selectors.ts
```diff
@@ -100,13 +109,14 @@ export const getPipelineService = createSelector(
 );
 
 export const getMetadataManagementServiceBaseUrl = createSelector(
-    [getEnvironment],
-    (environment) => MMSBaseUrl[environment]
+    [getEnvironment, getEnvironmentOverrides],
+    (environment, overrides) =>
+        MMSBaseUrl[overrides[OverridableService.MetadataManagementService] ?? environment]
```
SeanDuHare, line 116: Why this override system rather than making environment an object that the modal would update?

dev-docs/03-using-localhost-datasource.md
```diff
@@ -13,3 +13,14 @@ and running on port `9081`. If you run `file-explorer-service` on one computer (
...
 2. From within the running Electron application, under the "Data Source" menu bar option, select "Localhost."
+
+Per-service environment overrides
+---------------------------------
+
+The "Data Source" menu points _every_ service at the selected environment. If you instead want to mix environments —
+e.g., a locally running `file-explorer-service` while MMS and FSS2 continue to point at staging — use the developer
+environment switch modal:
+
+1. Press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>E</kbd> (available to AICS employees only).
```
SeanDuHare, line 24: The variable in the system says isAicsEmployee for simplicity but devs here would probably want to know they actually need to connect and there is no real auth system
```suggestion
1. Press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>E</kbd> (available when connected to Allen network only).
```

BrianWhitneyAI: fair will add

packages/core/state/interaction/selectors.ts
```diff
@@ -75,43 +77,57 @@ export const getVisibleModal = (state: State) => state.interaction.visibleModal;
...
 export const getPipelineService = createSelector(
-    [getEnvironment],
-    (environment) =>
+    [getJSSBaseUrl, getMetadataManagementServiceBaseUrl],
+    (jssBaseUrl, metadataManagementServiceBaseUrl) =>
         new PipelineService({
-            jssBaseUrl: JSSBaseUrl[environment],
-            metadataManagementServiceBaseURl: MMSBaseUrl[environment],
+            jssBaseUrl,
+            metadataManagementServiceBaseURl: metadataManagementServiceBaseUrl,
```
aswallace, line 124: Is this a typo in the original/old config? @ capital R in `metadataManagementServiceBaseURl`

packages/core/components/Modal/EnvironmentSwitch/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+        });
+        dispatch(interaction.actions.setEnvironmentOverrides(newOverrides));
+        dispatch(interaction.actions.refresh());
+        onDismiss();
+    };
+
+    const body = (
+        <div>
+            <p className={styles.text}>
+                Point individual services at a different environment for this session. Services left
```
aswallace, line 127: Wondering if it can be made clearer that "for this session" means until page reload

aswallace (approved):

aswallace:
Minor thought about whether there are any unit tests that would be useful here

packages/core/state/interaction/selectors.ts
BrianWhitneyAI: Originally I was worried about users getting stuck somehow on the wrong services, I opted to move the tracking to state in [1d46db8](https://github.com/AllenInstitute/biofile-finder/pull/910/commits/1d46db85c67700164cfb99e2d539812ec4568468)

dev-docs/03-using-localhost-datasource.md
BrianWhitneyAI: [f235f1e](https://github.com/AllenInstitute/biofile-finder/pull/910/commits/f235f1e18722a28a33f85cd39c484491ef52cc23)

packages/core/state/interaction/selectors.ts
BrianWhitneyAI: Am I hallucinating? where is the 'R'

packages/core/components/Modal/EnvironmentSwitch/index.tsx
BrianWhitneyAI: switched to state so, arguably should change the wording here even more as a result

packages/core/state/interaction/selectors.ts
aswallace: The last one in `Url` for the key (not the value). Only noting bc `jssBaseUrl` is using property shorthand and mms BaseUrl can't bc of the capitalization

SeanDuHare (approved):

Issue #726, opened 2026-04-08
Feature: User selection for Annotation picking style in filter options


BrianWhitneyAI:
## Description

In working on #709 It came up that we default to Listpicker for string values under 100 total distinct entries. This led to a conversation about when the ListPicker was useful and when Range Pickers or fuzzy search was useful. We may want a way for users to define a switch to list picking for ranged annotations. For example a date range may be useful at some points where list selection of dates may be useful at other times

lynwilhelm:
@BrianWhitneyAI Do you know if there's a list/documentation of all the value types that lives somewhere? Realizing maybe that would be good to preview before we talk about this! Thao and I might be game to discuss sooner than the next BFF meeting.

lynwilhelm:
@aswallace Do have any thoughts on this? I basically just wasn't sure if the work Brian was doing required a closer look at different value types and what affordances we should serve up to the users.

...maybe @BrianWhitneyAI I was initially thinking this related to more than once value type but it's only one specific case???

BrianWhitneyAI:
Theres two definitions here, Types and Pickers. Type is referring to the native type of the input data from the data source. Theyre documented here (or at least the parquet ones are) https://github.com/apache/parquet-format/blob/master/LogicalTypes.md. These are less important to this ticket since we bucket these types into Date, Number, Text, Boolean which we have "Pickers" for each. 

Heres a breakdown of how we map these:

 | AnnotationType | Picker | Notes |                                                                                                                                                                                                                             
  |---|---|---|                                                                                                                                                                                                                                                                                                                                                                                                                                                          
  | `DateTime` | `DateRangePicker` | same component, time selection enabled |
  | `Number` | `NumberRangePicker` | min/max inputs, emits `RANGE()` operator |                                                                                                                                                                                   
  | `Text` | `ListPicker` or `SearchBoxForm` | `ListPicker` if ≤100 distinct values; `SearchBoxForm` (with fuzzy toggle) if 0 or >100 |
  | `Boolean` | `ListPicker` | shows raw true/false values |                                                                                                                                                                                                        
  | `Duration` | `ListPicker` | no dedicated picker |                                                                                                                                                                                                                                                                                                                                                                                                                                     

The question is should we switch to our ListPicker for some values over their more specific picker (`DateRangePicker` or `NumberRangePicker` etc) or provide a toggle or something.

SeanDuHare:
Considerations discussed:
[...]

PR #913, opened 2026-08-22
Allow toggling between list picker and search box for string annotations

BrianWhitneyAI:
## Context
Addresses #726. String annotations with more than 100 distinct values only rendered the exact-match search box, so users could not multi-select values (e.g. selecting several plate barcodes at once).

## Changes
- Added a toggle button (rendered below the picker) for string annotations that lets users switch between the `ListPicker` and the `SearchBoxForm` in both directions. Defaults are unchanged: list picker for 1–100 distinct values, search box otherwise.

- Note: rendering the full list for >100 values is cheap — `useAnnotationValues` already fetches all distinct values regardless, and `ListPicker` virtualizes rows past 100 items.

## Testing
- Two new unit tests: 150 values defaults to the search box and toggles to the list picker and back; 4 values defaults to the list picker and toggles to the search box. Each state asserts both the presence of the expected picker and the absence of the other.

### DIFF
--- a/packages/core/components/AnnotationFilterForm/index.tsx
+++ b/packages/core/components/AnnotationFilterForm/index.tsx
@@ -1,4 +1,4 @@
-import { IChoiceGroupOption } from "@fluentui/react";
+import { ActionButton, IChoiceGroupOption } from "@fluentui/react";
 import classNames from "classnames";
 import { isNil } from "lodash";
 import * as React from "react";
@@ -23,6 +23,10 @@ import styles from "./AnnotationFilterForm.module.css";
 interface AnnotationFilterFormProps {
     annotation: Annotation;
 }
+
+// Above this many distinct values, string annotations default to the search box
+// rather than the list picker; users can still toggle between the two.
+const MAX_VALUES_FOR_DEFAULT_LIST_PICKER = 100;
 
 /**
  * A form that provides a user the ability to select particular annotation values with which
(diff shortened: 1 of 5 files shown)


packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -194,23 +202,47 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
...
+                                onDeselectAll={onDeselectAll}
+                                onSearch={onSearch}
+                                fuzzySearchEnabled={fuzzySearchEnabled}
+                                fieldName={props.annotation.displayName}
+                                defaultValue={filtersForAnnotation?.[0]}
+                                hideFuzzyToggle={!canFuzzySearch}
+                            />
+                        )}
+                        {items.length > 0 && (
+                            <ActionButton
```
SeanDuHare, line 233: Can you switch back?

SeanDuHare (approved):
I think you should be able to toggle between the two - besides that this LGTM

packages/core/components/AnnotationFilterForm/index.tsx
BrianWhitneyAI: yeah I put it up on staging if you wanna try. I dont like the design but the functionality and performance seem good

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -59,6 +63,10 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
 
     const [filterType, setFilterType] = React.useState<FilterType>(defaultFilterType);
 
+    // User override for which picker to use for string annotations;
+    // undefined means no explicit choice has been made and the default applies
+    const [showListPicker, setShowListPicker] = React.useState<boolean | undefined>(undefined);
```
aswallace, line 68: this is functionally equivalent to `React.useState<boolean>()`, am fine with this though if it's a stylistic choice

aswallace (approved):

BrianWhitneyAI:
Had a conversation with Thao, she pointed out that the real functionality users want is just to do ONE OF ( CONTAINS "substring a", CONTAINS "substring b"). Instead of toggling between modals to find that functionality we should just add multi selection (with no prefetching for performance) to the string picker. Design in the works

PR #915, opened 2026-08-24
Add Delta Lake file handling

SeanDuHare:
## Context
`duckdb-wasm` does not yet support Delta Lake so we have to have some customized code here. The behavior here is intended to mirror how the Node `duckdb` works.

## Changes
Teaches DatabaseService to load a Delta Lake formatted Parquet file.
A Delta table is many parquet files rather than one, so this generalises the existing single-handle assumption: each source now maps to a list of DuckDB file handles, and the parquet view, DESCRIBE and parquet_metadata queries all scan that list.

`DeltaLakeService` is a no-op dummy for now - this branch will not be merged into `main` until downstream branches of this are approved and the integration is tested fully.

### DIFF
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -68,8 +68,7 @@ function isParquetBacked(type?: SourceType): boolean {
 }
 
 // A Delta Lake source is backed by many parquet files, so it registers one handle
-// per data file. The index is zero-padded to a fixed width so that no handle can
-// ever be a prefix of another (see FILE_HANDLE_SUFFIX above).
+// per data file
 function deltaFileHandleName(name: string, index: number): string {
     return `${fileHandleName(name)}-${String(index).padStart(8, "0")}.parquet`;
 }
@@ -324,8 +323,7 @@ export default abstract class DatabaseService {
     }
 
     /**
-     * Release DuckDB file handles. Overridable alongside registerFileURLs so tests
-     * can exercise the surrounding logic without a database.
+     * Release DuckDB file handles
      */
     protected async dropFileHandles(handles: string[]): Promise<void> {
         if (!this.database) return;
@@ -1126,9 +1124,8 @@ export default abstract class DatabaseService {
             selectParts.push(fileNameSelectPart);
         }
         // "file_row_number" restarts at 0 in every parquet file, so on its own it
-        // does not identify a row once the scan spans more than one file -- which
-        // is every Delta Lake table and every multi-source aggregate. Qualifying it
-        // with "filename" makes it unique. Kept VARCHAR even for a single file,
+        // does not identify a row once the scan spans more than one file. Qualifying
+        // it with "filename" makes it unique. Kept VARCHAR even for a single file,
         // both for consistency and because a 0-valued uid reads as falsy downstream.
         selectParts.push(
             `("filename" || '#' || CAST("file_row_number" AS VARCHAR)) AS "${HIDDEN_UID_ANNOTATION}"`
--- a/packages/core/services/DatabaseService/index.ts
+++ b/packages/core/services/DatabaseService/index.ts
@@ -1674,8 +1674,9 @@ const dataFileUrls = await this.deltaLakeService.listDataFiles(uri);
     // Similar to getColumnsOnDataSource below, but suitable for use during the
     // data source preparation step.
     private async getRawParquetColumns(handles: string[]): Promise<string[]> {
-        const sql = `DESCRIBE SELECT * FROM parquet_scan(
-            ${handleArrayLiteral(handles)}
+const sql = `DESCRIBE SELECT * FROM parquet_scan(
+            ${handleArrayLiteral(handles)},
+            union_by_name = true
         )`;
         const rows = await this.query(sql).promise;
         return rows.map((row) => row["column_name"] as string);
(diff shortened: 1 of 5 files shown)


packages/core/services/DatabaseService/index.ts
```diff
@@ -313,6 +339,67 @@ export default abstract class DatabaseService {
...
+        if (!this.database) {
+            throw new Error("Database failed to initialize");
+        }
+
+        await Promise.all(
+            urls.map((url, index) =>
+                this.database?.registerFileURL(
+                    handles[index],
+                    url,
+                    duckdb.DuckDBDataProtocol.HTTP,
```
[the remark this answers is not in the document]
SeanDuHare: S3 urls are converted into HTTP upstream of this

packages/core/services/DatabaseService/index.ts
```diff
@@ -75,6 +76,30 @@ function fileHandleName(name: string): string {
     return name + FILE_HANDLE_SUFFIX;
 }
 
+export type SourceType = typeof ACCEPTED_SOURCE_TYPES[number];
+// Return true if type is parquet or parquet-like
+function isParquetBacked(type?: SourceType): boolean {
```
BrianWhitneyAI, line 81: Would we ever receive a pointer to say one part of a DeltaLake that is a single parquet? Is that just user error? makes me think of multifile images

packages/core/services/DatabaseService/index.ts
```diff
@@ -313,6 +346,66 @@ export default abstract class DatabaseService {
...
+    }
+
+    /**
+     * Register every parquet data file of a Delta Lake table as its own DuckDB
+     * file handle, then expose them all as a single view.
+     */
+    private async addDeltaDataSource(name: string, uri: string | File): Promise<void> {
+        if (uri instanceof File) {
+            throw new Error(
+                `Delta Lake tables cannot be uploaded from your computer because they are ` +
```
BrianWhitneyAI, line 368: Provide a URL instead is a little weird wording. maybe Upload your file to a remote source

packages/core/services/DatabaseService/test/DatabaseService.test.ts
```diff
@@ -428,7 +436,173 @@ describe("DatabaseService", () => {
...
+                public isDelta = true;
+
+                constructor(
+                    private readonly parquetColumnsBySource: Record<string, string[]>,
+                    dataFiles: string[]
+                ) {
+                    const deltaLakeService = {
+                        listDataFiles: () => Promise.resolve(dataFiles),
+                        isDeltaTable: () => {
+                            this.deltaProbeCount += 1;
```
BrianWhitneyAI: what are these used for?

packages/core/services/DatabaseService/index.ts
```diff
@@ -75,6 +76,30 @@ function fileHandleName(name: string): string {
...
+
+// A Delta Lake source is backed by many parquet files, so it registers one handle
+// per data file
+function deltaFileHandleName(name: string, index: number): string {
+    return `${fileHandleName(name)}-${String(index).padStart(8, "0")}.parquet`;
+}
+
+// Render file handles as a DuckDB list literal, e.g. ARRAY['a', 'b'].
+function handleArrayLiteral(handles: string[]): string {
+    return `ARRAY[${handles.map((handle) => `'${handle}'`).join(", ")}]`;
```
BrianWhitneyAI, line 93: escape chars for quotes?

BrianWhitneyAI (approved):

packages/core/services/DatabaseService/index.ts
SeanDuHare: Yes & no but it would still be a directory but it could just be one parquet underneath that directory + the delta log transaction history

packages/core/services/DatabaseService/test/DatabaseService.test.ts
SeanDuHare: whoop! old artifact, I was testing how many times a transaction log was pulled but the test felt too low level so I removed it

packages/core/services/DatabaseService/index.ts
SeanDuHare: For the `'${handle}'` portion? That is intentionally unescaped in order to escape the handle name in SQL

BrianWhitneyAI: wouldnt a quote in handle break the sql /allow for someone to input something weird

aswallace (approved):

packages/core/services/DatabaseService/index.ts
SeanDuHare: Yes to potential SQL injection, though we are unguarded from that in several places - definitely something we should look into addressing across the board, but this case isn't different than how we already allow inputs.

No to breaking the SQL - the opposite actually

packages/core/services/DatabaseService/index.ts
SeanDuHare: Done in e8e15f4c

packages/core/services/DatabaseService/test/DatabaseService.test.ts
SeanDuHare: Done in 38a31b9e945b2765571ea087eadc5905e7c8a9a1

PR #916, opened 2026-08-25
Derive data source type on the fly

SeanDuHare:
## Context
This adds support for detecting Delta Lake type files.

## Changes
Type resolution for files (csv, json, parquet, delta) now happens as lazily as possible. This type is largely just needed by the `DatabaseService` to perform various actions. A similar follow-up for removing `name` may be possible later, however this was necessary now because detecting if a file is a Delta Lake requires a network call.

## Testing
Tested with various deltalake files using the next PR - deltalake files were detected as expected and resolutions triggered once in the network tab

### DIFF
--- a/packages/core/components/DataSourcePrompt/FilePrompt.tsx
+++ b/packages/core/components/DataSourcePrompt/FilePrompt.tsx
@@ -8,7 +8,12 @@ import { useDispatch } from "react-redux";
 import MarkdownPreview from "./MarkdownPreview";
 import { SecondaryButton, TertiaryButton, TransparentIconButton } from "../Buttons";
 import Tooltip from "../Tooltip";
-import { Source, getNameAndTypeFromSourceUrl, isMarkdownType } from "../../entity/SearchParams";
+import {
+    Source,
+    getNameFromSourceUrl,
+    getResourceNameFromSourceUrl,
+    isMarkdownType,
+} from "../../entity/SearchParams";
 import { ParsedFrontmatter, processMarkdown } from "../../entity/MarkdownFrontMatter";
 import { interaction } from "../../state";
 
@@ -33,10 +38,10 @@ export default function FilePrompt(props: Props) {
     const [shouldHaveFrontmatter, setShouldHaveFrontmatter] = React.useState(false);
     const { onSelectFile } = props;
 
-    // Parse markdown files to provide a preview of the metadata we're able to find
+    // Parse markdown files to provide a preview of the metadata we're able to find.
     const handleMarkdownSource = React.useCallback(
-        (source: Source) => {
-            if (isMarkdownType(source.type)) {
+        (source: Source, extension?: string) => {
+            if (isMarkdownType(extension)) {
                 // Calls the standalone process instead of going through the DB service
                 // since we don't want to cache the result yet
                 processMarkdown(source, false) // skip source normalization since just previewing the data
@@ -63,13 +68,12 @@ export default function FilePrompt(props: Props) {
             const selectedFile = acceptedFiles?.[0];
             if (selectedFile) {
                 // Grab name minus extension
-                const nameAndExtension = selectedFile.name.split(".");
-                const name = nameAndExtension.slice(0, -1).join("");
+                const segments = selectedFile.name.split(".");
+                const name = segments.length > 1 ? segments.slice(0, -1).join(".") : segments[0];
                 // Extension validation is handled by the component itself
-                const extension = nameAndExtension.pop();
-                const source = { name, type: extension, uri: selectedFile };
+                const source = { name, uri: selectedFile };
                 onSelectFile(source);
-                handleMarkdownSource(source);
+                handleMarkdownSource(source, segments.length > 1 ? segments.pop() : undefined);
             }
         },
         [onSelectFile, handleMarkdownSource]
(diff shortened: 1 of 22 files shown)


packages/core/components/DataSourcePrompt/FilePrompt.tsx
```diff
@@ -131,7 +138,7 @@ export default function FilePrompt(props: Props) {
                     <Tooltip content={props.selectedFile.name}>
                         <p className={styles.selectedFile}>
                             {props?.fileLabel}
-                            {props.selectedFile.name}.{props.selectedFile.type}
+                            {props.selectedFile.name}
```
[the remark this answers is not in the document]
SeanDuHare: This is intended

packages/core/services/DatabaseService/index.ts
```diff
@@ -685,14 +706,20 @@ export default abstract class DatabaseService {
         dataSources: Source[],
         skipNormalization = false
     ): Promise<void> {
-        const markdownSources = dataSources.filter((source) => isMarkdownType(source.type));
+        const sourcesWithTypes = await Promise.all(
+            dataSources.map((dataSource) => this.withResolvedType(dataSource))
```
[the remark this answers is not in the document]
SeanDuHare: This shouldn't ever happen nor should it even be possible to reach this state. The top level URIs are cached by this class so these underlying sources, if ever supplied in duplicate, would only ever be novel one at a time.

packages/core/services/DatabaseService/test/DatabaseService.test.ts
```diff
@@ -522,13 +514,69 @@ describe("DatabaseService", () => {
                 "https://example.com/table/part-00001.snappy.parquet",
             ];
 
+            it("identifies an untyped URL as delta by probing for _delta_log", async () => {
+                // getNameAndTypeFromSourceUrl leaves type undefined for a URL with
+                // no recognized extension; this is where that gets settled.
```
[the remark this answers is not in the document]
SeanDuHare: Removed entirely in 8a38ba2e98ed518c0d057c92e4369a727ec445e7

packages/core/entity/SearchParams/index.ts
```diff
@@ -448,19 +464,46 @@ export default class SearchParams {
...
+        sourceType: SourceType | undefined,
+        location: string
+    ) {
+        // A Delta Lake table is a directory rather than a file, and pandas has no
+        // reader for it, so it needs the deltalake package.
+        if (sourceType === "delta") {
+            return {
+                imports: "from deltalake import DeltaTable # pip install deltalake\n",
+                code: `df = DeltaTable(${location}).to_pandas().astype('str')`,
+            };
```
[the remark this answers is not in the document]
SeanDuHare: True! Addressed in 06146f1b6105e1e00e8c8bde6e945269b4371f19

packages/core/entity/SearchParams/index.ts
```diff
@@ -448,19 +464,46 @@ export default class SearchParams {
         return `\`${filter.name}\`=="${filter.value}"`;
     }
 
-    private static convertDataSourceToPython(source: Source | undefined, userOS: string) {
+    /**
+     * The Python that loads a data source into a dataframe.
+     */
+    private static convertDataSourceToPythonRead(
```
BrianWhitneyAI, line 484: This is just for python code snippet generation yea?

packages/core/services/DatabaseService/index.ts
```diff
@@ -76,12 +77,31 @@ function fileHandleName(name: string): string {
...
 function isParquetBacked(type?: SourceType): boolean {
     return type === "parquet" || type === "delta";
 }
 
+/**
+ * The reader implied by a file name or URL's extension, or undefined when there
+ * is no recognized extension
+ */
+function sourceTypeFromExtension(nameOrUrl: string): SourceType | undefined {
+    const extension = getResourceNameFromSourceUrl(nameOrUrl).split(".").pop();
```
BrianWhitneyAI, line 92: A dotless final path segment is treated as a file extension

BrianWhitneyAI (approved):

packages/core/entity/SearchParams/index.ts
SeanDuHare: Yep

packages/core/entity/SearchParams/index.ts
```diff
@@ -5,6 +5,7 @@ import FileSort, { SortOrder } from "../FileSort";
 import type { DatasetSources } from "../MarkdownFrontMatter";
 import { AICS_FMS_DATA_SOURCE_NAME } from "../../constants";
 import { Column } from "../../state/selection/actions";
+import type { SourceType } from "../../services/DatabaseService";
```
aswallace: I'm a little wary that this entity file is importing from a service, especially since DatabaseService also imports from SearchParams

aswallace (approved):

packages/core/services/DatabaseService/index.ts
SeanDuHare: Yea thats fine! The line below this one checks to see if the `extension` found on this line is one we recognize ("csv", "json", "parquet", "delta") and if so will return that so unless a user names their file "csv" and its actually some other type entirely this shouldn't be a problem. This entire check is just a quick-path to resolve the type without probing, later on the real type gets resolved via DuckDB (or the custom code in the case of Delta Lake)

packages/core/entity/SearchParams/index.ts
SeanDuHare: I moved SourceType to SearchParams to avoid this, lmk if this doesn't resolve it or you dislike the change!
ae685da591b095510f2227af088406852058b667

aswallace: The change looks good to me!

PR #918, opened 2026-08-25
Implement Delta Lake reader

SeanDuHare:
## Context
This is the last in the stack of PRs related to adding support for Delta Lake parquet files in BFF

## Description
This change set implements the actual reader which is separated into its own class DeltaLakeService. To understand how this works, I'll first explain the gist of how Delta Lake works.

A Delta Lake is a directory containing a) the parquet files that represent the data and b) specially named parquet files that are logs of additions/removals of data from the aforementioned actual parquet files. Readers of Delta lake start from a "checkpoint" AKA essentially a starter parquet and apply additions/removals from there based on the "commits" (AKA changes) that have happened over time.

The service itself isn't trying to read the parquet files just figure out which parquet files should be read. With that said this essentially reads the first parquet file from the "checkpoint" and reads through the transaction log to figure out which other parquet files are needed from there.

## Testing
I've been using some Delta Lakes I custom made. Waiting for Kevin to get back to test out against his FMS data. An earlier version of his data was tested but from the sound of it may be quite different than what it looks like now so this needs to be triple checked before considered working.

[Example](http://localhost:9001/app?c=File+Name%3A15%2CAcquisition+Date%3A15%2CAlgorithm%3A15%2CAnalysis+Date%3A15%2CArchived%3A15%2CBackground%3A15&source=%7B%22name%22%3A%22test-delta-mock+%288%2F28%2F2026+4%3A33%3A51+PM%29%22%2C%22uri%22%3A%22s3%3A%2F%2Fstaging-biofile-finder-datasets%2Ftest-delta-mock%2F%22%7D&showNulls=false)


packages/core/services/DeltaLakeService/index.ts
```diff
@@ -1,26 +1,310 @@
...
+
+        const { activePaths, nextVersion } = await this.seedFromCheckpoint(logUrl);
+        await this.applyCommits(logUrl, nextVersion, activePaths);
+
+        if (isEmpty(activePaths)) {
+            throw new Error(
+                `No parquet data files were found in the Delta Lake table at "${rootUrl}".`
+            );
+        }
+        return [...activePaths].map((path) => `${baseUrl}/${path}`);
```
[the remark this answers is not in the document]
SeanDuHare: hmm I didn't think this was true, I'll double check with the documentation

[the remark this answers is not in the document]
SeanDuHare: It is! Fixed

packages/core/services/DeltaLakeService/index.ts
```diff
@@ -1,26 +1,310 @@
...
+        const activePaths = new Set<string>();
+        for (const action of await this.readCheckpoint(checkpointUrl)) {
+            DeltaLakeService.assertSupported(action);
+            if (action.add) activePaths.add(action.add.path);
+        }
+        if (activePaths.size > MAX_DELTA_DATA_FILES) {
+            throw new Error(
+                `This Delta Lake table contains more than ${MAX_DELTA_DATA_FILES} parquet ` +
+                    `files, more than this application can load at once. Consider compacting ` +
+                    `the table.`
```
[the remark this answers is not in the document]
SeanDuHare: Fixed, though unlikely for our use

packages/core/services/DeltaLakeService/index.ts
```diff
@@ -1,26 +1,310 @@
...
+        logBase: string
+    ): Promise<{ activePaths: Set<string>; nextVersion: number }> {
+        const version = await this.fetchCheckpointVersion(logBase);
+        if (isNil(version)) return { activePaths: new Set(), nextVersion: 0 };
+
+        const checkpointUrl = `${logBase}/${padToVersionWidth(version)}.checkpoint.parquet`;
+        const activePaths = new Set<string>();
+        for (const action of await this.readCheckpoint(checkpointUrl)) {
+            DeltaLakeService.assertSupported(action);
+            if (action.add) activePaths.add(action.add.path);
```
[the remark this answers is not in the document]
SeanDuHare: Feels like overkill to test this narrow case

packages/core/services/DeltaLakeService/test/DeltaLakeService.test.ts
```diff
@@ -0,0 +1,183 @@
+import { createMockHttpClient } from "@aics/redux-utils";
```
BrianWhitneyAI, line 1: AI Comment: no test in this suite exercises `seedFromCheckpoint` with actual add actions. worth adding a case where readCheckpoint returns [{ add: { path: "part-0.parquet" } }]

packages/core/services/DeltaLakeService/index.ts
```diff
@@ -1,26 +1,313 @@
...
+        const readerVersion = action.protocol?.minReaderVersion || 0;
+        if (readerVersion >= 3) {
+            throw new Error(
+                `This Delta Lake table requires reader version ` +
+                    `${readerVersion}, which this application does not support.`
+            );
+        }
+        if (action.metaData?.configuration) {
+            const columnMapping =
+                action.metaData?.configuration instanceof Map
```
BrianWhitneyAI, line 82: AI Comment: configuration arrives from JSON.parse or DuckDB's parquet deserializer. neither produces a native Map

packages/core/services/DeltaLakeService/index.ts
```diff
@@ -1,26 +1,313 @@
...
-     *
-     * Checks for presence of conventional log and checkpoint files,
-     * but does not guarantee the table is readable beyond that.
+     * Reject an action that uses a feature this reader cannot honour.
+     */
+    private static assertSupported(action: DeltaAction): void {
+        const readerVersion = action.protocol?.minReaderVersion || 0;
+        if (readerVersion >= 3) {
+            throw new Error(
+                `This Delta Lake table requires reader version ` +
```
BrianWhitneyAI, line 76: will this mean anything to users? not sure if I get what this is for. can we control this from somewhere else? feels like bad pattern.

packages/core/services/DeltaLakeService/test/DeltaLakeService.test.ts
```diff
@@ -0,0 +1,183 @@
...
+import { expect } from "chai";
+import * as sinon from "sinon";
+
+import DeltaLakeService from "..";
+import S3StorageService from "../../S3StorageService";
+
+const TABLE_URL = "https://files.example.com/data/table";
+const LOG = `${TABLE_URL}/_delta_log`;
+
+function commit(version: number): string {
```
BrianWhitneyAI, line 11: commit? but it returns a string? what is this doing?

packages/core/services/DeltaLakeService/index.ts
```diff
@@ -1,26 +1,313 @@
...
-    public async listDataFiles(_url: string): Promise<string[]> {
-        return [];
+    public async listDataFiles(rootUrl: string): Promise<string[]> {
+        const baseUrl = await this.toHttpBase(rootUrl);
+        const logUrl = `${baseUrl}/${DELTA_LOG_DIR}`;
+
+        const { activePaths, nextVersion } = await this.seedFromCheckpoint(logUrl);
+        await this.applyCommits(logUrl, nextVersion, activePaths);
+
+        if (isEmpty(activePaths)) {
```
BrianWhitneyAI, line 163: does this have the same behavior for an empty table vs no found files?

packages/core/services/DeltaLakeService/index.ts
```diff
@@ -1,26 +1,313 @@
...
+        if (action.add?.deletionVector) {
+            throw new Error(
+                `This Delta Lake table uses deletion vectors, so some rows in its parquet ` +
+                    `files are marked deleted without being rewritten. This application ` +
+                    `cannot read it yet.`
+            );
+        }
+    }
+
+    private static parseAction(
```
BrianWhitneyAI, line 101: so this function parses json into some action against the table? and returns some object that the table understands? Some documentation would be helpful here

BrianWhitneyAI (commented):
Is there some existing deltalake support we can use rather than doing all the fetching and reading ourself? Sequential HTTP fetching in applyCommits awaits one commit file at a time. A table with 200 commits since its last checkpoint makes 200 sequential round trips before returning any data. Additionally the service functions are really hard for me to follow here

packages/core/services/DeltaLakeService/index.ts
SeanDuHare: The error here serves to explain the error is a) with the format of the Delta Lake and b) can be fixed by changing their file. 
We could add support for these versions but since we don't need it internally, I see no reason to do so now.

packages/core/services/DeltaLakeService/test/DeltaLakeService.test.ts
SeanDuHare: The Delta Lake "commits" are string IDs, this creates the commit ID from the number using the pattern declared in the Delta Lake docs

packages/core/services/DeltaLakeService/index.ts
SeanDuHare: No, the checkpoint files from which this gets seeded should per the docs have at minimum one file so at this point the table should contain something

packages/core/services/DeltaLakeService/test/DeltaLakeService.test.ts
SeanDuHare: Hmm - I may be missing something, but a few of the tests rely on `seedFromCheckpoint` returning an add path

SeanDuHare:
> Is there some existing deltalake support we can use rather than doing all the fetching and reading ourself?

Nope, not for duckdb-wasm. This entire stack models what duckdb core does, but until someone implements support for it in duckdb-wasm we'll need our own to read Delta Lake. 

> Sequential HTTP fetching in applyCommits awaits one commit file at a time. A table with 200 commits since its last checkpoint makes 200 sequential round trips before returning any data.  Additionally the service functions are really hard for me to follow here

In general, not saving out a checkpoint frequently would be a problem for us memory wise (in the ETL) so I don't think this will be an issue, but it is something we can keep an eye on.

packages/core/services/DeltaLakeService/index.ts
SeanDuHare: Removed in 642651ce635b188f68ffe238d3a548c06f75697e

packages/core/services/DeltaLakeService/index.ts
SeanDuHare: Added a comment in 642651ce635b188f68ffe238d3a548c06f75697e.

Yes, Delta Lake transaction logs are JSON objects representing actions the DB took against the table. This takes in the raw JSON and parses it out into either an "add"/"remove" action or `null` if neither.

packages/core/services/DeltaLakeService/test/DeltaLakeService.test.ts
```diff
@@ -0,0 +1,183 @@
...
+                await service.listDataFiles(TABLE_URL);
+            } catch (err) {
+                caught = err;
+            }
+            expect((caught as Error)?.message).to.include("No parquet data files");
+        });
+    });
+
+    describe("hosts that answer 403 for a missing key", () => {
+        it("detects a table whose missing log files (403 case)", async () => {
```
aswallace: very small nit: "who is" or "that is" vs whose

packages/core/services/DeltaLakeService/index.ts
```diff
@@ -1,26 +1,316 @@
...
+            urlAsHttp = url;
+        }
+        if (replaceTrailingSlashes) return urlAsHttp.replace(/\/+$/, "");
+        return urlAsHttp;
+    }
+
+    /**
+     * GET a log file.
+     *
+     * Returns null for missing file, which is how the commit walk below
```
aswallace: commit walk below? is this referring to `applyCommits`?

aswallace (approved):

packages/core/services/DeltaLakeService/index.ts
SeanDuHare: Yea it is, but I can see how this is a weird comment - removed in favor of simplicity

packages/core/services/DeltaLakeService/test/DeltaLakeService.test.ts
SeanDuHare: Changed to "that is"

PR #921, opened 2026-08-26
Bundle multi-file web downloads into a single zipped download

### LINKED ISSUE #920: Multi-file selections on web only download the first file
## Description

When multiple files are selected in the web app and downloaded (e.g. via the context menu), only the first file actually downloads.

## Cause

Each selected file was downloaded via its own programmatic anchor click, but browsers only permit one download per user gesture. By the time the anchors are clicked, the original gesture has been consumed by intervening async work (fetching selection details, S3 size lookups), so downloads 2..N count as "automatic downloads" and are blocked behind Chrome's easy-to-miss "download multiple files" permission prompt. The click stagger added in #894/#901 does not address this: it only spaced out navigation commits, and results varied per user depending on whether the automatic-downloads permission had previously been granted for the site.

## Fix

Bundle multi-file selections into a single streamed ZIP download using the existing `StreamedZipDownloader` (already used for zarr directory downloads), so one gesture produces one download. This also restores real aggregate progress reporting and cancellation, which fire-and-forget anchor clicks could not provide.

BrianWhitneyAI:
Resolves #920

## Problem

Multi-file selections on web (chrome) only downloaded the first file. The rest were blocked behind Chrome's "download multiple files" permission prompt. The click stagger from #894/#901 didn't (and couldn't) fix this; it only appeared fixed on machines where the automatic-downloads permission had already been granted (like mine...)

## Solution

Zip Downloads! 

## Testing
Added unit tests for `downloadFilesAsZip`

- [x] Manual check multi-select download on web produces one ZIP
- [x] Cancel works
- [x] single file isnt zipped
- [x] desktop downloads seperate files 
- [x] multifile zip with zarrs works 

### DIFF
--- a/packages/core/services/FileDownloadService/index.ts
+++ b/packages/core/services/FileDownloadService/index.ts
@@ -58,6 +58,15 @@ export default abstract class FileDownloadService extends HttpServiceBase {
     ): Promise<DownloadResult>;
 
     /**
+     * Download multiple files bundled together as a single ZIP archive.
+     */
+    downloadFilesAsZip?(
+        files: FileInfo[],
+        downloadRequestId: string,
+        onProgress?: (bytesDownloaded: number) => void
+    ): Promise<DownloadResult>;
+
+    /**
      * Retrieves the file system's default download location.
      */
     abstract getDefaultDownloadDirectory(): Promise<string>;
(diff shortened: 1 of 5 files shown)


aswallace (approved):
Tested locally on mine as well! 
I know this is only needed for web, but would there be any advantage to also implementing for desktop?

BrianWhitneyAI:
> Tested locally on mine as well! I know this is only needed for web, but would there be any advantage to also implementing for desktop?

potentially not spamming someones download folder with a ton of files?

SeanDuHare (approved):
Would be nice to make this consistent across web and desktop but LGTM!

PR #926, opened 2026-08-31
Change the number range picker default to overall max+1

### LINKED ISSUE #816: Numeric range default may be off by one
## Use case

* User's table has a column A with values 0, 1, and 2.
* Click to add a filter on column A
* In the modal to select the filter bounds, the defaults are:
  * Min (inclusive): 0
  * Max (exclusive): 2
* So if the user clicks "Submit" with the default options, they'll get the rows with values 0 and 1.

I think the current default might incorrectly suggest to the user that the upper bound is inclusive.

## Suggest a Solution
Default to an upper bound that includes all the data, like 3 or 2.001

aswallace:
## Context
The numeric range filter's maximum field was defaulting to the overall maximum, which meant that values were getting excluded when that default filter was applied.
closes #816 

## Changes
Keeps the filter behavior the same (inclusive min, exclusive max) but changes the default max value to overall maximum + 1 when the modal opens/resets. For example, if full range of values is [100, 101, 102, 103], then when the modal opens the fields will auto-populate with min (inclusive) = 100 and max (exclusive) = 104

Initially considered doing some step-size inference to account for the magnitude of the values (e.g., 0.001 for small values, 1 for integers), but it wasn't worth the effort vs impact trade-off for just a default field value. Also, + 1 was actually already the fallback behavior when the max field is empty, so it keeps the UI self-consistent 

## Testing
Manual testing + updated unit tests

## Note
We may eventually move towards inclusive/inclusive range filters (see #931), in which case this would just be a temporary fix for this component

### DIFF
--- a/packages/core/components/NumberRangePicker/index.tsx
+++ b/packages/core/components/NumberRangePicker/index.tsx
@@ -45,6 +45,10 @@ export default function NumberRangePicker(props: NumberRangePickerProps) {
     const overallMax = React.useMemo(() => {
         return items.at(-1)?.value?.toString() ?? "";
     }, [items]);
+    // On component load, default to slightly more than max so that values aren't excluded
+    const defaultMax = React.useMemo(() => {
+        return overallMax ? (Number(overallMax) + 1).toString() : "";
+    }, [overallMax]);
     const overallMinDisplay = items[0]?.displayValue?.toString() ?? overallMin;
     const overallMaxDisplay = items.at(-1)?.displayValue?.toString() ?? overallMax;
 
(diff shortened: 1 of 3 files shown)


packages/core/components/NumberRangePicker/index.tsx
```diff
@@ -45,22 +45,26 @@ export default function NumberRangePicker(props: NumberRangePickerProps) {
     const overallMax = React.useMemo(() => {
         return items.at(-1)?.value?.toString() ?? "";
     }, [items]);
+    // On component load, default to slightly more than max so that values aren't excluded
+    const defaultMax = React.useMemo(() => {
+        return overallMax ? (Number(overallMax) + 1).toString() : "";
+    }, [overallMax]);
```
pgarrison: I suspect this `useMemo` is not worth it, since the computation is so simple and the result is a string. (I'd say the same for the two `useMemo`s above it.)

pgarrison (approved):
Nice!

SeanDuHare (approved):

packages/core/components/NumberRangePicker/index.tsx
aswallace: good point, removed in [f89839b](https://github.com/AllenInstitute/biofile-finder/pull/926/commits/f89839bd914a2b1a1dff127ec61b069070ffc4a6)

PR #927, opened 2026-09-01
(Very small) Consider any filter incompatible with hierarchy filters of same name

### LINKED ISSUE #906: Fuzzy search not working as expected
From Caroline:
I filtered by fuzzy search, and then organized my data and found that some things were organized where they should not be...
I filtered by "ROCK" then organized by "lumenoid condition" and "plate barcode." Something went wrong...The metadata associated with the files looks correct in the right hand menu, but files are mis-organized in the "folders" given in the middle panel. See the link provided - the "lumenoid condition" is "2D Lumenoid formation..." but the file shows up in the "folder" labeled "3D ROCK inhibitor 50 uM."

If I instead search for a specific lumenoid condition, I no longer have this problem, so I think it is something to do with fuzzy search?

([link to view](https://bff.allencell.org/app?c=file_name%3A30%2CKind%3A15%2CType%3A20%2Cfile_size%3A10%2C520+or+570%3F%3A15%2CAbsence+Of+Migrating+Cells+Coming+From+Colony+Out+Of+FOV+At+Time+Of+Migration%3A15&group=Lumenoid+Condition&group=Plate+Barcode&filter=%7B%22name%22%3A%22Lumenoid+Condition%22%2C%22value%22%3A%22ROCK%22%2C%22type%22%3A%22fuzzy%22%7D&openFolder=%5B%223D+ROCK+inhibitor+50+uM%22%5D&openFolder=%5B%223D+ROCK+inhibitor+50+uM%22%2C%223500009013%22%5D&source=%7B%22name%22%3A%22AICS+FMS%22%7D&prov=%7B%22name%22%3A%22AICS%22%2C%22type%22%3A%22csv%22%2C%22uri%22%3A%22https%3A%2F%2Fbiofile-finder-datasets.s3.us-west-2.amazonaws.com%2FFMS.provenance.csv%22%7D&sort=%7B%22annotationName%22%3A%22uploaded%22%2C%22order%22%3A%22DESC%22%7D))

SeanDuHare:
Small change to avoid letting filters like fuzzy/any/none filters override hierarchy filters when querying for files in a file set. Without this change the wider fuzzy search query for example is selected over the narrower equality filter from the hierarchy group.

**Before**
Note the "Lumenoid Condition" mismatch between the group and the actual value for the file
Screenshot 2026-09-01 at 12 01 43 PM

**After**
Note that only the groups themselves are filtered by the fuzzy filter rather than the files themselves since the group filter is narrower
Screenshot 2026-09-01 at 12 02 02 PM

### DIFF
--- a/packages/core/components/DirectoryTree/useDirectoryHierarchy.tsx
+++ b/packages/core/components/DirectoryTree/useDirectoryHierarchy.tsx
@@ -194,11 +194,7 @@ const useDirectoryHierarchy = (
                         // and also have filters applied for that field (e.g., barcode=1234, barcode=1357),
                         // then at the barcode=1234 group we should remove the barcode=1357 filter
                         const nonHierarchyFilters = selectedFileFilters.filter(
-                            (f) =>
-                                !(
-                                    take(hierarchy, depth + 1).includes(f.name) &&
-                                    f.type === FilterType.DEFAULT
-                                )
+                            (f) => !take(hierarchy, depth + 1).includes(f.name)
                         );
                         const filters = uniqWith(
                             [...hierarchyFilters, ...nonHierarchyFilters],


aswallace (approved):
Ooh that's an interesting bug, makes sense

BrianWhitneyAI (approved):

PR #932, opened 2026-09-01
Avoid error when annotations are loading

### LINKED ISSUE #944: Invalide validation error when switching data sources
## Context
The error seen below occurs when switching data sources (or sometimes on refreshing). Due to this being a race between the renderer and the db query it isn't seen consistently, but the steps to reproduce are below the screenshot
Screenshot 2026-09-01 at 2 21 03 PM

### How to reproduce

- Set up two queries on parquets with different schemas (the more different the better)
- Select a file so the details pane is populated, then switch to the other query. changeQueryLogic dispatches changeDataSources in its first batch (files start loading immediately), but requestAnnotations only fires later so the file is compared against the old schema thereby rendering an error.
- Watch the error notifications: Unable to find column metadata for field "<Section> : <Field>"

Or you can hard-reload with a local query source already selected. Annotations start as [], so if the file details pane paints first, every top-level field in every section errors at once. Throttling the network in devtools makes this reliable.

SeanDuHare:
## Context
The error seen below occurs when switching data sources (or sometimes on refreshing). Due to this being a race between the renderer and the db query it isn't seen consistently, but the steps to reproduce are below the screenshot
Screenshot 2026-09-01 at 2 21 03 PM

### How to reproduce

- Set up two queries on parquets with different schemas (the more different the better)
- Select a file so the details pane is populated, then switch to the other query. changeQueryLogic dispatches changeDataSources in its first batch (files start loading immediately), but requestAnnotations only fires later so the file is compared against the old schema thereby rendering an error.
- Watch the error notifications: Unable to find column metadata for field "<Section> : <Field>"

Or you can hard-reload with a local query source already selected. Annotations start as [], so if the file details pane paints first, every top-level field in every section errors at once. Throttling the network in devtools makes this reliable.

## Changes
Clear annotations when data sources change and in the component check if the annotations are loaded before considering anything an error.

## Testing
Tested using steps above across several data sources (downloading any of the data sources from the data source page works with this too)

### DIFF
--- a/packages/core/components/FileDetails/MetadataList/test/MetadataList.test.tsx
+++ b/packages/core/components/FileDetails/MetadataList/test/MetadataList.test.tsx
@@ -144,6 +144,58 @@ describe("<MetadataList />", () => {
             );
         });
 
+        it("does not report a missing annotation while annotations are loading", () => {
+            // Arrange
+            const fileWithUnknownField = () =>
+                new FileDetail(
+                    {
+                        file_path: "path/to/MyFile.txt",
+                        file_id: "abc123",
+                        file_name: "MyFile.txt",
+                        file_size: 7,
+                        uploaded: "01/01/01",
+                        annotations: [{ name: "Unknown Field", values: ["some value"] }],
+                    },
+                    Environment.TEST
+                );
+
+            function renderWith(annotations: Annotation[]) {
+                const { store, actions } = configureMockStore({
+                    state: mergeState(initialState, {
+                        metadata: {
+                            annotations,
+                        },
+                        interaction: {
+                            platformDependentServices: {
+                                executionEnvService: new ExecutionEnvServiceNoop(),
+                            },
+                        },
+                    }),
+                });
+                render(
+                    <Provider store={store}>
+                        <MetadataList isLoading={false} file={fileWithUnknownField()} />
+                    </Provider>
+                );
+                return actions;
+            }
+
+            const isMissingMetadataError = (action: any) =>
+                JSON.stringify(action?.payload ?? "").includes(
+                    "Unable to find column metadata for field"
+                );
+
+            // (sanity-check) check if error is present first
+            let actions = renderWith(TOP_LEVEL_FILE_ANNOTATIONS);
+            expect(actions.list.filter(isMissingMetadataError)).to.not.be.empty;
+
+            // Act
+            actions = renderWith([]);
+
+            // Assert
+            expect(actions.list.filter(isMissingMetadataError)).to.be.empty;
+        });
+
         it("has loading message when file is downloading", async () => {
             // Arrange
             class FakeExecutionEnvService extends ExecutionEnvServiceNoop {
--- a/packages/core/components/FileDetails/MetadataList/test/MetadataList.test.tsx
+++ b/packages/core/components/FileDetails/MetadataList/test/MetadataList.test.tsx
@@ -144,58 +144,6 @@ describe("<MetadataList />", () => {
             );
         });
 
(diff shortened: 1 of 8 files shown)


packages/core/state/metadata/reducer.ts
```diff
@@ -34,6 +35,10 @@ export default makeReducer<MetadataStateBranch>(
             ...state,
             annotations: action.payload,
         }),
+        [CHANGE_DATA_SOURCES]: (state) => ({
+            ...state,
+            annotations: [],
```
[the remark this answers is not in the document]
SeanDuHare: Fixed in 16c4a7e

packages/core/state/metadata/logics.ts
```diff
@@ -49,9 +50,17 @@ const requestAnnotations = createLogic({
...
+            if (selection.selectors.getSelectedDataSources(getState()) !== sources) return;
             dispatch(receiveAnnotations(annotations));
         } catch (err) {
             console.error("Failed to fetch annotations", err);
+            dispatch(
+                interaction.actions.processError(
+                    "requestAnnotations",
+                    `Failed to load column metadata for ${sources}. ${err}`
+                )
+            );
```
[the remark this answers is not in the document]
SeanDuHare: Fixed in 92ad373

packages/core/state/interaction/logics.ts
```diff
@@ -723,13 +723,16 @@ const refresh = createLogic({
...
+            const sources = selection.selectors.getSelectedDataSources(getState());
             const annotationService = interactionSelectors.getAnnotationService(getState());
 
             // Refresh list of annotations & which annotations are available
             const [annotations, availableAnnotations] = await Promise.all([
                 annotationService.fetchAnnotations(),
                 annotationService.fetchAvailableAnnotationsForHierarchy(hierarchy),
             ]);
+            // A late response would overwrite the newer source's schema with this one's.
+            if (selection.selectors.getSelectedDataSources(getState()) !== sources) return;
```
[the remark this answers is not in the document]
SeanDuHare: Fixed in 92ad373

packages/core/components/FileDetails/MetadataList/test/MetadataList.test.tsx
```diff
@@ -144,6 +144,58 @@ describe("<MetadataList />", () => {
             );
         });
 
+        it("does not report a missing annotation while annotations are loading", () => {
```
BrianWhitneyAI: This feels like a regression test do we need it for this case?

packages/core/state/metadata/test/logics.test.ts
```diff
@@ -60,6 +62,50 @@ describe("Metadata logics", () => {
             // assert
             expect(actions.includesMatch({ type: RECEIVE_ANNOTATIONS })).to.be.true;
         });
+
+        it("drops a response that arrives after the data source changed", async () => {
```
BrianWhitneyAI, line 66: duplicate test?

BrianWhitneyAI (approved):

packages/core/state/metadata/test/logics.test.ts
SeanDuHare: Between this and interaction/logics.test.ts? They test the same thing but for two different actions

packages/core/components/FileDetails/MetadataList/test/MetadataList.test.tsx
SeanDuHare: Hmm fair it might just be clutter to test that the error doesn't trigger in this case, removed

aswallace (approved):

PR #940, opened 2026-09-03
De-duplicate indexes on reselection

### LINKED ISSUE #874: Pagination counter and behavior is funky
## Description

If you select multiple files, then hold SHIFT+ Up arrow key to "page" through selected files, the counter total number math becomes wrong. So, 9/10 ends up being 9/11.

## Expected Behavior

_What did you expect to happen instead?_

Pagination/files counter should always remain accurate.
I notice that SHIFT+up/down keys will add more files to the total selection, which seems like the right behavior, but counter still needs to be accurate.

## Reproduction

See above

## Environment
Web, chrome, mac

SeanDuHare:
## Context
Lyndsay found a bug (resolves #874)
>If you select multiple files, then hold SHIFT+ Up arrow key to "page" through selected files, the counter total number math becomes wrong. So, 9/10 ends up being 9/11.

## Description
Essentially, the index range did account for duplicates when they are wholly subsets not partial subsets so when Lyndsay did the action above a partial range was added in duplicate.

## Testing
Seems fine locally, unit tests added to cover this

## Before
https://www.awesomescreenshot.com/video/56216061?key=5ac80b03c73bf6c7aa1f7efd3799aac2

## After

https://github.com/user-attachments/assets/799116e4-a9e1-4623-ac02-66b6c00323a0

### DIFF
--- a/packages/core/entity/FileSelection/test/FileSelection.test.ts
+++ b/packages/core/entity/FileSelection/test/FileSelection.test.ts
@@ -79,6 +79,43 @@ describe("FileSelection", () => {
             ).to.equal(true);
         });
 
+        it("does not double count file rows that are already selected", () => {
+            // Arrange
+            const fileSet = new FileSet();
+            const selection = new FileSelection().select({
+                fileSet,
+                index: new NumericRange(5, 14),
+                sortOrder: 0,
+            });
+
+            // Act: re-select a file row within the existing selection
+            const nextSelection = selection.select({ fileSet, index: 13, sortOrder: 0 });
+
+            // Assert
+            expect(nextSelection.count()).to.equal(10);
+            expect(nextSelection.getFocusedItemIndices().indexAcrossAllSelections).to.equal(8);
+        });
+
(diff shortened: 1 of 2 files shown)


packages/core/entity/FileSelection/test/FileSelection.test.ts
```diff
@@ -79,6 +79,43 @@ describe("FileSelection", () => {
             ).to.equal(true);
         });
 
+        it("does not double count file rows that are already selected", () => {
```
BrianWhitneyAI, line 82: Also feels like a regression test, it shouldnt count in the first place

BrianWhitneyAI (approved):

packages/core/entity/FileSelection/test/FileSelection.test.ts
SeanDuHare: because this causes really weird output downstream (like when saving a manifest) I'm in favor of keeping this test to make sure this entity performs its main function (keeping track of files) as accurate as possible

aswallace (approved):

Issue #469, opened 2025-03-11
Require a 'Program' annotation before proceeding with a download request


hughes036:
## Use Case

We want every file in cache to be annotated with the Program it is downloaded for (the idea came from the weekly VAST meeting - stakeholders are Jess and Vicki).  This will allow us to track how much storage each program is using, which is necessary to policy making and resource allocation on the PM team and leadership.

## Suggest a Solution

When a file is requested for download to VAST, proceed if the Program annotation exists on it.  Otherwise, prompt the user to add it.  Only proceed once they have done so.

## Additional Details

This relates to https://github.com/aics-int/aics-file-upload-app/issues/186

Many pre-existing FMS files will lack this annotation, however.

SeanLeRoy:
@hughes036 Metadata editing is currently only available to data leads - would this mean data leads are the only ones able to download then?

SeanDuHare:
@hughes036 @jessicasyu is this still important? This functionality may get replaced by the pipeline if we move caching into that feature for example

jessicasyu:
The end goal (files in cache should be annotated with a Program) is still important. I don't necessarily feel strongly about how we implement/enforce that. I imagine we could still do something like display a warning/error if a file doesn't have a Program annotation if caching were moved to the pipeline?

PR #950, opened 2026-09-14
feat: Make separate component for status messages

ShrimpCryptid:
## Context
Part 1 of 2 for #469. I split out this minor component refactor for ease of review.

*Estimated review size: tiny, 5 minutes*

## Changes
- Refactors the error cards in `PasswordForm` into the separate `StatusMessageCard` component, which can be used inline.

## Testing
- Tested by opening edit metadata panel and typing in an incorrect password (see below).

## Screenshots (if relevant)

{2893B583-2316-462C-B2C1-4B3436E7C1D3}


BrianWhitneyAI (commented):
Im wondering about moving this to a generalized component. Here theres only one case that triggers the error and in the added case in the next stacked case is only warnings, I feel a little like we are overengineering to bridge two things.

ShrimpCryptid:
> Im wondering about moving this to a generalized component. Here theres only one case that triggers the error and in the added case in the next stacked case is only warnings, I feel a little like we are overengineering to bridge two things.

Okay, that's fair. In TFE I've used a similar error message card across a few different modals for user feedback. Will BFF add more text/data entry modals in the future?

BrianWhitneyAI:
> Okay, that's fair. In TFE I've used a similar error message card across a few different modals for user feedback. Will BFF add more text/data entry modals in the future?

Actually, I think there are more cases that this could be useful for. the compute pipeline modal has a similar reject message for invalid user or params and I have a draft Pr for query toggling that also does something similar. I scraped the rest of the repo and these were additional cases that we could maybe use for 

```
  - EditMetadata .delete-warning-section (the byte-for-byte duplicate)
  - DataSourcePrompt .warning
  - ComputePipelineModal .error-banner/.success-banner/.param-error
  - NewAnnotationPathway .error-message/.warning-message
  ```

Another thing is that this is fairly similar to `StatusMessage` component. @aswallace @SeanDuHare whatcha think?

**Comment by aswallace** [comment]
I do think it's reasonable to have a separate component, since this case is slightly different than the purpose of `StatusMessage` (independent pop-up modal vs inline warning/errors). We'll want to make sure their styling aligns though; it looks like we use both ` --error-status-background-color` +  `--error-status-text-color` and `--error-background-color` + `--error-text-color` in various combinations in different places and I can't remember from just skimming whether the difference was intentional. May need to look back at the Figma global styling definitions

File: packages/core/components/StatusMessageCard/StatusMessageCard.module.css [hunk 1]
```diff
@@ -0,0 +1,13 @@
+.warning-message {
+    background-color: var(--warning-status-background-color);
+    border: var(--warning-background-color) 1px solid;
```
SeanDuHare: This variable is a bit misleading if it is only intended to be used for the border in combination with `warning-status-background-color` - perhaps `warning-border-color`?

SeanDuHare (approved):
LGTM! (Outside of Anya's concern)

SeanDuHare:
> > Okay, that's fair. In TFE I've used a similar error message card across a few different modals for user feedback. Will BFF add more text/data entry modals in the future?
> 
> Actually, I think there are more cases that this could be useful for. the compute pipeline modal has a similar reject message for invalid user or params and I have a draft Pr for query toggling that also does something similar. I scraped the rest of the repo and these were additional cases that we could maybe use for
> 
> ```
>   - EditMetadata .delete-warning-section (the byte-for-byte duplicate)
>   - DataSourcePrompt .warning
>   - ComputePipelineModal .error-banner/.success-banner/.param-error
>   - NewAnnotationPathway .error-message/.warning-message
> ```
> 
> Another thing is that this is fairly similar to `StatusMessage` component. @aswallace @SeanDuHare whatcha think?

I don't see why not - though if this isn't used more than 2 places I'd put the component as a sibling of whatever sole file (PasswordForm in this case) it is intended to be used in.

aswallace (approved):
LGTM too! 

re my other comment, after looking at Figma (e.g., [this component](https://www.figma.com/design/oNQvOAK11YINcbPvNmkgwO/BFF---File-Explorer?node-id=3861-11318&t=7cHIvcSB6ToNBiqU-0)) I'd be in favor of unifying the styling with our other [status messages](https://www.figma.com/design/oNQvOAK11YINcbPvNmkgwO/BFF---File-Explorer?node-id=1392-9384&t=7cHIvcSB6ToNBiqU-0) to just use `--error-status-background-color` for the background and `--error-status-text-color` for both the text and the border, and likewise `--warning-status-text-color` for both text & border.

I suspect `--error-background-color` is legacy from the old design and should eventually be removed globally, but that's probably out of scope for this ticket.

ShrimpCryptid:
> > Another thing is that this is fairly similar to `StatusMessage` component. @aswallace @SeanDuHare whatcha think?
> 
> I don't see why not - though if this isn't used more than 2 places I'd put the component as a sibling of whatever sole file (PasswordForm in this case) it is intended to be used in.

This will be used `packages/core/components/Modal/CopyFileModal` in the next PR in the chain.

Issue #578, opened 2025-08-29
Add import pattern to linting


aswallace:
We don't currently document or lint the pattern for imports in BFF, which is generally:
- external -- alphabetized by module name
- space
- internal -- by relative path distance and then alphabetized, e.g., `./`, then `../a`, then `../b`, then `../../a` etc
- space
- style.css files 
- IF we're in a .test.ts or .test.tsx file, the component that's being tested

_Originally posted by @aswallace in https://github.com/AllenInstitute/biofile-finder/pull/566#discussion_r2310898560_

PR #951, opened 2026-09-14
feat: Show warning when program annotation is missing for copied files

### LINKED ISSUE #469: Require a 'Program' annotation before proceeding with a download request

ShrimpCryptid:
## Context
Closes #469, "Require a 'Program' annotation when downloading files."

## Changes
- Adds a status message that appears when copying files to VAST and one or more files are missing program annotations.
  - Adds a warning icon in a column next to the files, and sorts them to the top.
- Fixed typos in CSS variable definitions in `FileList.module.css`.

## Testing
Manually tested on BFF desktop on Windows.

## Screenshots

Files with program annotation:
{C9C9F196-E3AC-44ED-8F4A-FCA9C5BA527E}

Files without program annotation:
{34BE3EEA-7394-4CCB-81CD-2E94E605AEEC}

Mixed files:
{40A24967-42E3-4B79-9D97-8DDAD867BCB0}


packages/core/components/FileList/FileList.module.css
```diff
@@ -1,6 +1,6 @@
 .container {
     --row-count-margin-top: 0.05em;
-    --row-count-intrisic-height: 12px; /* margin values are relative to this because it is used in font-size declaration */
```
ShrimpCryptid, line 3: Fixed a typo here

packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
```diff
@@ -93,32 +93,38 @@
...
 
 .footer-buttons {
     display: flex;
     justify-content: flex-end;
     gap: 8px;
 }
 
+
 .summary {
+    --row-count-intrinsic-height: 12px; /* margin values are relative to this because it is used in font-size declaration */
```
ShrimpCryptid, line 118: This variable is shared with `FileList.module.css`. I wasn't sure if this should go in some sort of root CSS file?

packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
```diff
@@ -93,32 +93,38 @@
...
+
     display: flex;
     justify-content: flex-end;
     gap: 1em;
-    margin-top: var(--row-count-margin-top);
+    margin-top: 0.25em;
```
ShrimpCryptid, line 123: Adjusted this a little taller than what's in `FileList.module.css`.

packages/core/components/Modal/CopyFileManifest/index.tsx
```diff
@@ -9,6 +10,7 @@ import AnnotationName from "../../../entity/Annotation/AnnotationName";
 import FileDetail from "../../../entity/FileDetail";
 import FileSelection from "../../../entity/FileSelection";
 import { interaction, selection } from "../../../state";
+import StatusMessageCard from "../../StatusMessageCard";
```
aswallace: nit: our linter doesn't handle import order unfortunately :/ this should go on line 9 after `../../Buttons`

packages/core/components/Modal/CopyFileManifest/index.tsx
```diff
@@ -25,6 +27,17 @@ function ResponsiveFileName({ fileName }: { fileName: string }) {
...
 }
 
+/**
+ * Formats a quantity string, using either the singular or plural form.
+ * @param quantity count of items.
+ * @param singular the singular form of the item name
+ * @param plural the plural form of the item name
+ * @returns a string formatted as "{quantity} {singular/plural}"
+ */
+function formatQuantityString(quantity: number, singular: string, plural: string): string {
```
aswallace: not 100% convinced this needs its own function, but open to it!

packages/core/components/Modal/CopyFileManifest/index.tsx
```diff
@@ -83,6 +128,19 @@ function FileTable({ files, title }: { files: FileDetail[]; title: string }) {
...
+            <div>
+                {areFilesMissingProgramAnnotation && (
+                    <StatusMessageCard type="warning">
+                        {formatQuantityString(
+                            filesWithoutProgramAnnotation.size,
+                            "file is",
+                            "files are"
+                        )}{" "}
+                        missing the &quot;Program&quot; annotation. Files will soon be required to
+                        have this field defined to be downloaded to VAST.
```
aswallace: I find the phrasing of the second sentence slightly confusing but I can't pinpoint exactly why, possibly the mismatch between field & annotation or the chain of `to`s. Maybe something like "Files will soon be required to have this annotation before they can be downloaded to VAST" or "This annotation will soon be required for files to be downloaded to VAST/cached in VAST"?

aswallace (approved):

packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
aswallace: it could! our root css variables are in global.css

packages/core/components/Modal/CopyFileManifest/index.tsx
ShrimpCryptid: 👀 Would you all want to install a plugin to do that? I have the prettier import sorting plugin setup on TFE, and there's also an eslint import sorting rule....

aswallace: I looked into this briefly last year, but wasn't able to find one that worked out of the box with our current import pattern. Definitely would be happy to have one though! Here's [the (closed) ticket for it](https://github.com/AllenInstitute/biofile-finder/issues/578)

packages/core/components/Modal/CopyFileManifest/CopyFileManifest.module.css
SeanDuHare: Hmm that is odd, I'd be in favor of moving it out

packages/core/components/Modal/CopyFileManifest/index.tsx
```diff
@@ -50,6 +63,23 @@ function FileTable({ files, title }: { files: FileDetail[]; title: string }) {
         return totalBytes ? filesize(totalBytes) : "Calculating...";
     };
 
+    // Show warning for files missing the `Program` annotation as an extra column
+    // in the table + a warning message below it.
+    const filesWithoutProgramAnnotation = files.reduce((acc, file) => {
+        const programAnnotation = file.getAnnotation("Program");
+        if (!programAnnotation || programAnnotation[0] === "None") {
```
SeanDuHare, line 60: `None` as an option 😵‍💫

packages/core/components/Modal/CopyFileManifest/index.tsx
```diff
@@ -50,6 +63,23 @@ function FileTable({ files, title }: { files: FileDetail[]; title: string }) {
...
+            acc.add(file.id);
+        }
+        return acc;
+    }, new Set<string>());
+    const areFilesMissingProgramAnnotation = filesWithoutProgramAnnotation.size > 0;
+
+    // Reorder files so that those missing the `Program` annotation appear first
+    const orderedFiles = [
```
SeanDuHare: Could just use a sorting function:
`const orderedFiles = files.sort((file) => filesWithoutProgramAnnotation.has(file.id) ? 1 : -1)`
but I do like the readability of this so 🤷

SeanDuHare (approved):

packages/core/components/Modal/CopyFileManifest/index.tsx
ShrimpCryptid: Opened a PR as #965!

packages/core/components/Modal/CopyFileManifest/index.tsx
ShrimpCryptid: I'd love to stick this in a utils file somewhere, but it does feel silly if this is currently the only use case 🤭

PR #953, opened 2026-09-15
Allow the open source datasets to use markdown sources

aswallace:
## Context
Since we now have a way of rendering dataset info via markdown files in the app, I wanted to add those markdown files to the public datasets that we load from the [Open Source Datasets page](https://bff.allencell.org/datasets). Those datasets (e.g., the Drug Perturbation dataset) sometimes provide their own preset queries, which allows them to default to a particular view of the files. However, the loading function currently assumes that the published data source itself is the only source, even if the preset query contains its own sources, so when I put markdown sources in the preset queries they just get overwritten. 

## Changes
For open source datasets that provide a preset query, if that query contains sources, default to using those. Otherwise, use the dataset url itself as the source.

## Testing
Tested manually:
- Added a markdown source to the Variance dataset's preset query and loaded that
- Loaded the Drug Perturbation dataset (has a preset query that doesn't contain any sources)
- Loaded the AIND Image Atlas dataset (does not have a preset query)


packages/core/util/test/googleSheets.test.ts
```diff
@@ -136,8 +136,8 @@ describe("googleSheets", () => {
 
     describe("isGoogleSheetUri", () => {
         it("recognizes a Sheets URL", () => {
-            expect(isGoogleSheetUri(`https://docs.google.com/spreadsheets/d/${SHEET_ID}/edit`)).to.be
-                .true;
+            expect(isGoogleSheetUri(`https://docs.google.com/spreadsheets/d/${SHEET_ID}/edit`)).to
+                .be.true;
```
aswallace, line 140: Unrelated; the linter auto-changed this during the merge process

BrianWhitneyAI (approved):

packages/web/src/components/OpenSourceDatasets/index.tsx
```diff
@@ -29,13 +29,13 @@ export default function OpenSourceDatasets() {
...
         datasetName: string,
-        source: Source,
+        sources: Source[],
         url?: Partial<SearchParamsComponents>
     ) => {
         dispatch(
             selection.actions.addQuery({
-                name: `New ${source.name} Query on ${datasetName || "open-source dataset"}`,
-                parts: { ...url, sources: [source] },
+                name: `New query on ${datasetName || "open-source dataset"}`,
```
SeanDuHare: It looks like `datasetName` is required, is this to catch when it is empty or can it be removed?

SeanDuHare (approved):

packages/web/src/components/OpenSourceDatasets/index.tsx
aswallace: Ah true, I think it can be removed

PR #956, opened 2026-09-16
Surface server error details when pipeline submission fails

### LINKED ISSUE #955: Surface server error details when pipeline submission fails
## Problem

When a pipeline submission fails in the compute pipeline modal (e.g. OME Zarr Conversion), the UI shows only a generic banner:

> **Submission failed**
> An error occurred while submitting the job.

The JSS API actually returns an informative error body that we discard. For example, submitting 180 files returns:

```json
{
  "error": "Validation failure: filePaths - size must be between 1 and 150",
  "hostname": "UnknownHost",
  "time": "2026-09-15T18:29:42.676053929Z",
  "message": null,
  "cause": null,
  "responseType": "CLIENT_ERROR"
}
```

The user has no way to know the fix is simply "select 150 or fewer files" — they just see that something went wrong.

BrianWhitneyAI:
## Summary

Closes #955

When a pipeline submission failed, the compute pipeline modal always showed a generic "An error occurred while submitting the job." banner, even though JSS returns an informative detail in the response body's `error` field (e.g. `Validation failure: filePaths - size must be between 1 and 150`).

- **`PipelineService.submitComputeTask`**: catches the axios error and rethrows the JSS `error` detail as the error message when present; other errors pass through unchanged.
- **`ComputePipelineModal`**: renders the caught error's message in the failure banner (generic fallback when there's no message) and includes it in the error toast.

- [x] Tested locally with 151+ files


BrianWhitneyAI:
Screenshot 2026-09-16 at 2 03 49 PM

packages/core/services/PipelineService/test/PipelineService.test.ts
```diff
@@ -0,0 +1,67 @@
...
+                await service.submitComputeTask(request);
+                expect(false, "Expected to throw").to.be.true;
+            } catch (e) {
+                expect((e as Error).message).to.equal(
+                    "Validation failure: filePaths - size must be between 1 and 150"
+                );
+            }
+        });
+
+        it("rethrows errors without an 'error' detail unchanged", async () => {
```
aswallace: I'm not sure if I understand what this test name is saying

aswallace (approved):

packages/core/services/PipelineService/test/PipelineService.test.ts
BrianWhitneyAI: the JSS errors return a response with 'error' as the message we want to convey, we could also receive something like a network error so this test checks that those still surface

aswallace: ahh the phrasing had me confused. It rethrows errors as is if they don't have an `error` property

BrianWhitneyAI: yeah Ill update it

BrianWhitneyAI: [ea53ed3](https://github.com/AllenInstitute/biofile-finder/pull/956/commits/ea53ed32dec1fc7d9e528440150a46da089d5e73)

SeanDuHare (approved):

PR #954, opened 2026-09-15
feat: Replace fuzzy search checkbox with operator dropdown in filter form

BrianWhitneyAI:
## Summary

Implements the Figma proposal "Suggestions to improve Fuzzy Search terminology & interactions", reworking how users filter on an annotation's values.

### Filter form UX
- **Operator dropdown** ("Exactly matches" / "Contains") replaces the "Fuzzy search (non-exact matching)" checkbox, defaulting to Contains. Each option has a hover tooltip describing what it does, and the dropdown is joined to the search box as a single control.
- **Search / Browse list tabs** replace the value-count-based either/or between the list picker and the search box. The default tab still depends on value count (≤100 values opens Browse list).
- **Committed values render as removable chips** under a label matching their operator ("Exactly matches:" / "Contains:").
- **Same-operator submissions accumulate** (OR-joined); submitting with a different operator **replaces** the annotation's filters, with an inline warning beforehand.
- Query sidebar wording aligned: `EXACTLY MATCHES` / `CONTAINS`, with multiple contains filters labeled `CONTAINS ONE OF`.

## Testing
- New/updated unit coverage: operator submit types, input clearing on commit, chip render/remove/clear-all, warning only on operator mismatch, default tab by value count, and accumulate-vs-replace semantics.
- Manually verified against the AICS FMS data source in the running app (text, number, date, datetime, boolean, and duration annotations).

## Follow-ups (stacked PRs)
- #959 — Browse list tab for number / date / datetime range pickers
- #961 — make "File Path (Local VAST)" filterable
- #962, #963 — filter-form formatting perf and value fetching for slash-named annotations (independent, off `main`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #959, opened 2026-09-18
feat: Browse list tab for number, date, and datetime filters

BrianWhitneyAI:
Stacked on #954. Expands the Search / Browse list tab layout to number, date, and datetime annotations: a "Range" tab sits beside "Browse list", so users can pick discrete values from the list or enter a range. Range pickers open on the range inputs; short text value lists still open on the list.

- The tab is hidden for top-level file attributes (file name, size, uploaded, etc.) because their values are never fetched, so the list would be empty.
- Verified against production FMS for Kind, Plate Barcode, Image Size X, Processing Date, Plate Seeded On, Imaging Date (73k values), Total Time Duration, and Is Aligned. A range filter and discrete list values on the same annotation are OR-joined, i.e. "in range or equal to".

Tests: Browse list on a number annotation; tab hidden for a top-level attribute.


packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -63,9 +64,26 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
...
+    // Types with a dedicated picker (text search, number/date range) also offer a
+    // "Browse list" tab. Short string value lists open on the list; range pickers
+    // always open on the range inputs. The default depends on asynchronously loaded
+    // values, so it stays separate from the user's explicit choice.
+    const defaultTab =
+        props.annotation.type === AnnotationType.STRING && items.length > 0 && items.length <= 100
+            ? "browse"
+            : "picker";
+    const [selectedTab, setSelectedTab] = React.useState<"picker" | "browse">();
     const activeTab = selectedTab ?? defaultTab;
```
aswallace: Similar but opposite request to [this one](https://github.com/AllenInstitute/biofile-finder/pull/954#discussion_r4050710218): should the default tab also depend on whether there are already active filters? e.g., if a user has already selected from the browse list should it open back to that list, & if they've made a range selection should that be what it reopens to

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -63,9 +64,26 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
...
+
+    // Types with a dedicated picker (text search, number/date range) also offer a
+    // "Browse list" tab. Short string value lists open on the list; range pickers
+    // always open on the range inputs. The default depends on asynchronously loaded
+    // values, so it stays separate from the user's explicit choice.
+    const defaultTab =
+        props.annotation.type === AnnotationType.STRING && items.length > 0 && items.length <= 100
+            ? "browse"
+            : "picker";
+    const [selectedTab, setSelectedTab] = React.useState<"picker" | "browse">();
```
aswallace, line 86: I'm not sure about "picker" as the term, since the "browse" component is also called a "list picker" which makes it confusing to differentiate which is which. So far I've got "form", "custom", "criteria", "condition" (since filtering on a condition instead of selecting), but unsure if any of those feels right either

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -17,6 +17,7 @@ import AnnotationName from "../../entity/Annotation/AnnotationName";
 import { AnnotationType } from "../../entity/AnnotationFormatter";
 import FileFilter, { FilterType } from "../../entity/FileFilter";
 import { interaction, selection } from "../../state";
+import { TOP_LEVEL_FILE_ANNOTATION_NAMES } from "../../constants";
```
aswallace, line 20: nit: import order

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -63,9 +64,26 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
...
+    // FILE_SIZE is excluded: range filtering is not yet supported for it in the backend.
+    const typeHasDedicatedPicker =
+        props.annotation.name !== AnnotationName.FILE_SIZE &&
+        [AnnotationType.NUMBER, AnnotationType.DATE, AnnotationType.DATETIME].includes(
+            props.annotation.type
+        );
+
+    // Types with a dedicated picker (text search, number/date range) also offer a
+    // "Browse list" tab. Short string value lists open on the list; range pickers
+    // always open on the range inputs. The default depends on asynchronously loaded
```
aswallace, line 80: > range pickers always open on the range inputs

Should this still be true if a user has selected values from the browse list and then closes/reopens the filter? (Related to [this comment](https://github.com/AllenInstitute/biofile-finder/pull/954#discussion_r4050710218) on the other PR)

aswallace (approved):

SeanDuHare (approved):

packages/core/components/AnnotationFilterForm/index.tsx
BrianWhitneyAI: updated in https://github.com/AllenInstitute/biofile-finder/pull/959/commits/f680a4a225f1af357eb95863d6c2a408f992fab7

packages/core/components/AnnotationFilterForm/index.tsx
BrianWhitneyAI: [39e1311](https://github.com/AllenInstitute/biofile-finder/pull/959/commits/39e1311872790e0983825760f0208e00bd953185)

packages/core/components/AnnotationFilterForm/index.tsx
BrianWhitneyAI: [39e1311](https://github.com/AllenInstitute/biofile-finder/pull/959/commits/39e1311872790e0983825760f0208e00bd953185)

PR #960, opened 2026-09-18
Explain that FMS file size cannot be filtered by value

BrianWhitneyAI:
The file explorer service returns 500 for `RANGE(...)` filters on `file_size`. This just puts a note 

CSV and Parquet sources are unaffected since their column is named "File Size" and DuckDB handles the range. The real fix is server-side support for `RANGE` on `file_size`; once that lands this note can be removed.


aswallace:
This will be resolved (hopefully) once FES is deployed, which should (hopefully) be in the next couple days

BrianWhitneyAI:
oh cool Ill close it and fix da stack then

PR #961, opened 2026-09-18
bugfix: Make "File Path (Local VAST)" filterable against FMS

### LINKED ISSUE #792: "File Path (Local VAST)" annotation appears as a filter option but fails with unknown annotation error
## Description
The "File Path (Local VAST)" annotation appears in the Filter annotation picker but cannot be used as a filter. Attempting to filter by it returns a server error.

## Steps to Reproduce
1. Open the Filter panel
2. Select **File Path (Local VAST)** from the annotation picker
3. Enter any value and submit

## Expected Behavior
Either the annotation is filterable and returns results, or it does not appear in the filter picker at all.

## Actual Behavior
The FES returns a CLIENT_ERROR:
```json
{
  "error": "Attempt to query files with an unknown annotation: File Path (Local VAST) is not a known annotation.",
  "responseType": "CLIENT_ERROR"
}
```

## Root Cause
`LOCAL_FILE_PATH` (`"File Path (Local VAST)"`) is returned by the FES `/annotations` endpoint (so it populates the annotation list in Redux and appears in the filter picker), but the FES `/files` query endpoint does not support filtering by it. There is no client-side guard to exclude it from the filter annotation picker.

**Relevant files:**
- `packages/core/entity/Annotation/AnnotationName.ts` — defines `LOCAL_FILE_PATH`
- `packages/core/components/QueryPart/QueryFilter.tsx` — renders the filter `AnnotationPicker` with no exclusions
- `packages/core/components/AnnotationPicker/index.tsx` — displays all annotations from the store without filtering by queryability

## Fix
Exclude `LOCAL_FILE_PATH` from the filter annotation picker. This can be done by adding an `excludedAnnotationNames` prop to `AnnotationPicker` and passing `[AnnotationName.LOCAL_FILE_PATH]` from `QueryFilter.tsx`.

BrianWhitneyAI:
### Description 

Resolves #792 

FES lists this annotation but synthesizes it per file record rather than indexing it: filtering on it or requesting its values is rejected as an unknown annotation, and it is reported with type "String" instead of "Text". Every local path is the file's `file_path` under a different NAS mount, so filters are rewritten client-side before they are sent:

- **Exactly match**: a full local path becomes an exact `file_path` filter, with the mount prefix selecting the environment (production, staging, ...)
- **Contains**: a leading mount prefix is stripped, then a fuzzy `file_path` filter is used
- **Any value / No value**: a file has a local path exactly when it should be in the local cache, so these become `Should Be in Local Cache` = true / false (every production file carries that flag, so this is exact)

Verified against production: Contains `3500008526_10X` returns 61 files; an exact full local path returns 1; Any value / No value split the baseline 70,640 into 47,276 + 23,364.


packages/core/services/AnnotationService/HttpAnnotationService/index.ts
```diff
@@ -51,6 +52,11 @@ export default class HttpAnnotationService extends HttpServiceBase implements An
                 (annotationResponse) =>
                     new Annotation({
                         ...annotationResponse,
+                        // FES reports this synthesized annotation's type as "String" rather than "Text"
```
aswallace, line 55: This seems like it should be a change in FES instead of being covered by BFF. It looks like the issue [is here](https://github.com/AllenCell/file-explorer-service/blob/68e36e8e6423af5633a4d5896381945e6b123816/src/main/java/org/alleninstitute/aics/fileexplorer/controller/v1/AnnotationController.java#L82), which I think could just be changed to `AnnotationType.Text`

packages/core/entity/FileFilter/index.ts
```diff
@@ -123,6 +125,29 @@ export default class FileFilter {
...
+        // "File Path (Local VAST)" is synthesized by FES on file records but is not queryable.
+        if (name === AnnotationName.LOCAL_FILE_PATH) {
+            if (this.type === FilterType.ANY) {
+                return `${AnnotationName.SHOULD_BE_IN_LOCAL}=true`;
+            }
+            if (this.type === FilterType.EXCLUDE) {
+                return `${AnnotationName.SHOULD_BE_IN_LOCAL}=false`;
+            }
+        }
+        if (name === AnnotationName.LOCAL_FILE_PATH && typeof this.value === "string") {
```
aswallace, line 138: a couple questions:
- this implies that if `typeof this.value !== "string"` it will fall through to the switch? which I assume we don't want if it's not queryable. also not sure what the case would be where that would happen, but regardless, doesn't seem to be handled here
- does this and line 130 need to be in separate if conditions like this?
Possibly this could just be its own breakout function for handling local_file_path since it doubles the length of the method for one special case

PR #962, opened 2026-09-18
feat: (perf improv) Speed up filter form display-value formatting for large value lists

BrianWhitneyAI:
### Description

The became apparent during the new annotation forum audit . Independent of the filter UI stack. The date and datetime formatters constructed an `Intl.DateTimeFormat` (directly, or implicitly via `toLocaleString`) for every value, at roughly 25 microseconds each. Production datetime annotations can have tens of thousands of distinct values (Imaging Date has ~74k), so opening their filter form spent about two seconds formatting. 

| Formatter (73,718 values) | Before | After |
|---|---|---|
| Datetime | 2,140 ms | 96 ms |
| Date | 1,816 ms | 36 ms |

Output is unchanged! 


aswallace (approved):

packages/core/components/AnnotationFilterForm/index.tsx
```diff
@@ -59,16 +59,26 @@ export default function AnnotationFilterForm(props: AnnotationFilterFormProps) {
 
     const [filterType, setFilterType] = React.useState<FilterType>(defaultFilterType);
 
+    // Format display values once per value list. Formatting is O(n) and can be expensive
+    // (e.g. Intl date formatting), so keep it out of the memo that reacts to filter changes.
+    const formattedValues = React.useMemo(
```
SeanDuHare: (minor nit) to provide type enforcement for the values you do provide

```suggestion
    const formattedValues: ListItem[] = React.useMemo(
```

or Partial<ListItem>[] if `selected` is required, didn't check

SeanDuHare (approved):

packages/core/components/AnnotationFilterForm/index.tsx
BrianWhitneyAI: this didnt work raw, had to do [3dc1649](https://github.com/AllenInstitute/biofile-finder/pull/962/commits/3dc1649c107b89dba58e503ba3c4d2bf0a83a84b)

PR #963, opened 2026-09-18
bugfix: Fetch values for annotations whose names contain a slash

### LINKED ISSUE #964: Bug: Filter form errors for annotations whose names contain a slash
### Bug
Opening the filter form for an annotation whose name contains a `/` (e.g. "Shear Stress 2 target (dyn/cm2)") shows "Whoops! Encountered an error" instead of the value picker.

### Cause
The annotation values endpoint takes the annotation name as a URL path segment. A raw `/` splits the route (404 Not Found), and an encoded `%2F` is rejected by Tomcat (400). Filtering itself is unaffected, since filters travel as query parameters.

### Affected annotations
The four "Shear Stress N … (dyn/cm2)" annotations in production — the only annotation names containing a slash today.

### Steps to reproduce
1. Open the app against the AICS FMS data source
2. Add a filter and search "Shear Stress"
3. Open "Shear Stress 2 target (dyn/cm2)"
4. The filter form shows "Whoops! Encountered an error: Not Found"

BrianWhitneyAI:
Closes #964

### Description
Found this while auditing our annotations via the new annotation picking. For Annotation names like "Shear Stress 2 target (dyn/cm2)" a raw slash breaks and causes (404 Not Found) and an encoded `%2F` is rejected by Tomcat (400), so the filter form showed "Whoops! Encountered an error" for these annotations. 

### Fix
Values for such names are now requested through the hierarchy root endpoint, which takes the name as a query parameter and returns the same distinct values. Filtering was never affected since filters already travel as query parameters.

Covers the four shear stress annotations in production, the only names with slashes. Verified: the form opens with range 6–35, Browse list shows 6 / 25.44 / 25.91 / 35, selecting 6 returns 11 files.


aswallace (approved):

packages/core/services/AnnotationService/HttpAnnotationService/index.ts
```diff
@@ -91,6 +91,14 @@ export default class HttpAnnotationService extends HttpServiceBase implements An
      * Fetch the unique values for a specific annotation.
      */
     public async fetchValues(annotation: string): Promise<AnnotationValue[]> {
+        // The values endpoint takes the annotation name as a path segment. A "/" in the name
+        // (e.g. "Shear Stress 1 target (dyn/cm2)") either splits the route (404) or, once encoded,
+        // is rejected by the server (400). The hierarchy root endpoint returns the same distinct
+        // values but takes the name as a query parameter, which encodes safely.
+        if (annotation.includes("/")) {
```
SeanDuHare: I'd never considered how those two endpoints serve the same function for that case - nice!

SeanDuHare (approved):

PR #965, opened 2026-09-18
feat: Added prettier import sorting plugin

ShrimpCryptid:
## Context

Experiment based on a discussion where Anya pointed me to #578, "add import pattern to linting."

This currently adds a prettier import sorting plugin with rules to match (most of) BFF's current import sorting standard, and there are some example included files to see the kinds of changes this file makes. Note that it **does not currently separate out the file being tested for test files**, as defined in Anya's original spec. 

I did want to ask if this is the *right* approach though-- this will automatically sort imports in files when they're modified or staged. Should we update the entire repository, or is this kind of incremental approach better?

## Changes
- Adds the `@trivago/prettier-plugin-sort-imports` package as a dev dependency.
- Configures prettier to match existing BFF import sorting rules.
- Updates the pre-commit hooks to apply import sorting when running prettier.


.prettierrc
```diff
@@ -1,7 +1,26 @@
...
+        "^../../[\\w/]*[\\w.]*(?<!.css)$",
+        "^../../../[\\w/]*[\\w.]*(?<!.css)$",
+        "^../../../../[\\w/]*[\\w.]*(?<!.css)$",
+        "^../../../../../[\\w/]*[\\w.]*(?<!.css)$",
+        "^../../../../../../[\\w/]*[\\w.]*(?<!.css)$",
+        "^../../../../../../../[\\w/]*[\\w.]*(?<!.css)$",
+        "^../../../../../../../../[\\w/]*[\\w.]*(?<!.css)$",
+        "<SEPARATOR>",
+        "^./.+\\.css$"
+    ],
```
ShrimpCryptid: (Mostly) follows the rules defined in https://github.com/AllenInstitute/biofile-finder/issues/578.

Third party modules have their own section, and all relative imports (besides CSS) are sorted into the middle block by distance from the current file. CSS files are imported last.

Note that the rule that the tested file be separated out is not kept-- I can't figure out a way to do that short of forking the plugin altogether.

.prettierrc
```diff
@@ -1,7 +1,27 @@
...
+        "^\\.\\./\\.\\./\\.\\.[\\w/]*[\\w.]*(?<!.css)$",
+        "^\\.\\./\\.\\./\\.\\./\\.\\.[\\w/]*[\\w.]*(?<!.css)$",
+        "^\\.\\./\\.\\./\\.\\./\\.\\./\\.\\.[\\w/]*[\\w.]*(?<!.css)$",
+        "^\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\.[\\w/]*[\\w.]*(?<!.css)$",
+        "^\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\.[\\w/]*[\\w.]*(?<!.css)$",
+        "^\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\.[\\w/]*[\\w.]*(?<!.css)$",
+        "^\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\./\\.\\.[\\w/]*[\\w.]*(?<!.css)$",
+        "<SEPARATOR>",
+        "^./.+\\.css$"
+    ],
```
ShrimpCryptid, line 23: So this is definitely a mess 🤦 Unfortunately, special characters (`.`) sort before alphanumeric ones, which means that I need to explicitly set a matching pattern for each depth level. `\\` is used to escape every `.` character, so it isn't interpreted as an "any character" regex.

The biggest import depth in BFF is currently five levels (`../../../../../`), and this stack of regex is setup to capture up to nine.

aswallace:
We talked about this offline, but personally my vote is to replace relative imports with aliasing `packages/core` to `@/core`, since the main reason we run into those long `../` chains is because of importing core components into web/desktop, and maybe even doing `@/web` and `@/desktop` as well. The challenge would be that this would touch literally ALL of our files, but the benefit would be that we would be able to switch to a purely alphabetical import structure instead of having to do this intermediary custom regex lint step.

We could also look into switching to using [npm workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces)? which could also theoretically make our imports cleaner

aswallace:
> this will automatically sort imports in files when they're modified or staged. Should we update the entire repository, or is this kind of incremental approach better?

If we go this route, I'm fine with it being incremental. Import order isn't a breaking issue, it's just stylistic, and imo this is really more for onboarding new devs than for strict enforcment across the codebase

SeanDuHare:
> > this will automatically sort imports in files when they're modified or staged. Should we update the entire repository, or is this kind of incremental approach better?
> 
> If we go this route, I'm fine with it being incremental. Import order isn't a breaking issue, it's just stylistic, and imo this is really more for onboarding new devs than for strict enforcment across the codebase

I think the opposite - incremental might clutter PRs for a long time

ShrimpCryptid:
> I think the opposite - incremental might clutter PRs for a long time

Okay, then in that case, I can open up a follow-up PR stacked after this one, that applies the Prettier rules to all existing files!

ShrimpCryptid:
> Okay, then in that case, I can open up a follow-up PR stacked after this one, that applies the Prettier rules to all existing files!

Attempted to do this and have a branch with the changes... it's nearly 10,000 line changes across 177 files, with many files also getting all whitespace updated by Prettier. I'm a little worried that this will be... infeasible to review?

